PR #5, opened 2015-05-04
Fixing potential RPMsg bug: rpmsg_endpoint structure is not cleared when...

MichalPrincFSL:
... dynamically created

- this causes the hardfault @line 570 when rp_chnl is not a null pointer (writing to undefined dereferenced address)

### DIFF
--- a/rpmsg/rpmsg_core.c
+++ b/rpmsg/rpmsg_core.c
@@ -216,6 +216,7 @@ struct rpmsg_endpoint *_create_endpoint(struct remote_device *rdev,
     if (!rp_ept) {
         return RPMSG_NULL ;
     }
+    env_memset(rp_ept, 0x00, sizeof(struct rpmsg_endpoint));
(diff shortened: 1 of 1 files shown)


wjliang:
The patch looks good to me. However, just want to know in what case this hardfault happens?

Just want to make sure whether there is other fix required to solve the issue, since in usual case, the rp_chanl is already created before it is used.

MichalPrincFSL:
Hi,

During the rpmsg init on master side, the name service announcement endpoint is created, i.e. the _create_endpoint() is called and the rp_ept->rp_chnl remains undefined / not a null value. Then, during the name service handshaking the rpmsg_rx_callback() is invoked on the master side, rp_chnl != null, the condition @line 567 passes and the wrong pointer is dereferenced @line 570. I think this situation happens when NS feature is switched on, correct?

Regards
Michal Princ

PR #7, opened 2015-08-21
Use stdint.h instead of private definition

weif:
This patch helps to avoid duplicated type definition issue when the
application uses stdint.h.
At the same time, buffer length type is uniformed.

Signed-off-by: Feng Wei <wei.feng@freescale.com>

### DIFF
--- a/rpmsg/remote_device.c
+++ b/rpmsg/remote_device.c
@@ -484,16 +484,16 @@ void rpmsg_rdev_set_status(struct virtio_device *dev, unsigned char status) {
 
 }
 
-unsigned long rpmsg_rdev_get_feature(struct virtio_device *dev) {
+uint32_t rpmsg_rdev_get_feature(struct virtio_device *dev) {
     return dev->features;
 }
 
-void rpmsg_rdev_set_feature(struct virtio_device *dev, unsigned long feature) {
+void rpmsg_rdev_set_feature(struct virtio_device *dev, uint32_t feature) {
     dev->features |= feature;
 }
 
-unsigned long rpmsg_rdev_negotiate_feature(struct virtio_device *dev,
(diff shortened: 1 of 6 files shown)


wjliang:
There is one change in the rpmsg_get_tx_buffer(), to change the input argument "len" from type "int", to "unsigned long". Why do we need this change? Should we keep it as "int"?

Thanks,
Wendy

weif:
@wjliang This is just to make rpmsg_get_tx_buffer arguments align with rpmsg_get_rx_buffer. You can see rpmsg_get_rx_buffer uses "unsigned long" for buffer length. I believe it doesn't make sense to have different type for rx and tx buffer length.

wjliang:
since this patch is trying to use, stdint, i would prefer to use int32_t for the "len" for both rpmsg_get_tx_buffer and the rpmsg_get_rx_buffer, and would prefer to add "#include <stdint.h>" to the

weif:
Well, this patch wants to fix the issue that some app uses stdint.h in its app code when rpmsg.h is included. It's not intended to change the RPMsg API to stdint types.
You may see the include path: rpmsg.h includes rpmsg_core.h includes virtio.h includes virtqueue.h, and although the app doesn't want to access the virtio structure, the duplicated definition happens.
We can't change the virtio type because the protocol need fixed length integer type. At the same time, we don't want to change the RPMsg API. That's why we need some type conversion in rpmsg core implementation.

About the buffer length type, it's an internal type and doesn't have to be same sign with payload length defined in rpmsg.h. I still prefer unsigned type because virtio uses unsigned type for buffer length. I didn't see any benefit to have the sign converted in rpmsg core.

wjliang:
it is merged into master

PR #9, opened 2015-09-02
iar support

MichalPrincNXP:
Allow IAR EWARM compiler to be used (pack structure syntax) when building open-amp projects

### DIFF
--- a/lib/rpmsg/remote_device.c
+++ b/lib/rpmsg/remote_device.c
@@ -128,6 +128,9 @@ int rpmsg_rdev_init(struct remote_device **rdev, int dev_id, int role,
 	rdev_loc->channel_destroyed = channel_destroyed;
 	rdev_loc->default_cb = default_cb;
 
+	/* Restrict the ept address - zero address can't be assigned */
+	rdev_loc->bitmap[0] = 1;
+
 	/* Initialize the virtio device */
 	virt_dev = &rdev_loc->virt_dev;
 	virt_dev->device = proc;
--- a/lib/rpmsg/rpmsg_core.c
+++ b/lib/rpmsg/rpmsg_core.c
@@ -709,7 +709,7 @@ int rpmsg_get_address(unsigned long *bitmap, int size)
 		tmp32 = get_first_zero_bit(bitmap[i]);
 
 		if (tmp32 < 32) {
-			addr = tmp32 + i + 1;
+			addr = tmp32 + (i*32);
 			bitmap[i] |= (1 << tmp32);
 			break;
 		}


wjliang:
You also change data[0] to data[1]. Since the rpmsg/remoteproc is from kernel, we should line up with what's in the Linux kernel, could you change it back? or does it also cause any issue when you use IAR compiler?

Others look good

MichalPrincFSL:
Hello wjliang,

Zero-length arrays have been replaced across all open-amp files (structures) because this is GCC – specific syntax and it is not possible to compile this in IAR EWARM.

As for remoteproc, we are not using it now but the changes in zero-length arrays have been done to have the open-amp code consistent. We have also compared the remoteproc.h from upstream Linux with the one from open-amp and there are lots of incompatibilities in structures (fw_rsc_carveout, fw_rsc_devmem, fw_rsc_trace, etc.) so it is not clear to me why it should be an issue just for the remoteproc component. Could you please clarify more? Thank you.

Best Regards
Michal Princ

wjliang:
After I apply this patch, i failed to setup connection between Linux and CortexR5 on zynqMP platform.

Have you tested the patch when using GCC compiler?

MichalPrincFSL:
Hello wjliang,

I am sorry for the late answer…
Yes, we have tried the GCC compiler. Communication between Cortex A7 (Linux) and Cortex M4 (baremetal, EWARM C++ compiler) works fine. I was not accurate in my previous answer, the zero-length arrays issue has been identified when using the EWARM C++ compiler, C compiler in IAR is OK with that syntax.

Regards
Michal

wjliang:
Hi Michal,

I still haven't got any luck when my baremetal remote is using Cortex R5 after applying the patch. I will try the patch on Xilinx Zynq platform (A9) to see.

Regards,
Wendy

MichalPrincFSL:
Hello Wendy,

Thanks for the update. Two points that should be considered:

a.       We have not used the remoteproc component in our projects. The requested changes in remoteproc.h have been done just to keep the consistence and make the changes in data structures consistent across all open-amp files. Maybe I was not accurate and have not explicitly describe that in my pull request and consequent emails. Could you please revert the changes in remoteproc.h to see if this is the root cause of failed setup connection between Linux and CortexR5 on zynqMP platform?

b.      The pull request mixes the structures packing issue and the zero-length array issue. Could you please try to adjust the code and apply just one of these change requests to find out what causes your malfunction, if it is ‘packing’ or ‘arrays’?

Thank you.

Regards
Michal

wjliang:
Hi Michal,

After some debugging, the failure is due to "fw_rsc_vdev " in remoteproc.h, after "struct fw_rsc_vdev_vring vring[0]" has been replaced by "struct fw_rsc_vdev_vring vring[1]". When I am testing Linux as master, and only use baremetal as slave, i can workaround this issue. However, zero-length array in GNU C means variable length object. I have tested that a header with zero-length array is different to a header with length "1" array.

I have checked that in the code, when the baremetal is used as master, there are places to calculate the length with sizeof(structu rpmsg_hdr) and so on.

Could you verify the code running the baremetal/rtos as master mode?

The test_suite modification looks good, just the modification to rpmsg and remoteproc is not complete.

Thanks,
Wendy

MichalPrincFSL:
Hi Wendy,

Thanks for finding the rootcause and separating the issue. As I have written before, we are not using the remoteproc and are unable to try the baremetal/rtos as master. What we have verified is that changing the zero-length arrays to [1] does not cause any compatibility issues in RPMsg and when mixing these two implementations ([0] and [1]) on a dual-core platform it works fine.

All the motivation for this change is to allow RPMsg compilation in C++. If there is no suitable solution for the remoteproc we can leave that change in RPMsg only and keep the original code in remoteproc.

Best Regards
Michal Princ

wjliang:
Hi Michal,

i just want to make sure it also works in master mode, i haven't got time yet to try the master mode. I will try the master mode as soon as possible and update you later.

Best Regards,
Wendy

MichalPrincFSL:
Thank you, Wendy. We will wait for your master mode results and the update…

Best Regards
Michal Princ

MichalChromec:
Hello Wendy,
I am colleague of Michal Princ and I would like to know the status of pull request.
Have you tried the master mode?

Best Regards
Michal Chromec

wjliang:
Hi Michal,

I haven't tried out the master mode yet. I haven't have a handy test.
I will try it next week.

About the patch, as mentioned earlier, for slave mode, additional change is required to remoteproc.h, please see the comment on Sep 20.

Best Regards,
Wendy

wjliang:
Hi Michal,

I have tried your patch in baremetal as master, it failed to start the communication. I have gone through the patch, could you do the following modification?:

 use "char data[1];" for "#if defined(__IAR_SYSTEMS_ICC__)" and use "char data[0]" for if "__IAR_SYSTEMS_ICC__" is not defined. And the same for other "0" length array declaration.

Thanks,
Wendy

wjliang:
Hi Michal,

I had a discussion on the patch with Cyril.

We cannot take this patch, because
 * change the array size from "0" to "1" will change the size of the structures in GCC. Will need to figure out a better solution.
 * we are seeing you use "__pack" before the structure, could you use it to the end of the struct in IAR compiler. If yes, we don't have to have "if defined" everywhere, we can define the pack attribute in one file according to the compiler, and place it to the end of the struct.

Best Regards,
Wendy

MichalPrincFSL:
Hello Wendy

  *   change the array size from "0" to "1" will change the size of the structures in GCC. Will need to figure out a better solution.
     *   Yes, the goal is to find the solution that is general, not the GCC-specific. What solution do you propose? We can confirm that applying the array size change from "0" to "1" in rpmsg_hdr structure works fine when using GCC on both the CM4 side and the CA7 side. It additionally allows building this in IAR C++. What about to take the rpmsg.h-related part of this patch only, i.e. not considering the rest of the patch (changes in remoteproc we were unable to verify/try on our side)?
  *   we are seeing you use "__pack" before the structure, could you use it to the end of the struct in IAR compiler. If yes, we don't have to have "if defined" everywhere, we can define the pack attribute in one file according to the compiler, and place it to the end of the struct.
     *   Unfortunately, IAR compiler syntax is a little different. To force the compiler to pack a structure, one can use either the _ _packed keyword or to use the #pragma pack(n) directive, see http://supp.iar.com/FilesPublic/UPDINFO/004916/arm/doc/EWARM_DevelopmentGuide.ENU.pdf /<http://supp.iar.com/FilesPublic/UPDINFO/004916/arm/doc/EWARM_DevelopmentGuide.ENU.pdf%20/> p. 217. Using the __attribute__((packed)) is not allowed. That is the reason for that “not nice” #if defined(__IAR_SYSTEMS_ICC__) construction.
Regards
Michal

wjliang:
Hi Michal,
* About changing the array size from "0" to "1"
      * Does the IAR compiler accept flexible array member that is e.g. "char data[];"?, if yes, could you try it?
      * If have to change from "0" to "1", all the code in the repo which using the structure need to adjust it, especially where it tries 'sizeof(xxxx)". E.g. in the rpmsg.c, it uses "sizeof(struct rpmsg_hdr)".

* About the "__packed", is is correct that it has to be put in the front of the struct? What we are thinking is have a generic arch dependent file. and then put sth like
'''
#ifdef defined(__IAR_SYSTEMS_ICC__)
# define PACK __pack
# else
#define PACK __attribute__((packed))
#endif
'''
And then in the struct definition, can just use "PACK" there, if __pack to be put to different places than the GNU GCC one, it can be sth like:
'''
#ifdef defined(__IAR_SYSTEMS_ICC__)
# define PACK(v) __pack v
# else
#define PACK(v) v __attribute__((packed))
#endif
'''
Or sth like that.

Best Regards,
Wendy

MichalPrincFSL:
Hello Wendy,


·         About changing the array size from "0" to "1"

o   I am afraid flexible array members are not allowed in C++. Having empty brackets in IAR/C++ leads to the “incomplete type is not allowed” error.

o   Yes, sizeof(struct rpmsg_hdr) does not work correctly when having data[1], using this we reduced the buffer size by one byte but the communication works (kind of workaround)

o   What about to avoid using data member of the rpmsg_hdr struct and each time the pointer to the data start address is necessary in the code it will be calculated using the virtio buffer start address and the sizeof(struct rpmsg_hdr)?

·         About the PACK(v) macro: yes, that is a good proposal and I have verified that is works in IAR. The point is where to place this PACK macro definitions:
#if defined(__IAR_SYSTEMS_ICC__)
#define PACK(v) __packed v
#else
#define PACK(v) v attribute((packed))
#endif

Best Regards
Michal

wjliang:
Hello Michal,

* About the rpmsg_hdr, I think it OK to make the rpmsg_hdr not to include the data.
* About the PACK(v) definition, how about put it to "common/compiler.h"?

By the way, could you send the patch to open-amp@googlegroups.com when you update the patch?

Best Regards,
Wendy

wjliang:
Hello Michal,

Cyril pointed out that the above method can have an issue when there is comma in the structure. And I confirmed it. This is a a better way:
```
#define PACKED(_type, ...) _type { __VA_ARGS__ }  __attribute__((packed))

PACKED(struct test_D,
	char a, z;
	int c;
	char b[10];
);
```
We tested it.
Please use newer PACKED(_type,...) instead of the old PACK(v)

Best Regards,
Wendy

MichalPrincFSL:
Hello Wendy,

agree. I can confirm that this construction works fine in IAR:
```
#if defined(__IAR_SYSTEMS_ICC__) 
#define PACKED(_type, ...) __packed _type { __VA_ARGS__ }
#else 
#define PACKED(_type, ...) _type { __VA_ARGS__ }  __attribute__((packed))
#endif
```
Will prepare the pull request update.

Best Regards
Michal

flit:
(Fyi, I'm a coworker of Michal.)

Instead of the PACKED() macro, what about defining both PACKED_PRE and PACKED_POST macros? I'm not all that comfortable with wrapping entire struct definitions inside a macro.

Example:
`````c
#if defined(__IAR_SYSTEMS_ICC__)
#define PACKED_PRE __packed
#define PACKED_POST
#else
#define PACKED_PRE
#define PACKED_POST __attribute__((packed))
#endif

typedef PACKED_PRE struct FooStruct { 
    char a, z;
    int c;
    char b[10];
} PACKED_POST foo_struct_t;
`````

It's more straightforward, and won't confuse various code-processing tools (perhaps code formatters, etc) or text editors.

On the topic of the 0/1 struct member array length: You could use an array value of 1, then define a macro per struct to compute the struct's size by taking into account the size of the array.

For instance:
`````c
struct FooStructWithArray {
    int x;
    int y;
    uint8_t data[1];
};

#define SIZEOF_FOO_STRUCT_WITH_ARRAY(datacount) \
    (sizeof(FooStructWithArray) - sizeof(uint8_t) + (sizeof(uint8_t) * (datacount)))
`````

This would guarantee code that works in all compilers and both C and C++. (We have a requirement to build with C++.) This is a very common pattern for compiler-agnostic or (pre-)C89 code.

wjliang:
Hi Chris,

PACK_PRE/PACK_POST are also fine. Just I also see some people define such as "PACK_PRE, PACK, and PACK_POST" for some compiler, and thus, I think one PACKED is simpler. Does it fail your C++ compiler?

About whether to put "unit8_t data[1]" in the structure, the good thing is it shows the data type. However, it is easy to make mistake when calculating the size of the struct (especially a packed one), unless defining macro to do so. In case of rpmsg, the length of header is used a lot, but not much of the full length of the header and array, and thus, i am more inclined to Michal's proposed solution not to have data[] in the header.

Best Regards,
Wendy

MichalPrincFSL:
Hi Chris, Wendy,

one PACKED is simpler and works fine with our C++ compiler. On the other hand I agree with Chris that using a pair of pack macros is more straightforward when displaying in various text editors. Also, we should consider that not only IAR and GCC but also other compilers our customers are using needs to be supported (ARM GCC etc.). We are pretty confident that having two packed macros works fine (from different projects). So, I am for having two pack macros.

As for the 0/1 struct member array length, I agree with Wendy to avoid accidental use of sizeof(struct rpmsg_hdr) instead of the dedicated macro.

Michal

flit:
Yes, I agree with you about the 0/1 struct since the header size is primarily used. But my suggestion should be considered for other cases outside of RPMsg. Just a thought.

wjliang:
Hi Michal, Chris,

I agree to go with two packed macros.
About the 0/1, i think in general, even outside rpmsg, it is better to exclude data in header struct. If we do need to have a struct with data, then as suggested by Chris, better to have dedicated macro to calculate size.

Best Regards,
Wendy

PR #15, opened 2016-02-11
Bugfix of memory leak in rpmsg init

MichalPrincNXP:
rpmsg_deinit() did not delete proc instance initialized during the rpmsg_rdev_init()


wjliang:
applied

PR #16, opened 2016-02-11
Bugfix of rpmsg_get_address implementation

MichalPrincNXP:
Fix the way the create endpoint function returns the address number when RPMSG_ADDR_ANY is passed the parameter

This bug has been revealed when performing following steps:
1. create one endpoint with address RPMSG_ADDR_ANY.
2. get the address of that endpoint after it is created successfully.
3. use that address to create another endpoint.
It should not create the second endpoint with address of the first endpoint but somehow both endpoints were created successfully.


wjliang:
cannot apply due to issue found in the zynq baremetal master and linux slave test.

Here is the suggested change:
diff --git a/lib/rpmsg/rpmsg_core.c b/lib/rpmsg/rpmsg_core.c index ccb09ca..800a245 100644
--- a/lib/rpmsg/rpmsg_core.c
+++ b/lib/rpmsg/rpmsg_core.c
@@ -709,7 +709,7 @@ int rpmsg_get_address(unsigned long *bitmap, int size)
                tmp32 = get_first_zero_bit(bitmap[i]);
 
                if (tmp32 < 32) {
-                       addr = tmp32 + i + 1;
+                       addr = tmp32 + (i * 32) + 1;
                        bitmap[i] |= (1 << tmp32);
                        break;
                }
@@ -734,7 +734,7 @@ int rpmsg_release_address(unsigned long *bitmap, int size, int addr)
        unsigned int i, j;
        unsigned long mask = 1;
 
-       if (addr >= size * 32)
+       if (--addr >= size * 32)
                return -1;
 
        /* Mark the addr as available */ @@ -763,7 +763,7 @@ int rpmsg_is_address_set(unsigned long *bitmap, int size, int addr)
        int i, j;
        unsigned long mask = 1;
 
-       if (addr >= size * 32)
+       if (--addr >= size * 32)
                return -1;
 
        /* Mark the id as available */
@@ -790,7 +790,7 @@ int rpmsg_set_address(unsigned long *bitmap, int size, int addr)
        int i, j;
        unsigned long mask = 1;
 
-       if (addr >= size * 32)
+       if (--addr >= size * 32)
                return -1;
 
        /* Mark the id as available */

MichalPrincNXP:
What about just to restrict the "0" endpoint address by default during the rdev init?


```
diff --git a/lib/rpmsg/remote_device.c b/lib/rpmsg/remote_device.c
index 451509e..5a68a30 100644
--- a/lib/rpmsg/remote_device.c
+++ b/lib/rpmsg/remote_device.c
@@ -127,6 +127,9 @@ int rpmsg_rdev_init(struct remote_device **rdev, int dev_id, int role,
 	rdev_loc->channel_destroyed = channel_destroyed;
 	rdev_loc->default_cb = default_cb;
 
+    /* Restrict the ept address - zero address can't be assigned */
+    rdev_loc->bitmap[0] = 1;
+
 	/* Initialize the virtio device */
 	virt_dev = &rdev_loc->virt_dev;
 	virt_dev->device = proc;
```

wjliang:
Hi Michal,

You proposal sounds good. Could you please try and provide a new patch?

Thanks,
Wendy

MichalPrincNXP:
Hi Wendy,

are you OK with adding a new commit into this branch or would you prefer to discard this pull request and to create a new pull request/patch that combines these two commits in one (single commit on master when integrated)?

Michal

wjliang:
Hi Michal, add a new patch should be fine.

thanks,
Wendy

MichalPrincNXP:
moved to the newly created pull request, see #19

PR #18, opened 2016-02-17
Avoid zero length struc members in rpmsg_hdr structure

MichalPrincNXP:
Avoid using zero length arrays in rpmsg_hdr struct, unsigned char data[1] removed from rpmsg_hdr structure, see #9 .


wjliang:
I changed "rp_ept->cb(rp_chnl, (rpmsg_rx_cb_t*)RPMSG_LOCATE_DATA(rp_hdr), rp_hdr->len," to "rp_ept->cb(rp_chnl, (void*)RPMSG_LOCATE_DATA(rp_hdr), rp_hdr->len," in the rpmsg_core.c function

PR #19, opened 2016-02-18
Bugfix of rpmsg_get_address implementation 2

MichalPrincNXP:
see #16 

Fix the way the create endpoint function returns the address number when RPMSG_ADDR_ANY is passed the parameter

This bug has been revealed when performing following steps:
1. create one endpoint with address RPMSG_ADDR_ANY.
2. get the address of that endpoint after it is created successfully.
3. use that address to create another endpoint.
It should not create the second endpoint with address of the first endpoint but somehow both endpoints were created successfully.

Restrict the endpoint address - zero address can't be assigned (due to the conflict in Linux world)


wjliang:
applied

PR #20, opened 2016-02-29
Avoid using zero length arrays in sh_mem_pool structure

MichalPrincNXP:
This is another step how to avoid using zero length arrays in structures, as discussed in https://github.com/OpenAMP/open-amp/pull/9 (flexible array members are not allowed in C++).

bitmap member of the sh_mem_pool struct has been removed and each time the pointer to the bitmap is necessary in the code it is calculated using the SH_MEM_POOL_LOCATE_BITMAP macro


wjliang:
applied

PR #21, opened 2016-03-08
Esw sync 1

samsor:
-Fix bug happening on deinit where memory deallocation is not done correctly.
-Fix bug replace logical-AND by bit-AND with ipi_chn_mask  …
-Invalidate cache instead of disabling on exit
-Get drivers file to use platform_isr instead of hil_isr
-Fix some compilation warnings, clean comments, file renaming, make some local variable static


wjliang:
Applied

PR #32, opened 2017-12-01
Irq rework

samsor:
-update to match libmetal IRQ API change
 -fix some typo in toolchain file
 -remove unused variables
 -change r5-generic toolchain file to use hard fp to match default latest Xilinx SDK setting


wjliang:
Merged with modification to commit messages

PR #39, opened 2018-02-05
remoteproc_shutdown fix

tyokota-eforce:
I think rproc->proc doesn't need to assign RPROC_NULL;
Then we can omit proc->proc assignment in follow code.

>  remoteproc_shutdown(proc);
  proc->proc = hproc;
  remoteproc_boot(proc)


wjliang:
Merged with commit message modification.
Please sign the patch in future pull requests.

PR #47, opened 2018-04-09
Update logging and assertions

skordal:
This replaces the openamp_print() and openamp_assert() macros with corresponding functions from libmetal. It also deletes env.h, which ended up being empty after replacing the macros.


wjliang:
it needs another  pull request to clean up commented out lines in the function

skordal:
Taken care of in PR #51

PR #48, opened 2018-04-10
rpmsg: make rpmsg_memb_match/cpy() internal

skordal:
Makes the `rpmsg_memb_match()` and `rpmsg_memb_cpy()` functions static in the modules where they are used, which allows the compiler to inline them.


wjliang:
merged

PR #49, opened 2018-04-10
Use standard macros for bools and NULL in virtqueue

skordal:
Replaces definitions for `true` and `false` with the definitions from stdbool.h, and uses `NULL` instead of `VQ_NULL`.


wjliang:
merged

PR #50, opened 2018-04-11
lib: virtio: suppress virtqueue memory allocation

galak:
To be able to support static allocation, vq should no more allocated in
virtio but in rpmsg or application.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>

### DIFF
--- a/lib/include/openamp/virtqueue.h
+++ b/lib/include/openamp/virtqueue.h
@@ -20,6 +20,7 @@ extern "C" {
 typedef uint8_t boolean;
 
 #include <openamp/virtio_ring.h>
+#include <metal/alloc.h>
 #include <metal/dma.h>
 #include <metal/io.h>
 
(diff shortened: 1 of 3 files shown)


wjliang:
I think the virtqueue_create() should be able to create the virtqueue structure, for static allocation, it looks like we may need to introduce some init() function.

galak:
> I think the virtqueue_create() should be able to create the virtqueue structure, for static allocation, it looks like we may need to introduce some init() function.

@arnop2 thoughts on this?

arnop2:
if we consider that using metal_allocate_memory is ok, we can probably Discard this patch. 
static allocation could also be done in metal_allocate_memory... means that a  big memory table could be defined in metal memory an allocator can just provide a pointer in this table ( in this case no free implemented).
advantage would be that openamp will not to be aware of the memory allocation method.

galak:
Updated to provide a helper alloc function virtqueue_alloc()

galak:
Closing this PR, will open one with all the virtqueue changes.

PR #53, opened 2018-04-12
[RFC] virtqueue: optimize away vq_name for DEBUG only

galak:
This saves about 70 bytes, and vq_name is only a debug option.  Wonder how we want to enable DEBUG from the build, do we use CMAKE_BUILD_TYPE or introduce a DEBUG flag option in cmake

### DIFF
--- a/lib/include/openamp/virtqueue.h
+++ b/lib/include/openamp/virtqueue.h
@@ -65,7 +65,9 @@ typedef enum {
 
 struct virtqueue {
 	struct virtio_device *vq_dev;
+#ifdef DEBUG
 	char vq_name[VIRTQUEUE_MAX_NAME_SZ];
+#endif
 	uint16_t vq_queue_index;
 	uint16_t vq_nentries;
 	uint32_t vq_flags;
--- a/lib/rpmsg/rpmsg_core.c
+++ b/lib/rpmsg/rpmsg_core.c
@@ -91,13 +91,17 @@ int rpmsg_start_ipc(struct remote_device *rdev)
 
 	/* Initialize names and callbacks based on the device role */
 	if (rdev->role == RPMSG_MASTER) {
+#ifdef DEBUG
 		vq_names[0] = "tx_vq";
 		vq_names[1] = "rx_vq";
+#endif
 		callback[0] = rpmsg_tx_callback;
 		callback[1] = rpmsg_rx_callback;
 	} else {
+#ifdef DEBUG
 		vq_names[0] = "rx_vq";
 		vq_names[1] = "tx_vq";
+#endif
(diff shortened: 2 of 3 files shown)


wjliang:
I think we can leave the name here for now, as this align with Linux kernel  implementation. and also it the name can be passed as argument to create or initialize virtquque. Besides print during debugging, we can also use the name for searching.

galak:
I think we should treat this as debug til we actually need it.  Also use of strings for this is expensive in space and time.  If we want something we can add a enum of such and this just map that for debug purposes.  Every byte is important in these microcontroller instances.

wjliang:
> I think we should treat this as debug til we actually need it. Also use of strings for this is expensive in space and time. If we want something we can add a enum of such and this just map that for debug purposes. Every byte is important in these microcontroller instances.

Maybe we keep this feature first but mark it as a place where maybe to optimized for now as this name attribute  matches the  virtqueue  linux kernel implementation. And it looks like we can use it for vq lookup. RPMsg-lite virtqueue implementation also has name attribute.

skordal:
If we have to keep it for compatibility with Linux, could it be an option to define a non-compatible mode for when we are using AMP without Linux on either processor? I think this could potentially help us reduce size in cases where something is only needed for Linux compatibility.

Maybe something like:
```C
// Define to disable linux compatibility and reduce code size:
#define EMBEDDED_MODE

...

#ifndef EMBEDDED_MODE
 	char vq_name[VIRTQUEUE_MAX_NAME_SZ];
#endif
```

galak:
> Maybe we keep this feature first but mark it as a place where maybe to optimized for now as this name attribute matches the virtqueue linux kernel implementation. And it looks like we can use it for vq lookup. RPMsg-lite virtqueue implementation also has name attribute.

I don't see this form of compatibility with linux being a requirement.  There is nothing in the rpmsg API set that exposes the virtqueue name.  There is nothing in virtio or rpmsg transport that requires the name.  The name in Linux is completely used for debug.  I don't see keeping this around as necessary at all as it impacts code size.

I'm also not sure why to wait on code size optimization.  This work can occur concurrently with the refactoring in areas like virtqueue which aren't likely to have much if any change from the refractoring work.

galak:
> If we have to keep it for compatibility with Linux, could it be an option to define a non-compatible mode for when we are using AMP without Linux on either processor? I think this could potentially help us reduce size in cases where something is only needed for Linux compatibility.

I'd prefer we not introduce such a dimension, either we do it based on feature/functionality or something like DEBUG or not.  In this case, the name is purely used for DEBUG.

wjliang:
obsolete change, decoupling has been merged into master

PR #62, opened 2018-04-17
Abstract virtqueue shared buff mem mgmt

galak:
* Removed unused virtqueue_add_single_buffer
* Reworked shared buff memory mgmt to decouple it from libmetal and allow for optimized implementations based on knowing specific details of a system (like there is no address translation so v2p and p2v are very simple).  We default to the existing libmetal & metal_io_region based solution.

In Zephyr these changes reduce code size by approx 130 bytes.


lib/include/openamp/virtqueue.h
```diff
@@ -78,8 +78,7 @@ struct virtqueue {
 	struct vring vq_ring;
 	uint16_t vq_free_cnt;
 	uint16_t vq_queued_cnt;
-	/** Shared memory I/O region */
-	struct metal_io_region *shm_io;
+	void * shm_io; /* opaque pointer to data needed to allow v2p & p2v */
```
wjliang: i think in open-amp layer we can have metal_io_region as the abstraction, for v2p, p2v functions, and the libmetal layer is the one to solve the conversion and i/o memory access, and the I/O address conversion or access optimisation can be done in libmetal layer

galak: I don't think that makes sense -- libmetal isn't going to know the implicit use case of the memory its managing, so it will always end up having more code to allow for the abstractions.

We have to do it on the consumer side of the memory like virtqueue - this way the implementation of the virtqueue memory knows exactly what it needs to provide.

galak:
Closing in favor of a new PR with all vq changes.

PR #63, opened 2018-04-17
virtqueue: Add virtqueue_set_shmem_io & remove passing shmem_io in virtqueue_create

galak:
Pull in changes to the virtqueue from staging branch. This is based on PR #50

### DIFF
--- a/lib/rpmsg/remote_device.c
+++ b/lib/rpmsg/remote_device.c
@@ -436,7 +436,6 @@ int rpmsg_rdev_create_virtqueues(struct virtio_device *dev, int flags, int nvqs,
 		status =
 		    virtqueue_create(dev, idx, (char *)names[idx], &ring_info,
 				     callbacks[idx], hil_vring_notify,
-				     rdev->proc->sh_buff.io,
 				     vqs[idx]);
 
 		virtqueue_set_shmem_io(vqs[idx], rdev->proc->sh_buff.io);
--- a/lib/rpmsg/remote_device.c
+++ b/lib/rpmsg/remote_device.c
@@ -428,11 +428,16 @@ int rpmsg_rdev_create_virtqueues(struct virtio_device *dev, int flags, int nvqs,
 				vring_table[idx].align));
 		}
 
+		vqs[idx] = virtqueue_allocate(ring_info.num_descs);
+		if (vqs[idx] == NULL) {
+			return (ERROR_NO_MEM);
+		}
+
 		status =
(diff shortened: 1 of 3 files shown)


lib/rpmsg/remote_device.c
```diff
@@ -428,11 +428,17 @@ int rpmsg_rdev_create_virtqueues(struct virtio_device *dev, int flags, int nvqs,
 				vring_table[idx].align));
 		}
 
+		vqs[idx] = virtqueue_allocate(ring_info.num_descs);
+		if (vqs[idx] == NULL) {
+			return (ERROR_NO_MEM);
```
skordal: I think this should be `return RPMSG_ERR_NO_MEM;` to be consistent with the namespace.

lib/virtio/virtqueue.c
```diff
@@ -44,49 +43,35 @@ int virtqueue_create(struct virtio_device *virt_dev, unsigned short id,
...
-		/* Initialize vring control block in virtqueue. */
-		vq_ring_init(vq, (void *)ring->vaddr, ring->align);
-
-		/* Disable callbacks - will be enabled by the application
-		 * once initialization is completed.
-		 */
-		virtqueue_disable_cb(vq);
-
-		*v_queue = vq;
+	if (vq == NULL) {
```
skordal: This could be changed to `VQ_PARAM_CHK(vq == NULL, status, ERROR_NO_MEM);`

lib/virtio/virtqueue.c
```diff
@@ -44,49 +43,35 @@ int virtqueue_create(struct virtio_device *virt_dev, unsigned short id,
...
+		return (ERROR_NO_MEM);
 	}
 
+	vq->vq_dev = virt_dev;
```
skordal: The remaining code in this function (all the way down to line 74) has to be within an `if (status == VQUEUE_SUCCESS)` block, due to the way `VQ_PARAM_CHK` is defined.

lib/rpmsg/remote_device.c
galak: Thanks will fix.

lib/virtio/virtqueue.c
galak: Will change.

lib/virtio/virtqueue.c
galak: Will fix.

galak:
Closing in favor of a new PR with all vq changes.

PR #67, opened 2018-04-18
Virtio changes

galak:


lib/include/openamp/virtio.h
```diff
@@ -111,6 +111,24 @@ struct virtio_buffer_info {
...
+ * @vq virtio queue
+ * @va logical address
+ * @notifyid vring notify id
+ * @num_descs number of descriptors
+ * @align vring alignment
+ * @io metal I/O region of the vring memory, can be NULL
+ */
+struct virtio_vring {
+	struct virtqueue *vq;
+	void *va;
```
galak: why void *, why not struct vring *va;

lib/include/openamp/virtio.h
```diff
@@ -111,6 +111,24 @@ struct virtio_buffer_info {
...
+	uint32_t notifyid;
+	unsigned int num_descs;
+	unsigned int align;
+	struct metal_io_region *io;
```
galak: what is this region for?

lib/include/openamp/virtio.h
```diff
@@ -111,6 +111,24 @@ struct virtio_buffer_info {
...
+ * @notifyid vring notify id
+ * @num_descs number of descriptors
+ * @align vring alignment
+ * @io metal I/O region of the vring memory, can be NULL
+ */
+struct virtio_vring {
+	struct virtqueue *vq;
+	void *va;
+	uint32_t notifyid;
+	unsigned int num_descs;
```
galak: Are num_descs & align really needed?  Seems like these are just init params - so maybe we need a virtio_init() or something

lib/include/openamp/virtio.h
```diff
@@ -111,6 +111,24 @@ struct virtio_buffer_info {
 	unsigned long size;
 };
 
+/**
+ * struct remoteproc_vring - remoteproc vring structure
```
galak: Seems like virtio_vring should go away, and just keep what we need in struct virtio_device directly.

lib/virtio/virtio.c
```diff
@@ -79,3 +79,44 @@ void virtio_describe(struct virtio_device *dev, const char *msg,
...
+
+int virtio_create_virtqueues(struct virtio_device *vdev, unsigned int flags,
+			     unsigned int nvqs, const char *names[],
+			     vq_callback *callbacks[])
+{
+	struct virtio_vring *rvring;
+	struct vring_alloc_info vring_info;
+	unsigned int num_vrings, i;
+	int ret;
+	(void)flags;
```
galak: if flags is void, lets remove it from the function as well.

lib/virtio/virtio.c
```diff
@@ -79,3 +79,44 @@ void virtio_describe(struct virtio_device *dev, const char *msg,
 	// TODO: Not used currently - keeping it for future use
 	virtio_feature_name(0, desc);
 }
+
+int virtio_create_virtqueues(struct virtio_device *vdev, unsigned int flags,
+			     unsigned int nvqs, const char *names[],
+			     vq_callback *callbacks[])
```
galak: should we be passing vring_info & vrings_num in, rather than holding onto them in the virtio_device?

lib/include/openamp/virtio.h
```diff
@@ -68,6 +68,7 @@ struct virtio_device_id {
 #define VIRTIO_TRANSPORT_F_END        32
 
 typedef struct _virtio_dispatch_ virtio_dispatch;
+typedef void (*vdev_reset_cb)(struct virtio_device *vdev);
```
galak: when is this suppose to be used or called?

Where are we holding on to it?

lib/include/openamp/virtio.h
```diff
@@ -182,7 +182,7 @@ struct _virtio_dispatch_ {
 	void (*write_config) (struct virtio_device * dev, uint32_t offset,
 			      void *src, int length);
 	void (*reset_device) (struct virtio_device * dev);
-
+	void (*notify) (struct virtqueue *vq);
```
galak: when is notify suppose to be called?

lib/virtio/virtio.c
galak: I was wondering if there was some value in having an API to create multiple queue's vs creating one queue at a time, and calling _create_virtqueue per queue that needs to be created.

lib/include/openamp/virtio.h
wjliang: this is used by the virtio host side application to get notified when the guest side got reset

lib/include/openamp/virtio.h
wjliang: this should not be in the virtio device, instead it should be in the virtqueue

galak:
I pulled the reset call back change into master-decouple-staging, with a minor cleanup

lib/include/openamp/virtio.h
wjliang: i rename this virtio_vring to vritio_vring_info, the va here is about the vrings area.
However, this struct is not aware of any specific data struct in the vrings area, this information will be passed to vq_ring_init() which will initialize the vring area.

lib/include/openamp/virtio.h
wjliang: vring area is also shared memory, and thus, it has I/O to abstract operation on the vring shared memory region.

lib/include/openamp/virtio.h
wjliang: it is used for vring initialization through virtqueue initialization. the resource table has this information. i am more prone to have this information and the virtqueue initialization which is the virtio_create_virqueues() function in the remoteproc_virtio in the staging branch.

lib/virtio/virtio.c
wjliang: the previous create_virtqeueus() is one API to create multiple queues. this patch tries to move it to the virtio

lib/virtio/virtio.c
wjliang: the flags here is for the features of the virtio user driver e.g. RPMsg, it just because we didn't implement that yet.
but maybe we should keep the argument in the API so that in future, when we we implement the feature, we don't need to change it.

galak: We should add a comment to the api that flags is for the future.

wjliang:
close this pull request, as they are covered in the latest pull requests

PR #69, opened 2018-04-19
remoteproc: expose the virtio dispatch table

skordal:
Exposes the remoteproc virtio function dispatch table through an extern declaration in the remoteproc_virtio.h header file, so it is available to applications.

The pointer to the function table must be set in the `virtio_device` struct before calling `rpmsg_init_vdev()`.

Alternatively, we could remove the `virtio_dispatch` structure and call the functions from it directly, which could reduce the overhead on the function calls. Is there a reason to keep the `virtio_dispatch` table?


lib/include/openamp/remoteproc_virtio.h
```diff
@@ -83,6 +83,8 @@ struct remoteproc_virtio {
 	struct virtio_device vdev;
 };
 
+extern virtio_dispatch remoteproc_virtio_dispatch_funcs;
```
arnop2: alternative would be to initialized it in rproc_virtio_create_vdev  func or too late? 
	vdev->reset_cb = rst_cb;
	vdev->vrings_num = num_vrings;
  +      vdev->func = remoteproc_virtio_dispatch_funcs;

PR #72, opened 2018-04-20
[RFC] Remove virtio_dispatch table

skordal:
As an alternative to PR#69, this PR proposes to remove the `virtio_dispatch function` table completely. This allows some wrapper functions to be removed, and many unnecessary indirections can be simplified to regular function calls.

This was developed on top of the changes from PR#58.


lib/include/openamp/remoteproc_virtio.h
```diff
@@ -66,6 +66,19 @@ struct remoteproc_virtio {
...
+uint32_t rproc_virtio_get_features(struct virtio_device *vdev);
+void rproc_virtio_set_features(struct virtio_device *vdev,
+			       uint32_t features);
+void rproc_virtio_read_config(struct virtio_device *vdev,
+			      uint32_t offset, void *dst, int length);
+void rproc_virtio_write_config(struct virtio_device *vdev,
+			       uint32_t offset, void *src, int length);
+void rproc_virtio_reset_device(struct virtio_device *vdev);
+
 /**
```
wjliang: Those functions are vdev operations, they will not be called directly by vdev client such as rpmsg virito, the rpmsg virtio will only call virtio functions

PR #77, opened 2018-04-26
A number of rpmsg related fixes

galak:
Please review, there are a few changes that I'm not sure if we need, and also not sure where we wanted to handle freeing of memory when we destroy endpoints.

### DIFF
--- a/lib/rpmsg/rpmsg_core.c
+++ b/lib/rpmsg/rpmsg_core.c
@@ -296,7 +296,7 @@ void rpmsg_rx_callback(struct virtqueue *vq)
 			/* Fatal error no endpoint for the given dst addr. */
 			return;
 
-		if (ept && ept->dest_addr == RPMSG_ADDR_ANY) {
+		if (ept && ept->dest_addr == RPMSG_ADDR_ANY && ept->addr != RPMSG_NS_EPT_ADDR) {
 			/*
 			 * First message received from the remote side,
 			 * update channel destination address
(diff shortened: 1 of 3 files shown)


lib/rpmsg/rpmsg_core.c
```diff
@@ -372,8 +372,10 @@ void rpmsg_ns_callback(struct rpmsg_endpoint *ept, void *data,
 			_ept->destroy_cb(_ept);
 
 		rpmsg_destroy_ept(_ept);
+#if 0
```
wjliang: The _ept has been freed in  the rpmsg_destroy_ept(), can just delete, instead of using "#if 0"

galak: I did this because we need to decide how we want to handle freeing of ept memory.  Since we have several ways in which the ept might be allocated:

1. static ept
2. ept allocated outside 
3. ept allocated by rpmsg_create_ept
4. ept explicilty by rpmsg

So we might not always want to free always with rpmsg_destroy_ept, thus doing the #if 0 for now.

wjliang: rpmsg_create_ept is for dynamic endpoints creation, in this case, it is safe to after the destroy_cb() is called.
Just static ept may not be freed. but just, if name service is used, i am not sure if there is case to use static endpoint in this situation.
But we can consider to have destroy_cb to have return value instead of void, if detroy_cb() returns non-zero, don't free endpoint, and we can have rpmsg_destroy_ept to have (free memory) allocation or not as one input argument.

galak:
Going to create a new PR w/latest tree so we can discuss this and other issues.

PR #78, opened 2018-05-03
Rpmsg decouple staging

arnop2:
New additional commits.
Main updates:
- split rpmsg_init_vdev in rpmsg_init_slave_vdev and rpmsg_init_smaster_vdev
- introduce VIRTIO_SLAVE_ONLY and VIRTIO_MASTER_ONLY compilation switch

ed27734 remoteproc_virtio: fixup test on uninitialized role  …
c52e89d  openamp: Add remoteproc virtio API 
bad6e2b  remoteproc virtio: clean up unused variables.
d663b40  virtio: optimize code
7d7530d  pmsg: split vdev initialisation for master and slave
66b4ec0 openamp: introduce compilation switch to optimize code


lib/rpmsg/rpmsg.c
```diff
@@ -39,6 +39,7 @@ static int rpmsg_get_buffer_size(struct rpmsg_virtio_device *rvdev)
 
 	metal_mutex_acquire(&rvdev->lock);
 
+#if !defined VIRTIO_SLAVE_ONLY  && !defined  VIRTIO_MASTER_ONLY
```
wjliang: How about VIRTIO_MASTER_DISABLE and VIRTIO_SLAVE_DISABLE?

lib/rpmsg/rpmsg_core.c
```diff
@@ -174,17 +174,28 @@ void *rpmsg_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
...
 		if (data == RPMSG_NULL) {
 			data = sh_mem_get_buffer(rvdev->shbuf);
 			*len = RPMSG_BUFFER_SIZE;
 		}
 	} else {
-		data =
-		    virtqueue_get_available_buffer(rvdev->svq, idx,
-						   (uint32_t *)len);
+		data = virtqueue_get_available_buffer(rvdev->svq, idx,
+						      (uint32_t *)len);
```
wjliang: What's the difference?

lib/rpmsg/rpmsg_core.c
```diff
@@ -174,17 +174,28 @@ void *rpmsg_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
...
+#if defined VIRTIO_MASTER_ONLY
+	data = virtqueue_get_buffer(rvdev->svq, (uint32_t *)len, idx);
+	if (data == RPMSG_NULL) {
+		data = sh_mem_get_buffer(rvdev->shbuf);
+		*len = RPMSG_BUFFER_SIZE;
+	}
+#else
+	data = virtqueue_get_available_buffer(rvdev->svq, idx, (uint32_t *)len);
+#endif
+#endif
```
wjliang: we have three implementations here, i think we should just have one implementation, and allows part of it disabled if the feature is disabled

lib/include/openamp/virtio.h
```diff
@@ -121,7 +121,6 @@ struct virtio_device {
 	uint32_t index; /**< unique position on the virtio bus */
 	struct virtio_device_id id; /**< the device type identification
 				         (used to match it with a driver). */
-	struct metal_spinlock lock; /**< spin lock */
```
wjliang: if we remove this lock how can we guarantee there is race condition in accessing the vrings? I think we should make use of the lock in the virtio access

lib/include/openamp/virtio.h
```diff
@@ -105,10 +105,8 @@ struct virtio_buffer_info {
  */
 struct virtio_vring_info {
 	struct virtqueue *vq;
-	void *va;
+	struct vring_alloc_info info;
```
wjliang: how about just remove vring_alloc_info? and just pass virtio_vring_info to virtqueue_create?

lib/virtio/virtio.c
```diff
@@ -96,22 +96,19 @@ int virtio_create_virtqueues(struct virtio_device *vdev, unsigned int flags,
...
-			unsigned int align = vring_info->align;
 
-			offset = metal_io_virt_to_offset(io, vring_info->va);
+			offset = metal_io_virt_to_offset(io, vring_alloc->vaddr);
 			metal_io_block_set(io, offset, 0,
-					   vring_size(num_descs, align));
+					   vring_size(vring_alloc->num_descs,
+						      vring_alloc->align));
 		}
 		ret = virtqueue_create(vdev, i,
```
wjliang: how about just pass virtio_vring_info to virtqueue_create? and remove the requirement of vring_alloc_info?

lib/rpmsg/rpmsg.c
```diff
@@ -251,57 +251,45 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept)
...
 {
 	const char *vq_names[RPMSG_NUM_VRINGS];
 	void (*callback[RPMSG_NUM_VRINGS]) (struct virtqueue *vq);
 	unsigned long dev_features;
 	static struct rpmsg_endpoint ns_ept;
+	struct virtqueue_buf vqbuf;
+	void * buffer;
 	int status;
-	unsigned int i;
+	unsigned int i, idx;
```
wjliang: how about just "i"?

lib/rpmsg/rpmsg.c
```diff
@@ -316,38 +304,113 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+	 * Create name service announcement endpoint if device supports name
+	 * service announcement feature.
+	 */
+	if ((dev_features & (1 << VIRTIO_RPMSG_F_NS))) {
+		ns_ept.addr = RPMSG_NS_EPT_ADDR;
+		ns_ept.cb = rpmsg_ns_callback;
+		strncpy(ns_ept.name, "NS", sizeof("NS"));
+		status = rpmsg_register_endpoint(rvdev, &ns_ept);
+
+	}
```
wjliang: check virtio features and endpoint creation is the common, i think can have a common function for them

lib/include/openamp/virtio.h
arnop2: lock is not used in implementation. A lock is already handled by rpmsg client. Do you think we need both?

lib/include/openamp/virtio.h
arnop2: yes, i hesitated between both solutions.  Drawback would that either virqueue should include virtio.h or this structure should migrate to virtqueue.h

lib/rpmsg/rpmsg.c
arnop2: yes,  no sense to re-declare a local variable, i will update this

lib/rpmsg/rpmsg.c
arnop2: i will have a look to this,

lib/rpmsg/rpmsg_core.c
wjliang: agree to line up with Linux coding style as long as it is not conflict to MISRA-C

lib/include/openamp/virtio.h
wjliang: when we use virtio, shall we always assume that who uses the virtio will make sure no race condition, maybe this is the right assumption

wjliang:
commit "remoteproc virtio: clean up unused variables."  has been applied to wjliang/deouple-virtio

galak:
I pushed and update against the last tree/work I've done.  I'm closing this so we can add comments on a new PR.

PR #79, opened 2018-05-09
[RFC] Staging - latest work on 'master-decouple-staging'

galak:


lib/rpmsg/rpmsg.c
```diff
@@ -365,7 +365,11 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 		ns_ept.cb = rpmsg_ns_callback;
 		strncpy(ns_ept.name, "NS", sizeof("NS"));
 		status = rpmsg_register_endpoint(rvdev, &ns_ept);
+
 	}
+	/* check if there is message sent from master */
+	if (rpmsg_virtio_get_role(rvdev) == RPMSG_REMOTE)
```
galak: Curious why we needed this?

lib/rpmsg/rpmsg_core.c
```diff
@@ -352,8 +352,10 @@ void rpmsg_ns_callback(struct rpmsg_endpoint *ept, void *data,
 			_ept->destroy_cb(_ept);
 
 		rpmsg_destroy_ept(_ept);
+#if 0
```
galak: We need to decide how we want to handle this.  Do we need to keep track of how and ept was created so we can decide how/if to destroy it?

lib/rpmsg/rpmsg_core.c
```diff
@@ -375,7 +377,7 @@ void rpmsg_ns_callback(struct rpmsg_endpoint *ept, void *data,
 			if (rvdev->new_endpoint_cb)
 				rvdev->new_endpoint_cb(_ept);
 		}
-		_ept->dest_addr = src;
+		_ept->dest_addr = ns_msg->addr;
```
galak: Should we move this earlier so when we call the callback its already set?

lib/rpmsg/rpmsg_core.c
```diff
@@ -276,7 +276,7 @@ void rpmsg_rx_callback(struct virtqueue *vq)
 			/* Fatal error no endpoint for the given dst addr. */
 			return;
 
-		if (ept && ept->dest_addr == RPMSG_ADDR_ANY) {
+		if (ept && ept->dest_addr == RPMSG_ADDR_ANY && ept->addr != RPMSG_NS_EPT_ADDR) {
```
galak: @arnop2 I don't think this is needed anymore due to us always calling the cb now.  While I don't think it hurts, do we need this if check at all anymore?

lib/rpmsg/rpmsg_core.c
```diff
@@ -239,36 +239,7 @@ int rpmsg_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
...
-
-			dev_features = rpmsg_virtio_get_features(rvdev);
-			if ((dev_features & (1 << VIRTIO_RPMSG_F_NS))) {
-				if (rpmsg_send_ns_message(rvdev, ept,
-					RPMSG_NS_CREATE) != RPMSG_SUCCESS)
-					return;
-			}
-		}
-	}
+	(void)vq;
```
galak: @wjliang Why is it that we don't need to do anything on tx?

lib/rpmsg/rpmsg.c
wjliang: if we split slave vdev and master vdev, we don't need this checking, but how to split is still under discussion

galak: Sorry, its not the checking I was asking about.  I was asking why we need to call rpmsg_rx_callback.

lib/rpmsg/rpmsg.c
```diff
@@ -110,5 +110,6 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept)
 		return;
 
 	rpmsg_unregister_endpoint(ept);
-	metal_free_memory(ept);
+	if (ept->destroy_cb)
+		ept->destroy_cb(ept);
```
arnop2: As rpmsg_create_ept dynamically allocated the ept,  the memory free seems missing for the destroy...
The result of the patch is that application has to free an ept that it does not create.
My opinion is that this is not coherent. Either application should  allocate and free ept structure or rpmsg should do it.

galak: Good point, I don't think I realized that rpmsg_create_ept allocated and ept.  Since that's the case will look to clean this up so we do the freeing as well.

wjliang: it depends how to use the  destroy_endpoint, if we allow endpoint to be statically created, which i think yes, destroy endpoint should not do free_endpoint.
I think we can document the pair of the functions.

PR #81, opened 2018-05-09
Virtqueue updates

galak:


lib/include/openamp/virtqueue.h
```diff
@@ -79,8 +79,7 @@ struct virtqueue {
 	struct vring vq_ring;
 	uint16_t vq_free_cnt;
 	uint16_t vq_queued_cnt;
-	/** Shared memory I/O region */
-	struct metal_io_region *shm_io;
+	void * shm_io; /* opaque pointer to data needed to allow v2p & p2v */
```
wjliang: Prefer not to use opaque pointer here. For I/O optimisation, prefer do that in the libmetal

lib/include/openamp/virtqueue.h
```diff
@@ -64,6 +63,11 @@ typedef enum {
 	VQ_POSTPONE_EMPTIED	/* Until all available desc are used. */
 } vq_postpone_t;
 
+struct virtqueue_buf {
+	void *buf;
+	int len;
+};
+
```
wjliang: Just looks like buffer/memory access should be out of virtqueue.

lib/include/openamp/virtqueue.h
galak: The reason its void * here is to keep the user from directly access shm_io.  All the internal uses inside virtqueue are treated as a metal_io_region pointer.

PR #91, opened 2018-05-24
[RFC] Decouple virtio -- remove sh_mem and other rpmsg virtio cleanup

wjliang:
* use metal bitmap helper function instead of creating rpmsg bitmap implementation
* remove sh_mem file
   * Allow user to implement their own shared buffer allocator
* clean up rpmsg send_raw function to reduce code redundancy


lib/include/openamp/rpmsg_virtio.h
```diff
@@ -29,16 +29,32 @@ extern "C" {
...
+ */
+struct rpmsg_virtio_shm_pool {
+	void *(*get_buffer)(struct rpmsg_virtio_shm_pool *shpool, size_t size);
+	void *base;
+	size_t avail;
+	size_t size;
+};
+
+/* Default rpmsg virtio shared memory pool */
+extern struct rpmsg_virtio_shm_pool rpmsg_virtio_default_shm_pool;
```
arnop2: This means that you define a common pool for all communications. So if  you have to communicate with 2 separate processors , this impose that in your ssytem  the shared memory is shared between all processors...
=> Need to define this pool in rpmsg_virtio_device struct not as global variable. Application should ensure the coherence if need to share the same memory for several rpmsg IPC links.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -23,6 +23,27 @@
 /* Time to wait - In multiple of 10 msecs. */
 #define RPMSG_TICKS_PER_INTERVAL                10
 
+static void *
+rpmsg_virtio_shm_pool_get_buffer(struct rpmsg_virtio_shm_pool *shpool,
+				 size_t size)
+{
+	void *buffer;
+
+	if(shpool->avail < size)
```
arnop2: what about creating a generic rpmsg_dev_ops structure with associted  APi register some app ops that will modify default behaviour?
rpmsg_virtio_shm_pool_get_buffer should be part of this ops structure an we could simply extend ops on need in future (e.g.  log).
Then in rpmsg_virtio_shm_pool_get_buffer we would have 
if (ops->get_buffer)
      ops->get_buffer (...);
else
   /* do the default action */

lib/include/openamp/rpmsg_virtio.h
wjliang: good point. cannnot have global variable as they are not just operations.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -23,6 +23,35 @@
...
+void rpmsg_virtio_init_shm_pool(struct rpmsg_virtio_shm_pool *shpool,
+				void *shb, size_t size)
+{
+	if (!shpool)
+		return;
+	shpool->base = shb;
+	shpool->size = WORD_ALIGN(size);
+	shpool->avail = WORD_ALIGN(size);
+	shpool->get_buffer = rpmsg_virtio_shm_pool_get_buffer;
+}
```
arnop2: what about  a simple macro?
application declare it with the macro:
struct rpmsg_virtio_shm_pool shpool = RPMSG_VIRTIO_INIT_SHM_POOL( *shmem, size, NULL);

else you need to return a status on error.

lib/rpmsg/rpmsg_virtio.c
arnop2: my remark here is more to introduce ops functions ( similar to kernel frameworks that offer ops for drivers that want a specific management of a service). Instead of updating API each  time we need a specific treatment by application, we can define ops structures that list all this kind of pointers. Perhaps we need rpmsg ops , rpmsg_virtio ops, remote proc ops and libmetal ops strutures...

wjliang: i agree with more ops, but just whether it should be rpmsg_virtio ops, rpmsg ops or shared buffered ops itself. it is more closer to rpmsg_virtio, however, i more inclined to shared buffers itself, as the shared buffers ops needs its own private data. And thus, i wrap the shared buffer ops + its data into the shared buffer pool struct

PR #94, opened 2018-05-28
introduce VIRTIO_MASTER_ONLY and VIRTIO_SLAVE_ONLY compilation flag

arnop2:
Tested in zephyr environment with VIRTIO_SLAVE_ONLY  enabled
Only compilation test with VIRTIO_MASTER_ONLY  enabled


lib/remoteproc/remoteproc_virtio.c
```diff
@@ -151,17 +156,20 @@ static void rproc_virtio_reset_device(struct virtio_device *vdev)
...
+#endif
 
 virtio_dispatch remoteproc_virtio_dispatch_funcs = {
 	.get_status =  rproc_virtio_get_status,
-	.set_status = rproc_virtio_set_status,
 	.get_features = rproc_virtio_get_features,
+	.read_config = rproc_virtio_read_config,
+	.notify = rproc_virtio_virtqueue_notify,
+#ifndef VIRTIO_SLAVE_ONLY
+	.set_status = rproc_virtio_set_status,
```
wjliang: Referring to Linux kernel include/uapi/linux/virtio_config.h, both slave and master can set status

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -151,17 +156,20 @@ static void rproc_virtio_reset_device(struct virtio_device *vdev)
...
 	.set_features = rproc_virtio_set_features,
 	.negotiate_features = rproc_virtio_negotiate_features,
-	.read_config = rproc_virtio_read_config,
 	.write_config = rproc_virtio_write_config,
 	.reset_device = rproc_virtio_reset_device,
-	.notify = rproc_virtio_virtqueue_notify,
+#endif
```
wjliang: As the virtio config space is supposed to be used for negotiation. As we also want to allow virtio to be used independent to rpmsg, for now, i think we can keep those functions for both slave and master

PR #108, opened 2018-06-13
Decoupling staging

arnop2:


galak:
I dropped

rpmsg_virtio: add check on end point callback return.
rpmsg: cleanup is_rpmsg_ept_ready

From the branch as I wanted to comment on them.

PR #109, opened 2018-06-13
OpenAMP Apps Restruct

wjliang:
* Split OpenAMP apps into tests and examples
* Hide remoteproc from RPMsg tests main implementation.
v2:
* copy echo_test to apps/examples/echo
* copy echo_test to apps/tests/messaging, will need to update this test to cover more messaging verification. e.g. remote update the message after it receives the old message. Flood ping tests
* rename the func test suite into rpmsg apis test. will need to change this test to test the updated RPMsg APIs in later commits


apps/machine/zynqmp/platform_info.c
```diff
@@ -94,6 +94,10 @@ struct remoteproc_priv rproc_priv = {
 
 static struct remoteproc rproc_inst;
 
+/* External functions */
+extern int init_system(void);
```
galak: Are these still used?  Should we rename to platform_init/platform_cleanup

wjliang: > Are these still used? Should we rename to platform_init/platform_cleanup

This is internal the platform implementation, the reason we do this is because in our case, we have common in implementation for the same platform across different systems, but each systems needs different systems specific initialization.

wjliang:
> for the split of apps into examples/test, are we going to rename func_test_suite? Should echo_test really be in samples, and we create a new msg_test in tests/?

the function test suite is more for verifying the APIs themselves, we can use loop back function for it.
And then we should have another application which will verify the communication between two processors. In this case, today, we only have echo_test for this purpose, we can update the echo_test to do more advanced checking later.

PR #111, opened 2018-06-14
rpmsg: Remove RPMSG_EPT_CB_* and use just RPMSG_SUCCESS

galak:
Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


wjliang:
merged

PR #112, opened 2018-06-14
Decoupling staging

arnop2:


lib/include/openamp/rpmsg_virtio.h
```diff
@@ -143,22 +144,37 @@ void rpmsg_deinit_vdev(struct rpmsg_virtio_device *rvdev);
...
 }
 
+/**
+ * rpmsg_virtio_shm_pool_get_buffer - get buffer in the shared memory pool
+ *
+ * RPMsg virtio has default shared buffers pool implementation.
+ * The memory assigned to this pool will be dedicated to the RPMsg
+ * virtio. If you prefer to have other shared buffers allocation,
+ * you can implement your rpmsg_virtio_shm_pool_get_buffer function.
+ *
```
galak: We should add something here like:

If you implement your own rpmsg_virtio_shm_pool_get_buffer the expectation is that one would embedded struct rpmsg_virtio_shm_pool object inside your own struct and utilize something like metal_container_of() to reference back to your own struct in your implementation of rpmsg_virtio_shm_pool_get_buffer

arnop2:
the pull request as been updated with the V2

PR #116, opened 2018-07-11
Master decouple staging: fixes up for endpoint destruction

arnop2:


lib/rpmsg/rpmsg.c
```diff
@@ -264,7 +264,8 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept)
 		return;
 
 	rdev = ept->rdev;
-	(void)rpmsg_send_ns_message(ept, RPMSG_NS_DESTROY);
+	if(ept->name && strncmp(ept->name, "NS", sizeof("NS")))
```
wjliang: instead of compare the name, how about check  the endpoint address

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -478,10 +478,11 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	_ept = rpmsg_get_endpoint(rdev, name, RPMSG_ADDR_ANY, dest);
 
 	if (ns_msg->flags & RPMSG_NS_DESTROY) {
-		if (_ept) {
-			metal_mutex_release(&rdev->lock);
-			rpmsg_destroy_ept(_ept);
-		}
+		if (_ept)
+			_ept->dest_addr = RPMSG_ADDR_ANY;
```
wjliang: why setting the dest_addr to ANY?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -478,10 +478,11 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
...
+		metal_mutex_release(&rdev->lock);
+		if (_ept && _ept->destroy_cb)
+				_ept->destroy_cb(ept);
```
wjliang: the destroyed callback is to used to notify application the endpoint has been destroyed, better to update the name service bind endpoint callback to also handle destroy callback by adding argument to indicate if it is destroy or announcement. but the rpmsg_virtio_ns_callback should still have default behavior if no such callback is defined by application.

lib/rpmsg/rpmsg_virtio.c
arnop2: if application don't destroy the endpoint, this allows to reconnect the end point on a new  ns announcement ( for instance crash usecase)

lib/rpmsg/rpmsg.c
arnop2: yes will change this

lib/rpmsg/rpmsg_virtio.c
arnop2: i will do a patch in this way

arnop2: I reanalyses the point. We have following use cases:
- we receive NS_DESTROY message:  
  => call a unbind call back to inform application that the remote endpoint is destroyed. apllication can take appropriate action (suspend communication for this service, destroy the local endpoint...)
- the local endpoint is destroyed. 
=> only use case is the rpmsg_deinit_vdev call. In this case the rpmsg dev is released. I think there is no reason to call a destroy callback during the deinit; application should ensure that all the communications are stopped before calling the denit . So we could just clean the callback call in  rpmsg_deinit_vdev.

In conclusion, i will propose a patch that :
- rename "destroy_cb" in "unbind_cb"
- suppress the callback call in rpmsg_deinit_vdev.

lib/rpmsg/rpmsg.c
```diff
@@ -150,24 +150,26 @@ struct rpmsg_endpoint *rpmsg_get_endpoint(struct rpmsg_device *rdev,
...
 			name_match = !strncmp(ept->name, name,
 					      sizeof(ept->name));
-		if (dest_addr == RPMSG_ADDR_ANY &&
-		    ept->addr == RPMSG_ADDR_ANY &&
-		    name_match)
-			return ept;
-		if (addr == RPMSG_ADDR_ANY &&
-		    ept->dest_addr == RPMSG_ADDR_ANY &&
-		    name && name_match)
+		if (!name || !name_match)
```
wjliang: looks like name_match may be used as not initialized. As in the previous line name_match is only set when name is not NULL.

lib/include/openamp/rpmsg.h
```diff
@@ -72,7 +73,7 @@ struct rpmsg_endpoint {
 	uint32_t addr;
 	uint32_t dest_addr;
 	rpmsg_ept_cb cb;
-	rpmsg_ept_destroy_cb destroy_cb;
+	rpmsg_ns_unbind_cb unbind_cb;
```
wjliang: I think the destroy callback is used in the destroy endpoint function call, as the destroy endpoint function can be called internally, it will be good to have a callback to notify user the endpoint has been destroyed, the "unbind" endpoint is used for ns destroy name service. I think it can be merged into the rpmsg_ns_bind_cb maybe rename rpmsg_ns_bind_cb to rpmsg_ns_cb()?

lib/rpmsg/rpmsg.c
arnop2: No issue as name is tested first.  so test here is equivalent to
if( !name) 
     continue
else if (!name_match)
     continue

lib/include/openamp/rpmsg.h
arnop2: Please have a look to the commit message, from my POV destroy_cb is not useful.
Furthermore  we need a callback per endpoint for the unbind, not a global call back, has endpoints can be handled by different users.

lib/rpmsg/rpmsg.c
```diff
@@ -268,6 +268,4 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept)
 	metal_mutex_acquire(&rdev->lock);
 	rpmsg_unregister_endpoint(ept);
 	metal_mutex_release(&rdev->lock);
-	if (ept->destroy_cb)
-		ept->destroy_cb(ept);
 }
```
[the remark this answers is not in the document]
arnop2: Could you share a use case where the callback is needed? From my POV  this function is called in 2 cases:
1) by user, to destroy the end point,  cb not needed in this case.
2) user calls rpmsg_deinit_vdev, in this case application should free all its resources after the function call, no need of a specific callback.

lib/rpmsg/rpmsg.c
wjliang: you're right. Should be fine. Just not sure if there will be compilation warning. Will know when I compile with the patch.

lib/include/openamp/rpmsg.h
wjliang: if replace unbind_cb, what should we do with destroy_endpoint()? and who should call destroy_endpoint?
in the destroy_endpoint, we should call unbind_cb() first? and then unregister the endpoint if user want to unregister?
How does user application descides when to remove the endpoint or keep it?
or always assume the unbind_cb() is only called from NS destroy? in this case, should the callback name be more clearer so that user knows it is only called when the NS destroy is received?

arnop2: >if replace unbind_cb, what should we do with destroy_endpoint()? and who should call >destroy_endpoint?
rpmsg_destroy_ept should be called by application when stop the service, to unregister the local end point and send associated NS_DESTROY message.  

>in the destroy_endpoint, we should call unbind_cb() first? and then unregister the endpoint if >user want to unregister?
No, unbind_cb is used only to inform application that remote end point is destroyed. So unbind_cb() can call rpmsg_destroy_ept to desrtroy the local one and free the associated memory ( dynamic allocation case).

>How does user application descides when to remove the endpoint or keep it?
>or always assume the unbind_cb() is only called from NS destroy?
Exactly, unbind_cb() only called on NS_DESTROY to inform that local endpoint is no more bound to the remote one( that is destroyed)

> in this case, should the callback name be more clearer so that user knows it is only called when the NS destroy is received?
What about "ns_unbind_cb" that is symetric with the "ns_bind_cb" defined in  rpmsg_device struct?

lib/rpmsg/rpmsg.c
wjliang: when I compile it i did got warning saying that it may be used uninitialized, i will change to assign default value to 0 when declaring the variable.

apps/examples/echo/rpmsg-ping.c
```diff
@@ -49,7 +49,7 @@ static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
 	(void)ept;
 	(void)src;
 	(void)priv;
-	LPRINTF(" received payload number %lu of size %d \r\n",
+	LPRINTF(" received payload number %lu of size %ld \r\n",
```
wjliang: it doesn't work for 32bit system. i will make the change.

wjliang:
merged with modification

PR #117, opened 2018-07-13
separate hil_enalbe_vring_notifications

tyokota-eforce:
Hi. all
I used ISR (not polling) based remoteproc dirver on our RTOS.
In case an rpu application was loaded by Linux remoteproc and rpmsg_user_dev_driver 
on ZynqMP, ISR is called before saving second vring handle when first vring is enabling 
notfications, and then hil_notified is refer NULL pointer vring by called ISR. 

I need separate this function into two, set vring handle and enable notification.

### DIFF
--- a/lib/common/hil.c
+++ b/lib/common/hil.c
@@ -418,15 +418,32 @@ int hil_enable_vdev_notification(struct hil_proc *proc, int id)
 }
 
 /**
- * hil_enable_vring_notifications()
+ * hil_save_vring_handle()
  *
  * This function is called after successful creation of virtqueues.
  * This function saves queue handle in the vring_info_table which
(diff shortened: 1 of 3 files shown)


wjliang:
It looks like this patch is a workaround. I will take this patch for master. But virtio itself doesn't rely on interrupt. We are working on master-decouple-staging. Whenever we create an endpoint, we send out name service announcement instead of relying on the first kick. You can take a look at this implementation: https://github.com/OpenAMP/open-amp/blob/master-decouple-staging/lib/rpmsg/rpmsg.c#L200

tyokota-eforce:
That's right. It's a workaroud for master.  
Thank you introducing master-decouple-staging version, I'll check and try this version.

PR #119, opened 2018-07-24
Master decouple staging with event controller

bentheredonethat:
update with libmetal events controller


wjliang:
This change is based on the libmetal RFC request. will close this pull request for now. as the master-decouple-staging is merged to master. Please change the pull request to targeting master when the libmetal changes are in.

PR #136, opened 2018-09-24
remoteproc: virtio: add service to get vdev status.

arnop2:
On slave side, during the init sequence we need to wait
the vdev status before calling rproc_virtio_init_vring.
Indeed the virtual address(VA), provided as function
parameter, is based on the device address(DA).
Depending on the environment this DA is retrieved from
the resource table, filled by the master processor.
this service allow to block the processor waiting the
status update.

### DIFF
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -309,3 +309,22 @@ int rproc_virtio_notified(struct virtio_device *vdev, uint32_t notifyid)
 	}
 	return 0;
 }
+
+void rproc_virtio_wait_remote_ready(struct virtio_device *vdev)
+{
+	uint8_t status;
+
+	/*
+	 * No status available for slave. As Master has not to wait
+	 * slave action, we can return. Behavior should be updated
+	 * in future if a slave status is added.
+	 */
+	if (vdev->role == VIRTIO_DEV_MASTER)
+		return;
+
+	while (1) {
+		status = rproc_virtio_get_status(vdev);
+		if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK)
+			return;
+	}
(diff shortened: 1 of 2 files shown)


lib/remoteproc/remoteproc_virtio.c
```diff
@@ -309,3 +309,14 @@ int rproc_virtio_notified(struct virtio_device *vdev, uint32_t notifyid)
 	}
 	return 0;
 }
+
+void rproc_virtio_wait_remote_ready(struct virtio_device *vdev)
+{
+	uint8_t status;
+
+	while (1) {
+		status = rproc_virtio_get_status(vdev);
```
wjliang: This checking is only available to virtio slave, please add "if (vdev->role == VIRTIO_DEV_SLAVE)"
if it is MASTER, for not, always return.
There can be future improvement to allow non-busy-looping wait, we can define default poll in libmetal for the platform, which will do nothing, and allow application to overwrite it, and in this wait function, we call the platform poll function.

arnop2: Concerning the test: this was my first implementation. But rproc_virtio_get_status is available for the slave and the master. So it could be available for both (no use case in mind for a usage on master side). 
Please tell me your preference.

Concerning the wait: not sure to understand... do you suggest to not implement the loop here but only provide access to the status? in this case an alternative could be to export the rproc_virtio_get_status function instead of implementing this new service.

wjliang: I should have put my comments one line after, the patch is checking status bit DRIVER_OK which will be set by master, and now, slave is not able to tell its status unless we update the rsc table.
About waiting, for now, it is busy wait, but it can be future improved to call platform wait if the status bit is not set

PR #140, opened 2018-12-04
rpmsg: fix virtio set_status usage of vdev status for slave

arnopo:
Today only one vdev status is available for the processors
status. To avoid overwrite, this status has to be updated
only by the master.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -10,7 +10,7 @@ names to be CC'd when submitting a patch.
 
 ## Project Administration
 Wendy Liang <wendy.liang@xilinx.com>
-Ed T. Mooring <emooring@xilinx.com>
+Ed T. Mooring <ed.mooring@linaro.org>
 Arnaud Pouliquen <arnaud.pouliquen@st.com>
 
 ### All patches CC here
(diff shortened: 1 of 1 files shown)


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -223,10 +223,14 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
...
 		/* Busy wait until the remote is ready */
+#ifndef VIRTIO_SLAVE_ONLY
 		if (status & VIRTIO_CONFIG_STATUS_NEEDS_RESET) {
 			rpmsg_virtio_set_status(rvdev, 0);
 			/* TODO notify remote processor */
-		} else if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK) {
+			continue;
+		}
+#endif /*!VIRTIO_MASTER_ONLY*/
+		if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK) {
```
wjliang: should it be "!VIRTIO_SLAVE_ONLY" ?

wjliang: Should this driver ok checking only apply to virtio slave?

arnop2: oops, need fix this copy/past , thanks

arnop2: For me both, the slave and the master can call it, but now regarding the implementation the function is only called by slave in rpmsg_init_vdev . 
So perhaps a best solution is  to simply suppress this piece of code.

wjliang: On the slave side, if the master hasn't set the driver_ok, the slave side cannot finish the init as it will need to use the vrings. Agree both sides can check the status, but as the status is set by master for now, and driver_ok bit can only set by master, only slave need to check the bit.

arnop2:
rpmsg_virtio_wait_remote_ready is only call by rpmsg_init_vdev by remote. So just clean the rpmsg_virtio_set_status that is only updated by the master.

PR #145, opened 2019-01-04
replace all OPENAMP_PACKED_ to METAL_PACKED_

xiaoxiang781216:


wjliang:
> I think we should isolate the use of libmetal in openamp as much as possible, so rather than replacing OPENAMP_PACKED_* in the code, we should keep that and just having compiler.h look like:

Just as libmetal provides a layer for OpenAMP to enable the library to be used across different environment, this compiler specific macro, I agree with @xiaoxiang781216 that it should be placed in libmetal. As the OpenAMP library itself better to be environment independent.

PR #146, opened 2019-01-04
Minor fix in the code review

xiaoxiang781216:
@wjliang, the change in each patch is very small, please review them.


lib/proxy/rpmsg_retarget.c
```diff
@@ -154,7 +154,7 @@ void rpmsg_set_default_rpc(struct rpmsg_rpc_data *rpc)
  *       Open a file.  Minimal implementation
  *
  *************************************************************************/
-#define MAX_BUF_LEN 496UL
+#define MAX_BUF_LEN 496
```
galak: I feel like this might because of MISRA rules

xiaoxiang781216:
Abandon this PR, replace with the small PR series.

PR #147, opened 2019-01-05
Handle interrupt suppression correctly

xiaoxiang781216:
@wjliang, make the interrupt suppression like Linux behavior.

### DIFF
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -63,7 +63,7 @@ static void rproc_virtio_set_status(struct virtio_device *vdev,
 }
 #endif
 
-static uint32_t rproc_virtio_get_features(struct virtio_device *vdev)
+static uint32_t rproc_virtio_get_dfeatures(struct virtio_device *vdev)
 {
 	struct remoteproc_virtio *rpvdev;
 	struct fw_rsc_vdev *vdev_rsc;
@@ -73,11 +73,28 @@ static uint32_t rproc_virtio_get_features(struct virtio_device *vdev)
 	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
 	vdev_rsc = rpvdev->vdev_rsc;
 	io = rpvdev->vdev_rsc_io;
-	/* TODO: shall we get features based on the role ? */
 	features = metal_io_read32(io,
 			metal_io_virt_to_offset(io, &vdev_rsc->dfeatures));
 
 	return features;
+}
+
+static uint32_t rproc_virtio_get_features(struct virtio_device *vdev)
+{
+	struct remoteproc_virtio *rpvdev;
+	struct fw_rsc_vdev *vdev_rsc;
+	struct metal_io_region *io;
+	uint32_t gfeatures;
+	uint32_t dfeatures;
+
+	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
+	vdev_rsc = rpvdev->vdev_rsc;
+	io = rpvdev->vdev_rsc_io;
+	gfeatures = metal_io_read32(io,
+			metal_io_virt_to_offset(io, &vdev_rsc->gfeatures));
+	dfeatures = rproc_virtio_get_dfeatures(vdev);
+
+	return dfeatures & gfeatures;
 }
 
 #ifndef VIRTIO_SLAVE_ONLY
@@ -91,22 +108,20 @@ static void rproc_virtio_set_features(struct virtio_device *vdev,
 	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
 	vdev_rsc = rpvdev->vdev_rsc;
 	io = rpvdev->vdev_rsc_io;
-	/* TODO: shall we set features based on the role ? */
 	metal_io_write32(io,
-			 metal_io_virt_to_offset(io, &vdev_rsc->dfeatures),
+			 metal_io_virt_to_offset(io, &vdev_rsc->gfeatures),
 			 features);
 	rpvdev->notify(rpvdev->priv, vdev->notifyid);
 }
-#endif
 
 static uint32_t rproc_virtio_negotiate_features(struct virtio_device *vdev,
 						uint32_t features)
 {
-	(void)vdev;
-	(void)features;
-
+	uint32_t dfeatures = rproc_virtio_get_dfeatures(vdev);
+	rproc_virtio_set_features(vdev, dfeatures & features);
 	return 0;
 }
+#endif
 
 static void rproc_virtio_read_config(struct virtio_device *vdev,
 				     uint32_t offset, void *dst, int length)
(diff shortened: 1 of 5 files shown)


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -638,13 +637,11 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
 
-	dev_features = rpmsg_virtio_get_features(rvdev);
-
 	/*
 	 * Create name service announcement endpoint if device supports name
 	 * service announcement feature.
 	 */
-	if ((dev_features & (1 << VIRTIO_RPMSG_F_NS))) {
+	if (vdev->features & (1 << VIRTIO_RPMSG_F_NS)) {
 		rpmsg_init_ept(&rdev->ns_ept, "NS",
```
wjliang: This modification looks good, just I cannot see where did you set the vdev->features?
From, master, it should set the features before setting the driver_OK, just I didn't see the modification, looks like it should be done in the negotiate_features() API, for slave, it should set after it checks DRIVER_OK. Looks like it needs an virtio API abstraction for it. Or we should have the negotiate_features() API for both roles, but for devices, it will not set the virtio config space, but instead, it should set the features locally, that is the vdev->features.

xiaoxiang781216: > This modification looks good, just I cannot see where did you set the vdev->features?
In rproc_virtio_create_vdev, I add one line to set it:
vdev->features = rproc_virtio_get_features(vdev);

> From, master, it should set the features before setting the driver_OK, just I didn't see the modification, looks like it should be done in the negotiate_features() API,
From virtio spec, master should:
1.Read dfeature field from resource table
2.Mask the feature which doesn't support
3.Write the result to gfeature field
The current implementation doesn't confirm the spec, do you want to fully implement the negotiation process.

>for slave, it should set after it checks DRIVER_OK. Looks like it needs an virtio API abstraction for it. Or we should have the negotiate_features() API for both roles, but for devices, it will not set the virtio config space, but instead, it should set the features locally, that is the vdev->features.
No, the slave side shouldn't set any field in config space, instead it should get the negotiated result from gfeature directly.

xiaoxiang781216: @wjliang I add new patch to negotiate the feature, please review.

lib/virtio/virtqueue.c
```diff
@@ -550,8 +566,14 @@ static int vq_ring_enable_interrupt(struct virtqueue *vq, uint16_t ndesc)
...
-	if (virtqueue_nused(vq) > ndesc) {
-		return 1;
+	if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
+		if (virtqueue_nused(vq) > ndesc) {
+			return 1;
+		}
+	} else {
+		if (virtqueue_navail(vq) > ndesc) {
+			return 1;
+		}
```
wjliang: It looks like only virtqueue_enable_cb() calls this enable interrupt API,  where  does enable_cb()  get called in this change?

xiaoxiang781216: Nobody call virtqueue_enable_cb now, since the interrupt is enabled by default. We only need to call virtqueue_disable_cb for the receive queue.
The change just make virtqueue_enable_cb work correctly with both master/slave role, so the result don't surprise us if we call it in the furture.

arnop2:
Hello,
Sorry, too busy this week to test your patches, I will try to test them next week.
One remark, could you add VIRTIO_SLAVE_ONLY and VIRTIO_MASTER_ONLY for code optimization?

xiaoxiang781216:
> Hello,
> Sorry, too busy this week to test your patches, I will try to test them next week.
> One remark, could you add VIRTIO_SLAVE_ONLY and VIRTIO_MASTER_ONLY for code optimization?

Ok, I will add the #ifdef/#else/#endif as need

xiaoxiang781216:
@arnop2 the new patch address your VIRTIO_SLAVE_ONLY/VIRTIO_MASTER_ONLY, please review.

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -114,22 +129,20 @@ static void rproc_virtio_set_features(struct virtio_device *vdev,
...
-#endif
 
 static uint32_t rproc_virtio_negotiate_features(struct virtio_device *vdev,
 						uint32_t features)
 {
-	(void)vdev;
-	(void)features;
-
+	/* Assume the master support all slave features */
+	rproc_virtio_set_features(vdev, features);
```
wjliang: How about do gfeatures = (features & with vdev_rsc->dfeatures) in this function for master?
I think the comments can be removed, from this function, instead, the comments can be put to the rproc_create_vdev().

We have a gap in feature negotiation. I am thinking:
* Opt 1:
to have feature negotiate called for slave too.
The feature negotiate can only be called for slave when the DRIVER_OK is set.
The final features will be dfeatures & gfeatures
The negotiate_features function will set the local features variable. The future get_features will not need to check the virtio config space, but just return the local features variable value.

* opt 2:
get_features() functions always get dfeatures and gfeatures from virtio config space and return (dfeatures & gfeatures).

Maybe Opt2 is better

lib/rpmsg/rpmsg_virtio.c
wjliang: > No, the slave side shouldn't set any field in config space, instead it should get the negotiated result from gfeature directly.

What i mean by "set" is to set the local features field which is the negotiation result. In another commit of the same pull request, i have comments in the negotiate_features.

Just for get_features for master, it should not just get the dfeatures, it should be the negotiated result. It looks like based on your modification, gfeatures sounds like the result, but i am not sure if this is always the case in future, maybe better to do dfeatures & gfeatures in the get_features() function.

xiaoxiang781216: Ok, I will change to dfeatures & gfeatures.

lib/virtio/virtqueue.c
xiaoxiang781216: > It looks like only virtqueue_enable_cb() calls this enable interrupt API, where does enable_cb() get called in this change? Should it get called when it wants to know when there is available buffer (that is the buffer is full)?
Master/slave need always enable rx queue callback to response the command quickly, and never need enable tx queue callback, since:
1.each tx queue callback just notify one buffer is returned to the pool
2.if the pool doesn't empty before, we don't need the notification
3.if the pool is empty, rpmsg_virtio_send_offchannel_raw check the buffer available by busy loop
unless we utilize the callback wakeup the send like Linux, we don't need tx queue callback.

lib/remoteproc/remoteproc_virtio.c
xiaoxiang781216: > How about do gfeatures = (features & with vdev_rsc->dfeatures) in this function for master?

Good point.

> I think the comments can be removed, from this function, instead, the comments can be put to the rproc_create_vdev().
> 

Yes, I should put there.

> We have a gap in feature negotiation. I am thinking:
> 
> * Opt 1:
>   to have feature negotiate called for slave too.
>   The feature negotiate can only be called for slave when the DRIVER_OK is set.
>   The final features will be dfeatures & gfeatures
>   The negotiate_features function will set the local features variable. The future get_features will not need to check the virtio config space, but just return the local features variable value.
> * opt 2:
>   get_features() functions always get dfeatures and gfeatures from virtio config space and return (dfeatures & gfeatures).
> 
> Maybe Opt2 is better

I think too.

wjliang:
merged to master

arnop2:
Sorry for the late answer. 
FYI : non regression test done with success on ST plaform.

PR #148, opened 2019-01-08
Fix for rpmsg_virtio

xiaoxiang781216:
@galak, @wjliang and @arnop2, this patch series fix the bug found in rpmsg_virito.c

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -18,11 +18,11 @@
 
 #define RPMSG_NUM_VRINGS (2)
 
-/* Total tick count for 15secs - 1msec tick. */
-#define RPMSG_TICK_COUNT                        15000
+/* Total tick count for 15secs - 1usec tick. */
+#define RPMSG_TICK_COUNT                        15000000
 
-/* Time to wait - In multiple of 10 msecs. */
-#define RPMSG_TICKS_PER_INTERVAL                10
+/* Time to wait - In multiple of 1 msecs. */
+#define RPMSG_TICKS_PER_INTERVAL                1000
 
 #ifndef VIRTIO_SLAVE_ONLY
 metal_weak void *
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -8,7 +8,6 @@
  */
 
 #include <metal/alloc.h>
-#include <metal/cache.h>
 #include <metal/sleep.h>
 #include <metal/utilities.h>
 #include <openamp/rpmsg_virtio.h>
@@ -191,15 +190,6 @@ static void *rpmsg_virtio_get_rx_buffer(struct rpmsg_virtio_device *rvdev,
 		    virtqueue_get_available_buffer(rvdev->rvq, idx, len);
 	}
 #endif /*!VIRTIO_MASTER_ONLY*/
-
-	if (data) {
-		/* FIX ME: library should not worry about if it needs
-		 * to flush/invalidate cache, it is shared memory.
-		 * The shared memory should be mapped properly before
-		 * using it.
-		 */
-		metal_cache_invalidate(data, (unsigned int)(*len));
-	}
 
 	return data;
 }
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -24,10 +24,6 @@
 /* Time to wait - In multiple of 10 msecs. */
 #define RPMSG_TICKS_PER_INTERVAL                10
 
-#define WORD_SIZE	sizeof(unsigned long)
-#define WORD_ALIGN(a)	((((a) & (WORD_SIZE - 1)) != 0) ? \
-			(((a) & (~(WORD_SIZE - 1))) + WORD_SIZE) : (a))
-
 #ifndef VIRTIO_SLAVE_ONLY
 metal_weak void *
 rpmsg_virtio_shm_pool_get_buffer(struct rpmsg_virtio_shm_pool *shpool,
@@ -50,8 +46,8 @@ void rpmsg_virtio_init_shm_pool(struct rpmsg_virtio_shm_pool *shpool,
 	if (!shpool)
 		return;
 	shpool->base = shb;
-	shpool->size = WORD_ALIGN(size);
-	shpool->avail = WORD_ALIGN(size);
+	shpool->size = size;
+	shpool->avail = size;
 }
 
 /**
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -408,22 +408,21 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
 		metal_mutex_release(&rdev->lock);
 
-		if (!ept)
-			/* Fatal error no endpoint for the given dst addr. */
-			return;
+		if (ept) {
+			if (ept->dest_addr == RPMSG_ADDR_ANY) {
+				/*
+				 * First message received from the remote side,
+				 * update channel destination address
+				 */
+				ept->dest_addr = rp_hdr->src;
+			}
+			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
+					 rp_hdr->len, rp_hdr->src, ept->priv);
 
-		if (ept->dest_addr == RPMSG_ADDR_ANY) {
-			/*
-			 * First message received from the remote side,
-			 * update channel destination address
-			 */
-			ept->dest_addr = rp_hdr->src;
+			RPMSG_ASSERT(status == RPMSG_SUCCESS,
+				     "unexpected callback status\n");
 		}
-		status = ept->cb(ept, (void *)RPMSG_LOCATE_DATA(rp_hdr),
-				 rp_hdr->len, rp_hdr->src, ept->priv);
 
-		RPMSG_ASSERT(status == RPMSG_SUCCESS,
-			     "unexpected callback status\n");
 		metal_mutex_acquire(&rdev->lock);
 
 		/* Return used buffers. */
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -115,7 +115,7 @@ static int rpmsg_virtio_enqueue_buffer(struct rpmsg_virtio_device *rvdev,
 		/* Initialize buffer node */
 		vqbuf.buf = buffer;
 		vqbuf.len = len;
-		return virtqueue_add_buffer(rvdev->svq, &vqbuf, 0, 1, buffer);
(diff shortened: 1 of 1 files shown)


wjliang:
I have just reviewed 
rpmsg_virtio_enqueue_buffer should make the buffer readable, and accidentally marked "approved". I haven't finished reviewing. Please ignore my "Changes approved".

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -578,13 +553,6 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 #endif /*!VIRTIO_MASTER_ONLY*/
 	rvdev->shbuf_io = shm_io;
 
-#ifndef VIRTIO_MASTER_ONLY
-	if (role == RPMSG_REMOTE) {
-		/* wait synchro with the master */
-		rpmsg_virtio_wait_remote_ready(rvdev);
-	}
-#endif /*!VIRTIO_MASTER_ONLY*/
-
```
arnop2: this is right if remoteproc_create_virtio is used but it is not the case in zephyr sample that uses only rpmsg without the remoteproc_virtio:
https://github.com/zephyrproject-rtos/zephyr/pull/11355/commits/fa8147a265b0317fa070f202bce0b4cc107465d2 
@galak: Could you confirm?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -410,22 +410,21 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
...
 
-		if (ept->dest_addr == RPMSG_ADDR_ANY) {
-			/*
-			 * First message received from the remote side,
-			 * update channel destination address
-			 */
-			ept->dest_addr = rp_hdr->src;
+			RPMSG_ASSERT(status == RPMSG_SUCCESS,
+				     "unexpected callback status\n");
 		}
```
wjliang: better to have RPMSG_ASSERT for  for if (!ept) case.

lib/rpmsg/rpmsg_virtio.c
wjliang: Agree with @arnop2, remoteproc_create_virtio() is only used when remoteproc is used. However, RPMsg and virtio is allowed to be used without remoteproc.
We had discussion before on waiting remote to be ready https://github.com/OpenAMP/open-amp/pull/141
We'd better keep checking if virtio ready inside the library, otherwise, additional step will be introduced for user to make sure the virtio is ready before calling the API.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -432,6 +432,10 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 
 		rp_hdr = (struct rpmsg_hdr *)
 			 rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
+		if (rp_hdr == NULL) {
+			/* tell peer we return some rx buffer */
+			virtqueue_kick(rvdev->rvq);
```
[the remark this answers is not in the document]
xiaoxiang781216: Two case here:
1.Linux implementation enable rx callback only when it can't get the buffer from the virtio queue and will disable rx callback in all other condition.
Please reference rpmsg_upref_sleepers/rpmsg_downref_sleepers here:
https://github.com/torvalds/linux/blob/master/drivers/rpmsg/virtio_rpmsg_bus.c
2.OpenAMP use the busy loop to wait the buffer available, so we should always disable rx callback uncoditionally, please see this pull request:
https://github.com/OpenAMP/open-amp/pull/147/commits/370b5692b45607fb2826f4895243f25701528d5b
So the kick only happen when Linux is waiting for the buffer.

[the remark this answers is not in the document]
xiaoxiang781216: @arnop2 this patch isn't optimizition, it's a bug fix to avoid kernel side stuck to wait for rx callback.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: Ok, I will remove this patch in the next update

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: The message come from outside world, should we more robust here to avoid the DoS attack?
Actually, Linux kernel do the similar thing in rpmsg_recv_single:
https://github.com/torvalds/linux/blob/master/drivers/rpmsg/virtio_rpmsg_bus.c

xiaoxiang781216:
@galak, change id remove from all pull request, please review again.

lib/rpmsg/rpmsg_virtio.c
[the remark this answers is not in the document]
arnop2: @xiaoxiang781216 You right, this fix is needed, to avoid that Linux is stuck, waiting free buffers.
so i'm ok with the fix, even if it generates extra interrupt in Linux when there is no more message to treat. So potentially for each message sent by Linux. 
I need to go deep in vring flag management (not fluent with this part...) to review 370b569. Anyway is it a way to not introduce the master/slave role notion in it? i will expect more a sender/receiver view for the vrings.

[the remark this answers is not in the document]
xiaoxiang781216: If some free buffer is available in queue, kernel(rpmsg_downref_sleepers) will disable the callback immediately by set VRING_USED_F_NO_NOTIFY flag, so no extra interrupt happen here. If we trigger the interrupt because VRING_USED_F_NO_NOTIFY isn't set, which mean kernel really need this interrupt to wake up.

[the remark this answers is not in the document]
xiaoxiang781216: To work correctly, vring need to know virtio device contain it is slave or master, since vring need check/modify the different field for interrupt management. The detailed difference could reference this pull request:
https://github.com/OpenAMP/open-amp/pull/147

lib/rpmsg/rpmsg_virtio.c
wjliang: agreed.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -140,15 +140,15 @@ static int rpmsg_virtio_enqueue_buffer(struct rpmsg_virtio_device *rvdev,
...
+					unsigned int *len,
 					unsigned short *idx)
 {
 	unsigned int role = rpmsg_virtio_get_role(rvdev);
 	void *data = NULL;
 
 #ifndef VIRTIO_SLAVE_ONLY
 	if (role == RPMSG_MASTER) {
-		data = virtqueue_get_buffer(rvdev->svq, (uint32_t *)len, idx);
+		data = virtqueue_get_buffer(rvdev->svq, len, idx);
```
wjliang: I have got compilation error with this patch on cortexr5 with gnu gcc toolchain:

ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c: In function 'rpmsg_virtio_get_tx_buffer':
/ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:146:43: error: passing argument 2 of 'virtqueue_get_buffer' from incompatible pointer type [-Werror=incompatible-pointer-types]
   data = virtqueue_get_buffer(rvdev->svq, len, idx);
                                           ^~~
In file included from /ws/jliang/source/open-amp/lib/include/openamp/virtio.h:10:0,
                 from /ws/jliang/source/open-amp/lib/include/openamp/rpmsg_virtio.h:18,
                 from /ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:13:
/ws/jliang/source/open-amp/lib/include/openamp/virtqueue.h:191:7: note: expected 'uint32_t * {aka long unsigned int *}' but argument is of type 'unsigned int *'
 void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx);
       ^~~~~~~~~~~~~~~~~~~~
/ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:157:58: error: passing argument 3 of 'virtqueue_get_available_buffer' from incompatible pointer type [-Werror=incompatible-pointer-types]
   data = virtqueue_get_available_buffer(rvdev->svq, idx, len);
                                                          ^~~
In file included from /ws/jliang/source/open-amp/lib/include/openamp/virtio.h:10:0,
                 from /ws/jliang/source/open-amp/lib/include/openamp/rpmsg_virtio.h:18,
                 from /ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:13:
/ws/jliang/source/open-amp/lib/include/openamp/virtqueue.h:193:7: note: expected 'uint32_t * {aka long unsigned int *}' but argument is of type 'unsigned int *'
 void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c: In function 'rpmsg_virtio_get_rx_buffer':
/ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:185:43: error: passing argument 2 of 'virtqueue_get_buffer' from incompatible pointer type [-Werror=incompatible-pointer-types]
   data = virtqueue_get_buffer(rvdev->rvq, len, idx);
                                           ^~~
In file included from /ws/jliang/source/open-amp/lib/include/openamp/virtio.h:10:0,
                 from /ws/jliang/source/open-amp/lib/include/openamp/rpmsg_virtio.h:18,
                 from /ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:13:
/ws/jliang/source/open-amp/lib/include/openamp/virtqueue.h:191:7: note: expected 'uint32_t * {aka long unsigned int *}' but argument is of type 'unsigned int *'
 void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx);
       ^~~~~~~~~~~~~~~~~~~~
/ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:192:55: error: passing argument 3 of 'virtqueue_get_available_buffer' from incompatible pointer type [-Werror=incompatible-pointer-types]
       virtqueue_get_available_buffer(rvdev->rvq, idx, len);
                                                       ^~~
In file included from /ws/jliang/source/open-amp/lib/include/openamp/virtio.h:10:0,
                 from /ws/jliang/source/open-amp/lib/include/openamp/rpmsg_virtio.h:18,
                 from /ws/jliang/source/open-amp/lib/rpmsg/rpmsg_virtio.c:13:
/ws/jliang/source/open-amp/lib/include/openamp/virtqueue.h:193:7: note: expected 'uint32_t * {aka long unsigned int *}' but argument is of type 'unsigned int *'
 void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

As we use uint32_t in the virtqueue.h, will need to make sure the input arg to this function is uint32_t.

xiaoxiang781216: Ok, I will change the type to uint32_t*.

wjliang:
tested, and merged

PR #149, opened 2019-01-08
Fix in rpmsg

xiaoxiang781216:
@galak, @wjliang and @arnop2, this patch series fix the bug found in rpmsg.c


lib/rpmsg/rpmsg.c
```diff
@@ -185,7 +187,7 @@ static void rpmsg_unregister_endpoint(struct rpmsg_endpoint *ept)
 	rdev = ept->rdev;
 
 	if (ept->addr != RPMSG_ADDR_ANY)
-		rpmsg_release_address(rdev->bitmap, RPMSG_ADDR_BMP_SIZE,
+		rpmsg_release_address(rdev->bitmap, RPMSG_ADDR_BMP_BITS,
```
arnop2: Thanks to point the bug, but your patch seems not fix it. Right parameter here is RPMSG_ADDR_BMP_BITS that corresponds to the size of the bitmap table.
The issue is in rpmsg_is_address_set, rpmsg_set_address and rpmsg_release_address functions that should do the match conversion between the ept addr and the bit offset.

lib/rpmsg/rpmsg.c
```diff
@@ -192,20 +192,18 @@ static void rpmsg_unregister_endpoint(struct rpmsg_endpoint *ept)
...
-	return RPMSG_SUCCESS;
 }
 
 int rpmsg_create_ept(struct rpmsg_endpoint *ept, struct rpmsg_device *rdev,
 		     const char *name, uint32_t src, uint32_t dest,
 		     rpmsg_ept_cb cb, rpmsg_ns_unbind_cb unbind_cb)
 {
-	int status;
+	int status = 0;
 	uint32_t addr = src;
```
arnop2: seems that the initialization is not mandatory but if you keep it please use RPMSG_SUCCESS

xiaoxiang781216: The assignment is required otherwise it will be random value if src equals RPMSG_ADDR_ANY and dst doesn't equals RPMSG_ADDR_ANY.
Ok, I will change 0 to RPMSG_SUCCESS.

lib/rpmsg/rpmsg.c
xiaoxiang781216: Ok, I will move the conversion inside the function

xiaoxiang781216:
> how about make the RPMSG_ADDR_BMP_SIZE as number of bits, and in the rpmsg_device, use metal_bitmap_longs() macro to set the array size?
Ok.

wjliang:
Changes look good to me, I will just need to test.
@arnop2 please test to see if there's anything break your platform.

arnop2:
> Changes look good to me, I will just need to test.
> @arnop2 please test to see if there's anything break your platform.
@wjliang: i will try to find time this week to test it

PR #150, opened 2019-01-08
Fix in remoteproc_virtio

xiaoxiang781216:
@galak, @wjliang and @arnop2, this patch series fix the bug found inremoteproc_virtio.c


lib/remoteproc/remoteproc_virtio.c
```diff
@@ -193,9 +193,6 @@ rproc_virtio_create_vdev(unsigned int role, unsigned int notifyid,
 		vrings_info[i].vq = vq;
 	}
 
-	/* FIXME commended as seems not nedded, already stored in vdev */
-	//rpvdev->notifyid = notifyid;
-	rpvdev->notify = notify;
```
wjliang: where is rpvdev->notify is set if this line is removed?

xiaoxiang781216:
> Where is rpvdev->notify is set? Or should you leave that line in the code?

Fix in the new pull request, thanks for point out this defect.

PR #151, opened 2019-01-08
fix in remoteproc

xiaoxiang781216:


lib/remoteproc/remoteproc.c
```diff
@@ -734,7 +733,7 @@ int remoteproc_load_noblock(struct remoteproc *rproc,
 
 			goto error1;
 		}
-		last_load_state = loader->get_load_state(limg_info);
+		last_load_state = ret;
 		if (*nlen != 0 &&
```
wjliang: the last_load_state is unsigned, do you need to force cast?

xiaoxiang781216: No, last_load_state is signed at the begin of remoteproc_load:
	size_t len, nlen;
	int last_load_state;
       ^^^
	metal_phys_addr_t da, rsc_da;

wjliang:
merged with adding commit description. Please make sure  to have commit description in future  pull request

PR #153, opened 2019-01-08
fix for resource table

xiaoxiang781216:


lib/remoteproc/remoteproc.c
```diff
@@ -846,12 +846,16 @@ unsigned int remoteproc_allocate_id(struct remoteproc *rproc,
 
 	if (start == RSC_NOTIFY_ID_ANY)
 		start = 0;
-	if (end == RSC_NOTIFY_ID_ANY)
+	if (end == 0)
```
wjliang: it is possible  that end == start == 0?

lib/remoteproc/rsc_table_parser.c
```diff
@@ -133,24 +133,22 @@ int handle_vdev_rsc(struct remoteproc *rproc, void *rsc)
 
 	/* only assign notification IDs but do not initialize vdev */
 	notifyid = vdev_rsc->notifyid;
-	if (notifyid == RSC_NOTIFY_ID_ANY) {
-		notifyid = remoteproc_allocate_id(rproc,
-						  notifyid, notifyid + 1);
+	notifyid = remoteproc_allocate_id(rproc,
+					  notifyid, notifyid + 1);
```
wjliang: looks like you have some white space changes, could you use tab for line alignment?

PR #154, opened 2019-01-08
fix in virito

xiaoxiang781216:


xiaoxiang781216:
@galak, the change id is removed from commit message

wjliang:
pushed with updated commit message. Please make sure to have commit description for every commit

PR #155, opened 2019-01-18
 Negotiate individual buffer size dynamically

xiaoxiang781216:
@wjliang and @arnop2,
512 byte buffer size isn't suitable in all case, so this patch let slave specify the best size from resource table.
To keep the compatibility, the new behavior is guarded by VIRTIO_RPMSG_F_BUFSZ.


lib/remoteproc/remoteproc_virtio.c
```diff
@@ -111,20 +111,37 @@ static uint32_t rproc_virtio_negotiate_features(struct virtio_device *vdev,
...
-	(void)src;
-	(void)length;
+	struct remoteproc_virtio *rpvdev;
+	struct fw_rsc_vdev *vdev_rsc;
+	struct metal_io_region *io;
+	char *config;
+
+	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
+	vdev_rsc = rpvdev->vdev_rsc;
+	config = (char *)(&vdev_rsc->vring[vdev->vrings_num]);
```
arnop2: Need to check the config_len first.  Please, check that config_len is not null but also that the requested size + offset is  not higher than the config_len.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -27,7 +27,8 @@ extern "C" {
 #endif
 
 /* The feature bitmap for virtio rpmsg */
-#define VIRTIO_RPMSG_F_NS 0 /* RP supports name service notifications */
+#define VIRTIO_RPMSG_F_NS	0 /* RP supports name service notifications */
+#define VIRTIO_RPMSG_F_BUFSZ	2 /* RP supports get buffer size from config space */
```
arnop2: Very Interesting feature. I would suggest you to propose this in a first step in Linux, to have a common implementation in  Linux and OpenAMP. If someone push another implementation on Linux side, this could not be compatible...

xiaoxiang781216: Yes, I am preparing Linux side patch. once OpenAMP get approved, we will submit the similar change to kernel side too.

lib/remoteproc/remoteproc_virtio.c
xiaoxiang781216: Sure, I will update in the next pull request.

lib/include/openamp/rpmsg_virtio.h
arnop2: let's Wendy decide the best strategy. By experience i would suggest to push first or in parallel in Linux...

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -545,6 +545,14 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 	rdev->ns_bind_cb = ns_bind_cb;
 	vdev->priv = rvdev;
 	rdev->ops.send_offchannel_raw = rpmsg_virtio_send_offchannel_raw;
+
+	if (vdev->features & (1 << VIRTIO_RPMSG_F_BUFSZ)) {
+		rpmsg_virtio_read_config(rvdev, 0,
+					 &rvdev->shbuf_size,
```
wjliang: It looks like you assume offset 0 is the shared buffer size in virito config space.
Better to make it macro in rpmsg_virito.h file.

lib/remoteproc/remoteproc_virtio.c
wjliang: looks like new remoteproc_virtio variable is required to keep the config len, can we always assume that the config space is always right after the predefined vdev space if the shared buffer feature is set?
Also, just thinking if we want to make sure offset is 32bit aligned.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: Sure, I will add it

lib/include/openamp/rpmsg_virtio.h
xiaoxiang781216: Sure, once the review is good here, I will submit the patch to Linux kernel ASAP.

xiaoxiang781216:
@wjliang and @arnop2 the new patch address all your comment, please review.

lib/remoteproc/remoteproc_virtio.c
wjliang: > > looks like new remoteproc_virtio variable is required to keep the config len,
> 
> Yes.
> 
> > can we always assume that the config space is always right after the predefined vdev space if the shared buffer feature is set?
> 
> From remoteproc.h
> /**
> 
>     * struct fw_rsc_vdev - virtio device header
> 
>     * @config_len is the size of the virtio config space of this vdev. The config
> 
>     * space lies in the resource table immediate after this vdev header.
> 
>     * Note: immediately following this structure is the virtio config space for
> 
>     * this vdev (which is specific to the vdev; for more info, read the virtio
> 
>     * spec). the size of the config space is specified by @config_len.
>       */
> 
> 
> As far as I know, there isn't config defined yet, so the buffer size feature should be first one.

As the config space can already be used for customized configuration. And thus, I want to see if everyone is OK to assume that once RPMsg shared buffer size feature bit is set, we assume that the first offset is for the RPMsg shared buffer size.

Or we make some virtio config structure like 
uint32 num_config_objs;
array = {
{
uint32 config_obj_type;
uint32 offset;
}
...
}
followed by config objs

This will make the config space more complicated. But maybe more flexible even for future if we want to add more "common" configurable objects.

> 
> > Also, just thinking if we want to make sure offset is 32bit aligned.
> 
> What do you want to point out?

Sorry that I didn't clarify what I meant. I was thinking more like "common config space alignment".
assuming that the first N bytes are are for "common" predefined configuration (This is for future features extension), which can be followed by user defined configuration. 32bit is not enough.
The feature you are adding is an interesting feature, I don't want to over complicate things. but my concern is for those who are already using the "config" space for their own platform. Any new common "config" we add, they will need to change their resource table. And thus, want to see if we also want to make agreement on the virtio config space with this patch.

xiaoxiang781216: > > > looks like new remoteproc_virtio variable is required to keep the config len,
> > 
> > 
> > Yes.
> > > can we always assume that the config space is always right after the predefined vdev space if the shared buffer feature is set?
> > 
> > 
> > From remoteproc.h
> > /**
> > ```
> > * struct fw_rsc_vdev - virtio device header
> > 
> > * @config_len is the size of the virtio config space of this vdev. The config
> > 
> > * space lies in the resource table immediate after this vdev header.
> > 
> > * Note: immediately following this structure is the virtio config space for
> > 
> > * this vdev (which is specific to the vdev; for more info, read the virtio
> > 
> > * spec). the size of the config space is specified by @config_len.
> >   */
> > ```
> > As far as I know, there isn't config defined yet, so the buffer size feature should be first one.
> 
> As the config space can already be used for customized configuration. And thus, I want to see if everyone is OK to assume that once RPMsg shared buffer size feature bit is set, we assume that the first offset is for the RPMsg shared buffer size.

1.Config space concept doesn't expose to rpmsg driver API level
2.No virtio based backend(e.g. QCOM SMD) doesn't have this concept too
so rpmsg driver(client) is hard(or should avoid) to utilize this feature, and reserve the config spase to the internal implementation?
If somebody use the config space internally, since:
1.They don't share the information publicly
2.The internal field may not general enough to other people
It's better to put the general information at the first, then the optional private field last.

> 
> Or we make some virtio config structure like
> uint32 num_config_objs;
> array = {
> {
> uint32 config_obj_type;
> uint32 offset;
> }
> ...
> }
> followed by config objs
> 
> This will make the config space more complicated. But maybe more flexible even for future if we want to add more "common" configurable objects.
> 

This approach will complicate the implementation and also require the change for current config space user.

> > > Also, just thinking if we want to make sure offset is 32bit aligned.
> > 
> > 
> > What do you want to point out?
> 
> Sorry that I didn't clarify what I meant. I was thinking more like "common config space alignment".
> assuming that the first N bytes are are for "common" predefined configuration (This is for future features extension), which can be followed by user defined configuration. 32bit is not enough.
> The feature you are adding is an interesting feature, I don't want to over complicate things. but my concern is for those who are already using the "config" space for their own platform. Any new common "config" we add, they will need to change their resource table. And thus, want to see if we also want to make agreement on the virtio config space with this patch.

How about this:
struct virtio_rpmsg_config {
  __u32 txbuf_size;
  __u32 rxbuf_size;
  __u32 reserved[14];
}; __attribute__((packed));
so
1.The standard feature could use the reserved field in the furture without breaking compatiblity
2.The internal config could use the space after reserved field
3.Keep the implementation simple

wjliang:
@andersson, could you check this proposal to see if there is any conflict to the Linux kernel implementation?

lib/include/openamp/rpmsg_virtio.h
wjliang: For the master sending message to slave, the buffer size ideally should depends on the actually message size. It looks like this flag is because the master also allocates buffers for slave to send messages, and the message for the slave to send can be larger than the fixed size defined by the master. Have a discussion with @andersson, this change negotiate the size before vrings setup. Will need more time to consider if this is an appropriate change.

xiaoxiang781216: @andersson, the kernel side hard code rpmsg buffer size to MAX_RPMSG_BUF_SIZE(512), it isn't always best choice for the remote side:
1.For small MCU, we want to reduce the buffer size to 256 or even 128
2.But for NBIoT(WiFi), we want to extend to MTU(about 1500)
My proposal is:
1.Add new rx/tx buffer size field into config space
2.Master allocate and split the buffer based on these value
So firmware writer could choice:
1.The best rpmsg buffer number from fw_rsc_vdev_vring->num
2.The best rpmsg buffer size from virtio_rpmsg_config->[rx|tx]buf_size
Of course, to keep the backward compatibility, all new behaviour is guarded by feature bit(VIRTIO_RPMSG_F_BUFSZ).
Do you prefer I submit a kernel patch for discussion?

wjliang: > Do you prefer I submit a kernel patch for discussion?

Yes, please, as this is an interesting feature, if you post it to the Linux kernel, more people will be involved in the discussion. Thanks, Wendy

xiaoxiang781216:
@wjliang, @arnop2, @galak and @@andersson, here is patch for kernel side:
https://patchwork.kernel.org/cover/10790823/

rjmiller62:
How about a generic bit for the RX/TX buf size settings as I believe that virtio-net devices will also need the ability to "set" the sizes vs having random values being picked by each front end?

xiaoxiang781216:

xiaoxiang781216:
Move rpmsg_virtio_read_config after device status become ready.

xiaoxiang781216:
> How about a generic bit for the RX/TX buf size settings as I believe that virtio-net devices will also need the ability to "set" the sizes vs having random values being picked by each front end?

But from my reading virtio spec, the feature bit is specified by individual virtio device, nobody try to define the common bits before.

xiaoxiang781216:
@arnopo @edmooring all my PR update to the latest mainline and pass the check, feel free give me the feedback.

arnopo:
Hi @xiaoxiang781216,
During in the openamp-rp meeting, we started some discussions around the evolution of the resource table evolution including dynamic rpmsg buffer size. One advantage of using the resource table is that this also answers to solution that would not rely on virtio ( possible future extension).
I'm not in favor of implementing a temporary solution that would be revert if we decide to move to  resource table. 
That said, i don't remember if you pushed the patch on Linux side. But feel free to push it on resent it to trig the discussion.

xiaoxiang781216:
Here is the patch set:
https://patchwork.kernel.org/cover/10790823/

xiaoxiang781216:
> Hi @xiaoxiang781216,
> During in the openamp-rp meeting, we started some discussions around the evolution of the resource table evolution including dynamic rpmsg buffer size. One advantage of using the resource table is that this also answers to solution that would not rely on virtio ( possible future extension).
> I'm not in favor of implementing a temporary solution that would be revert if we decide to move to resource table.

This method is also based on resource table. The key difference is where we put the information:
1.Add bufsize field or use reserved field in fw_rsc_vdev_vring:
```
struct fw_rsc_vdev_vring {
	uint32_t da;
	uint32_t align;
	uint32_t num;
	uint32_t notifyid;
	uint32_t reserved;
};
```
The problem is how we pass this information from remoteproc layer to rpmsg layer with a standard and compatible way.
2.Add buf_size to virtio config space(rpmsg doesn't use this before):
```
struct fw_rsc_config {
	/* The tx/rx individual buffer size(if VIRTIO_RPMSG_F_BUFSZ) */
	uint32_t txbuf_size;
	uint32_t rxbuf_size;
	uint32_t reserved[14]; /* Reserve for the future use */
	/* Put the customize config here */
};
```
Since virtio has many functions to access the config space, the 2nd method is more simple and compatible.
If you study virtio spec carefully, you should find that virtio already define how to expose the addtional information in the standard way:
https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html
I suggest to read this spec first and learn how other virtio device resolve the similar issue before you modify the resource table. Many thing can be done with config space without breaking change. 

> That said, i don't remember if you pushed the patch on Linux side. But feel free to push it on resent it to trig the discussion.

arnopo:
@xiaoxiang781216 
You are right, your implementation is also based on resource table, i did not take enough time to resfresh my mind on your patches... 
Anyway, your proposal have an impact on the RPMsg protocol. So we need here a  common solution with Linux. The best process is that you resent you Linux patch ( or send  a V2) .  Without agreement on Linux side this feature will beOpenAMp centric because not "standard" for the rpmsg protocol point of view.

hubertmis:
What's the status of this PR? Any chances to merge it soon?

If not, would it be OK to split it in two:
1. Configurable buffer sizes in rpmsg_virtio layer
2. Setting buffer sizes configuration by remoteproc module

I understand the 2nd one must be compatible with Linux and requires more discussion. I hope the 1st one can be merged earlier.

arnopo:
> 

it seems that we come to the same conclusion (see https://github.com/OpenAMP/open-amp/issues/322#issuecomment-986573828)

PR #159, opened 2019-02-01
Fix sytle

xiaoxiang781216:
@wjliang this series just fix the sytle issue, no behaviour thange, please review.

### DIFF
--- a/apps/examples/echo/rpmsg-ping.c
+++ b/apps/examples/echo/rpmsg-ping.c
@@ -48,14 +48,14 @@ static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
 		r_payload->num, (unsigned long)len);
 
 	if (r_payload->size == 0) {
-		LPERROR(" Invalid size of package is received.\n");
+		LPERROR(" Invalid size of package is received.\r\n");
 		err_cnt++;
 		return RPMSG_SUCCESS;
 	}
 	/* Validate data buffer integrity. */
 	for (i = 0; i < (int)r_payload->size; i++) {
 		if (r_payload->data[i] != 0xA5) {
-			LPRINTF("Data corruption at index %d\n", i);
+			LPRINTF("Data corruption at index %d\r\n", i);
 			err_cnt++;
 			break;
 		}
@@ -68,16 +68,16 @@ static void rpmsg_service_unbind(struct rpmsg_endpoint *ept)
 {
 	(void)ept;
 	rpmsg_destroy_ept(&lept);
-	LPRINTF("echo test: service is destroyed\n");
+	LPRINTF("echo test: service is destroyed\r\n");
 	ept_deleted = 1;
 }
 
 static void rpmsg_name_service_bind_cb(struct rpmsg_device *rdev,
 				       const char *name, uint32_t dest)
 {
-	LPRINTF("new endpoint notification is received.\n");
+	LPRINTF("new endpoint notification is received.\r\n");
 	if (strcmp(name, RPMSG_SERVICE_NAME))
-		LPERROR("Unexpected name service %s.\n", name);
+		LPERROR("Unexpected name service %s.\r\n", name);
 	else
 		(void)rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
 				       APP_EPT_ADDR, dest,
@@ -97,11 +97,11 @@ int app (struct rpmsg_device *rdev, void *priv)
 	int expect_rnum = 0;
 
 	LPRINTF(" 1 - Send data to remote core, retrieve the echo");
-	LPRINTF(" and validate its integrity ..\n");
+	LPRINTF(" and validate its integrity ..\r\n");
 
 	max_size = rpmsg_virtio_get_buffer_size(rdev);
 	if (max_size < 0) {
-		LPERROR("No avaiable buffer size.\n");
+		LPERROR("No avaiable buffer size.\r\n");
 		return -1;
 	}
 	max_size -= sizeof(struct _payload);
@@ -111,7 +111,7 @@ int app (struct rpmsg_device *rdev, void *priv)
 				      max_size);
 
 	if (!i_payload) {
-		LPERROR("memory allocation failed.\n");
+		LPERROR("memory allocation failed.\r\n");
 		return -1;
 	}
 
@@ -121,14 +121,14 @@ int app (struct rpmsg_device *rdev, void *priv)
 			       rpmsg_endpoint_cb, rpmsg_service_unbind);
 
 	if (ret) {
-		LPERROR("Failed to create RPMsg endpoint.\n");
+		LPERROR("Failed to create RPMsg endpoint.\r\n");
 		return ret;
 	}
 
 	while (!is_rpmsg_ept_ready(&lept))
 		platform_poll(priv);
 
-	LPRINTF("RPMSG endpoint is binded with remote.\n");
+	LPRINTF("RPMSG endpoint is binded with remote.\r\n");
 	for (i = 0, size = PAYLOAD_MIN_SIZE; i < num_payloads; i++, size++) {
 		i_payload->num = i;
 		i_payload->size = size;
@@ -136,7 +136,7 @@ int app (struct rpmsg_device *rdev, void *priv)
 		/* Mark the data buffer. */
 		memset(&(i_payload->data[0]), 0xA5, size);
 
-		LPRINTF("sending payload number %lu of size %lu\n",
+		LPRINTF("sending payload number %lu of size %lu\r\n",
 			i_payload->num,
 			(unsigned long)(2 * sizeof(unsigned long)) + size);
 
@@ -144,10 +144,10 @@ int app (struct rpmsg_device *rdev, void *priv)
 				 (2 * sizeof(unsigned long)) + size);
 
 		if (ret < 0) {
-			LPERROR("Failed to send data...\n");
+			LPERROR("Failed to send data...\r\n");
 			break;
 		}
-		LPRINTF("echo test: sent : %lu\n",
+		LPRINTF("echo test: sent : %lu\r\n",
 			(unsigned long)(2 * sizeof(unsigned long)) + size);
 
 		expect_rnum++;
@@ -157,12 +157,12 @@ int app (struct rpmsg_device *rdev, void *priv)
 
 	}
 
-	LPRINTF("**********************************\n");
-	LPRINTF(" Test Results: Error count = %d \n", err_cnt);
-	LPRINTF("**********************************\n");
+	LPRINTF("**********************************\r\n");
+	LPRINTF(" Test Results: Error count = %d \r\n", err_cnt);
+	LPRINTF("**********************************\r\n");
 	/* Destroy the RPMsg endpoint */
 	rpmsg_destroy_ept(&lept);
-	LPRINTF("Quitting application .. Echo test end\n");
+	LPRINTF("Quitting application .. Echo test end\r\n");
 
 	metal_free_memory(i_payload);
 	return 0;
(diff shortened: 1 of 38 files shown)


xiaoxiang781216:
@wjliang How about this pull requst?

lib/remoteproc/elf_loader.c
```diff
@@ -445,51 +445,49 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
 		const void *img_shdrs;
 
 		metal_log(METAL_LOG_DEBUG, "Loading ELF section header.\r\n");
 		shdrs_offset = elf_shoff(*img_info);
 		if (elf_shnum(*img_info) == 0) {
 			*load_state = (*load_state & (~ELF_STATE_MASK)) |
 				       ELF_STATE_HDRS_COMPLETE;
-		       *nlen = 0;
-			return (int)*load_state;
+			*nlen = 0;
```
wjliang: while space

apps/examples/echo/rpmsg-ping.c
```diff
@@ -44,7 +44,7 @@ static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
 	(void)ept;
 	(void)src;
 	(void)priv;
-	LPRINTF(" received payload number %lu of size %lu \r\n",
+	LPRINTF(" received payload number %lu of size %lu \n",
```
wjliang: In some system, if you don't use "\r", the next new line doesn't start from the beginning. make the output hard to read.
Prefer not to remove "\r".

lib/remoteproc/elf_loader.c
```diff
@@ -184,33 +184,32 @@ static void elf_parse_segment(void *elf_info, const void *elf_phdr,
 		const Elf32_Phdr *phdr = elf_phdr;
 
 		if (p_type != NULL)
-			*p_type = (unsigned int)phdr->p_type;
+			*p_type = phdr->p_type;
 		if (p_offset != NULL)
-			*p_offset = (size_t)phdr->p_offset;
+			*p_offset = phdr->p_offset;
```
wjliang: The casting is required as the program header is using 32/bit type, but the size_t, metal_phys_addr_t is system dependent. it maynot be the same as the one defined in the program header structure.

lib/remoteproc/elf_loader.c
```diff
@@ -650,17 +648,17 @@ metal_phys_addr_t elf_get_entry(void *elf_info)
 		return METAL_BAD_PHYS;
 
 	if (elf_is_64(elf_info) == 0) {
-		Elf32_Ehdr *elf_ehdr = (Elf32_Ehdr *)elf_info;
+		Elf32_Ehdr *elf_ehdr = elf_info;
 		Elf32_Addr e_entry;
 
 		e_entry = elf_ehdr->e_entry;
-		return (metal_phys_addr_t)e_entry;
+		return e_entry;
```
wjliang: should still use the metal_phys_addr_t to force casting, to avoid compilation warning in some system as the size of the entry may not be the same as metal_phys_addr_t

lib/remoteproc/elf_loader.c
```diff
@@ -650,17 +648,17 @@ metal_phys_addr_t elf_get_entry(void *elf_info)
...
 	} else {
-		Elf64_Ehdr *elf_ehdr = (Elf64_Ehdr *)elf_info;
+		Elf64_Ehdr *elf_ehdr = elf_info;
 		Elf64_Addr e_entry;
 
 		e_entry = elf_ehdr->e_entry;
-		return (metal_phys_addr_t)(e_entry & (metal_phys_addr_t)(-1));
+		return e_entry;
```
wjliang: should keep the casting, but may not need e_entry & (metal_phys_addr_t)(-1)

lib/remoteproc/remoteproc.c
```diff
@@ -521,7 +521,7 @@ int remoteproc_load(struct remoteproc *rproc, const char *path,
 			img_data = NULL;
 			/* get the I/O region from remoteproc */
 			pa = METAL_BAD_PHYS;
-			(void)remoteproc_mmap(rproc, &pa, &da, nmemsize, 0, &io);
+			remoteproc_mmap(rproc, &pa, &da, nmemsize, 0, &io);
```
wjliang: this is to avoid compilation warning in some system.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -251,7 +251,7 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
 		 * so get the buffer size from the virtqueue.
 		 */
 		length =
-		    (int)virtqueue_get_desc_size(rvdev->svq) -
+		    virtqueue_get_desc_size(rvdev->svq) -
```
wjliang: this returns unsigned, it can be compilation error if assign unsigned to int.

lib/virtio/virtqueue.c
```diff
@@ -180,7 +180,7 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 
 	atomic_thread_fence(memory_order_seq_cst);
 
-	desc_idx = (uint16_t)uep->id;
+	desc_idx = uep->id;
```
wjliang: uint16_t casting is required as the uep->id is uint32_t, but the desc_idx is uint16_t.

lib/remoteproc/elf_loader.c
xiaoxiang781216: Fix

apps/examples/echo/rpmsg-ping.c
xiaoxiang781216: But, many place use only "\n", we need make the consistency here, either "\r\n" or "\n". From the C standard, the console driver should auto add "\r" before "\n" if the platform need, so it's better to use "\n".

lib/remoteproc/elf_loader.c
xiaoxiang781216: Fix

lib/remoteproc/elf_loader.c
xiaoxiang781216: Fix

lib/remoteproc/elf_loader.c
xiaoxiang781216: Fix

lib/remoteproc/remoteproc.c
xiaoxiang781216: Fix

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: Fix

lib/virtio/virtqueue.c
xiaoxiang781216: Fix

xiaoxiang781216:
@wjliang do you have any other comments?

lib/remoteproc/elf_loader.c
```diff
@@ -429,13 +429,13 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
 	load_state = elf_load_state(*img_info);
-	if (last_load_state != (int)*load_state)
+	if (last_load_state != *load_state)
 		return -RPROC_EINVAL;
 	/* Get ELF program headers */
 	if (*load_state == ELF_STATE_WAIT_FOR_PHDRS) {
 		size_t phdrs_size;
 		size_t phdrs_offset;
-		char **phdrs;
+		void **phdrs;
```
wjliang: void *phdrs;

lib/remoteproc/elf_loader.c
```diff
@@ -445,51 +445,49 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
+		memcpy(*phdrs, img_phdrs, phdrs_size);
 		*load_state = ELF_STATE_WAIT_FOR_SHDRS |
 			       RPROC_LOADER_READY_TO_LOAD;
 	}
 	/* Get ELF Section Headers */
 	if ((*load_state & ELF_STATE_WAIT_FOR_SHDRS) != 0) {
 		size_t shdrs_size;
 		size_t shdrs_offset;
-		char **shdrs;
+		void **shdrs;
```
wjliang: void *shdrs;

lib/include/openamp/rpmsg.h
```diff
@@ -344,8 +343,8 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept);
  */
 static inline unsigned int is_rpmsg_ept_ready(struct rpmsg_endpoint *ept)
 {
-	return (ept->dest_addr != RPMSG_ADDR_ANY &&
-		ept->addr != RPMSG_ADDR_ANY);
+	return ept->dest_addr != RPMSG_ADDR_ANY &&
+		ept->addr != RPMSG_ADDR_ANY;
```
wjliang: better to be (ept->dest_addr != RPMSG_ADDR_ANY) &&
		(ept->addr != RPMSG_ADDR_ANY) for easier to read

lib/include/openamp/virtqueue.h
```diff
@@ -88,7 +84,7 @@ struct virtqueue {
 	uint16_t vq_available_idx;
 
 #ifdef VQUEUE_DEBUG
-	boolean vq_inuse;
+	bool vq_inuse;
```
wjliang: the OpenAMP defines boolean because some compiler doesn't have bool defined. Better to define bool in generic compiler header if it is not defined.

wjliang:
> @wjliang do you have any other comments?

Thanks for your update. It is a very good pull request to fix the syntax errors. I just added a few comments. Just in future, it is better to have one commit for one thing,  such that it is possible to get some commits in first.

xiaoxiang781216:
> > @wjliang do you have any other comments?
> 
> Thanks for your update. It is a very good pull request to fix the syntax errors. I just added a few comments. Just in future, it is better to have one commit for one thing, such that it is possible to get some commits in first.

Sure, will apply in the future pull request.

lib/include/openamp/rpmsg.h
xiaoxiang781216: Ok

lib/include/openamp/virtqueue.h
xiaoxiang781216: but stdbool.h is already included by rpmsg.h and virtqueue.h, according this link:
http://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdbool.h.html
the purpose of stdbool.h is just for bool/false/true definition, so we could safely assume bool is defined if stdbool.h exist.

lib/remoteproc/elf_loader.c
xiaoxiang781216: > void *shdrs;

Can't change to void *, since we need save the allocated memory:
*shdrs = metal_allocate_memory(shdrs_size);

lib/remoteproc/elf_loader.c
xiaoxiang781216: Can't change to void *, since we need save the allocated memory:
*phdrs= metal_allocate_memory(phdrs_size);

apps/examples/echo/rpmsg-ping.c
wjliang: There are console drivers don't auto add "\r", prefer to keep "\r".

lib/include/openamp/virtqueue.h
wjliang: Good catch. just bool doesn't exist in C89/90. and thus, for those C compilers, there will be issue.
but, as you point out, there is inconsistency in the current code base. I am fine to just use bool and remove the boolean. Please also remove "typedef uint8_t boolean;" in lib/include/openamp/virtqueue.h

xiaoxiang781216: Already remove in the same patch

apps/examples/echo/rpmsg-ping.c
xiaoxiang781216: Ok, I will change all single \n to \r\n to keep the consistency.

xiaoxiang781216:
The new patch change all \n to \r\n, please review @wjliang.

wjliang:
> The new patch change all \n to \r\n, please review @wjliang.

I have reviewed your patches, just i am trying to test on our platform.

wjliang:
merged

PR #161, opened 2019-02-06
Shouldn't send ns message if no VIRTIO_RPMSG_F_NS or name

xiaoxiang781216:
@wjliang and @arnop2 let focus on the bug fix first, then the enhancement.
I will rebase https://github.com/OpenAMP/open-amp/pull/160 on this pull after this pull get approved.

### DIFF
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -98,6 +98,7 @@ struct rpmsg_device_ops {
  * @ns_bind_cb: callback handler for name service announcement without local
  *              endpoints waiting to bind.
  * @ops: RPMsg device operations
+ * @support_ns: create/destroy namespace message
  */
 struct rpmsg_device {
 	struct metal_list endpoints;
@@ -106,6 +107,7 @@ struct rpmsg_device {
 	metal_mutex_t lock;
 	rpmsg_ns_bind_cb ns_bind_cb;
 	struct rpmsg_device_ops ops;
+	bool support_ns;
 };
 
 /**
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -288,7 +288,7 @@ static inline void rpmsg_init_ept(struct rpmsg_endpoint *ept,
 				  rpmsg_ept_cb cb,
 				  rpmsg_ns_unbind_cb ns_unbind_cb)
 {
-	strncpy(ept->name, name, sizeof(ept->name));
+	strncpy(ept->name, name ? name : "", sizeof(ept->name));
 	ept->addr = src;
 	ept->dest_addr = dest;
 	ept->cb = cb;
--- a/lib/rpmsg/rpmsg.c
+++ b/lib/rpmsg/rpmsg.c
@@ -227,7 +227,7 @@ int rpmsg_create_ept(struct rpmsg_endpoint *ept, struct rpmsg_device *rdev,
 	rpmsg_init_ept(ept, name, addr, dest, cb, unbind_cb);
 	rpmsg_register_endpoint(rdev, ept);
 
-	if (ept->dest_addr == RPMSG_ADDR_ANY) {
+	if (rdev->support_ns && ept->dest_addr == RPMSG_ADDR_ANY) {
 		/* Send NS announcement to remote processor */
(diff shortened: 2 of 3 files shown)


lib/include/openamp/rpmsg.h
```diff
@@ -325,6 +325,31 @@ int rpmsg_create_ept(struct rpmsg_endpoint *ept, struct rpmsg_device *rdev,
...
+ * @ept: pointer to rpmsg endpoint
+ * @src: local address of the endpoint
+ * @dest: target address of the endpoint
+ * @cb: endpoint callback
+ *
+ * In essence, an rpmsg endpoint represents a listener on the rpmsg bus, as
+ * it binds an rpmsg address with an rx callback handler.
+ */
+
+static inline int rpmsg_create_anon_ept(struct rpmsg_endpoint *ept,
```
wjliang: Why it needs a function to create endpoint without a name? if the endpoint creation function itself allows empty name.
I am fine that to have endpoint without name assuming it is static endpoint. I think it ok to have static endpoint even when NS feature is enabled, but just I don't need it requires a specific wrapper for it. @arnop2, do you have any concern of it?

lib/rpmsg/rpmsg.c
```diff
@@ -227,7 +227,7 @@ int rpmsg_create_ept(struct rpmsg_endpoint *ept, struct rpmsg_device *rdev,
 	rpmsg_init_ept(ept, name, addr, dest, cb, unbind_cb);
 	rpmsg_register_endpoint(rdev, ept);
 
-	if (rdev->support_ns && ept->dest_addr == RPMSG_ADDR_ANY) {
+	if (ept->name[0] && rdev->support_ns && ept->dest_addr == RPMSG_ADDR_ANY) {
```
wjliang: why there is endpoint without a name but it doesn't have dest_addr? I had thought the endpoint without a name is static endpoint.

lib/include/openamp/rpmsg.h
xiaoxiang781216: Just make the creation of static point a little bit easier and visible, and align with the kernel api rpmsg_create_ept.

lib/rpmsg/rpmsg.c
xiaoxiang781216: Yes, the static endpoint should always have the dest_addr, the modification just make it's symmetry with rpmsg_create_ept

lib/include/openamp/rpmsg.h
wjliang: I still doubt if it is necessary to define such a function and it is only used for a subset of static endpoints.

lib/rpmsg/rpmsg.c
wjliang: name is empty and dest address is ANY endpoint is just invalid, it looks like this should be check when creating the endpoint, and then, we don't need to check if the name is empty or not here.

lib/rpmsg/rpmsg.c
```diff
@@ -257,7 +257,7 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept)
 		return;
 
 	rdev = ept->rdev;
-	if (rdev->support_ns && ept->addr != RPMSG_NS_EPT_ADDR)
+	if (ept->name[0] && rdev->support_ns && ept->addr != RPMSG_NS_EPT_ADDR)
```
wjliang: The same comment as before.

lib/rpmsg/rpmsg.c
xiaoxiang781216: Yes, it's better to avoid the invalid combination in the first place.

lib/include/openamp/rpmsg.h
xiaoxiang781216: Ok, I will remove rpmsg_create_anon_ept, since the nameless endpoint is a seldom used feature.

lib/rpmsg/rpmsg.c
xiaoxiang781216: The check is required since the nameless endpoint have the valid local address but RPMSG_NS_DESTROY shouldn't send out.

xiaoxiang781216:
@wjliang do you have any concern about the new patch?

wjliang:
merged.

PR #162, opened 2019-02-17
rpmsg: bring back zero copy transfer

xiaoxiang781216:
Two simplify the review, I split two small independent patch to new PR:
https://github.com/OpenAMP/open-amp/pull/240
https://github.com/OpenAMP/open-amp/pull/241

### DIFF
--- a/apps/tests/msg/rpmsg-nocopy-ping.c
+++ b/apps/tests/msg/rpmsg-nocopy-ping.c
@@ -1,0 +1,199 @@
+/*
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright (c) 2020, Xiaomi Inc. All rights reserved.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <openamp/open_amp.h>
+#include <metal/alloc.h>
+#include "platform_info.h"
+#include "rpmsg-ping.h"
+
+#define LPRINTF(format, ...) printf(format, ##__VA_ARGS__)
+#define LPERROR(format, ...) LPRINTF("ERROR: " format, ##__VA_ARGS__)
+
+struct _payload {
+	unsigned long num;
+	unsigned long size;
+	unsigned char data[];
+};
+
+static int err_cnt;
+
+#define PAYLOAD_MIN_SIZE	1
+
+/* Globals */
+static struct rpmsg_endpoint lept;
+static int rnum;
+static int err_cnt;
+static int ept_deleted;
+
+/* External functions */
+extern int init_system(void);
+extern void cleanup_system(void);
+
+static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
+			     uint32_t src, void *priv)
+{
+	int i;
+	struct _payload *r_payload = (struct _payload *)data;
+
+	(void)ept;
+	(void)src;
+	(void)priv;
+	LPRINTF(" received payload number %lu of size %lu \r\n",
+		r_payload->num, (unsigned long)len);
+
+	if (r_payload->size == 0) {
+		LPERROR(" Invalid size of package is received.\r\n");
+		err_cnt++;
+		return RPMSG_SUCCESS;
+	}
+	/* Validate data buffer integrity. */
+	for (i = 0; i < (int)r_payload->size; i++) {
+		if (r_payload->data[i] != 0xA5) {
+			LPRINTF("Data corruption at index %d\r\n", i);
+			err_cnt++;
+			break;
+		}
+	}
+	rnum = r_payload->num + 1;
+	return RPMSG_SUCCESS;
+}
+
+static void rpmsg_service_unbind(struct rpmsg_endpoint *ept)
+{
+	(void)ept;
+	rpmsg_destroy_ept(&lept);
+	LPRINTF("echo test: service is destroyed\r\n");
+	ept_deleted = 1;
+}
+
+static void rpmsg_name_service_bind_cb(struct rpmsg_device *rdev,
+				       const char *name, uint32_t dest)
+{
+	LPRINTF("new endpoint notification is received.\r\n");
+	if (strcmp(name, RPMSG_SERVICE_NAME))
+		LPERROR("Unexpected name service %s.\r\n", name);
+	else
+		(void)rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
+				       RPMSG_ADDR_ANY, dest,
+				       rpmsg_endpoint_cb,
+				       rpmsg_service_unbind);
+
+}
+
+static int app(struct rpmsg_device *rdev, void *priv)
+{
+	int ret;
+	int i, num_payloads;
+	uint32_t size, max_size;
+	int expect_rnum = 0;
+
+	LPRINTF(" 1 - Send data to remote core, retrieve the echo");
+	LPRINTF(" and validate its integrity ..\r\n");
+
+	max_size = rpmsg_virtio_get_buffer_size(rdev);
+	if ((int32_t)max_size < 0) {
+		LPERROR("No available buffer size.\r\n");
+		return -1;
+	}
+	max_size -= sizeof(struct _payload);
+	num_payloads = max_size - PAYLOAD_MIN_SIZE + 1;
+
+	/* Create RPMsg endpoint */
+	ret = rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
+			       RPMSG_ADDR_ANY, RPMSG_ADDR_ANY,
+			       rpmsg_endpoint_cb, rpmsg_service_unbind);
+
+	if (ret) {
+		LPERROR("Failed to create RPMsg endpoint.\r\n");
+		return ret;
+	}
+
+	while (!is_rpmsg_ept_ready(&lept))
+		platform_poll(priv);
+
+	LPRINTF("RPMSG endpoint is binded with remote.\r\n");
+	for (i = 0, size = PAYLOAD_MIN_SIZE; i < num_payloads; i++, size++) {
+		struct _payload *i_payload;
+
+		i_payload = rpmsg_get_tx_payload_buffer(&lept, &max_size, 1);
+		if (!i_payload) {
+			LPERROR("Failed to get payload...\r\n");
+			break;
+		}
+
+		i_payload->num = i;
+		i_payload->size = size;
+
+		/* Mark the data buffer. */
+		memset(&(i_payload->data[0]), 0xA5, size);
+
+		LPRINTF("sending payload number %lu of size %lu\r\n",
+			i_payload->num,
+			(unsigned long)(2 * sizeof(unsigned long)) + size);
+
+		ret = rpmsg_send_nocopy(&lept, i_payload,
+					(2 * sizeof(unsigned long)) + size);
+
+		if (ret < 0) {
+			LPERROR("Failed to send data...\r\n");
+			break;
+		}
+		LPRINTF("echo test: sent : %lu\r\n",
+			(unsigned long)(2 * sizeof(unsigned long)) + size);
+
+		expect_rnum++;
+		do {
+			platform_poll(priv);
+		} while ((rnum < expect_rnum) && !err_cnt && !ept_deleted);
+
+	}
+
+	LPRINTF("**********************************\r\n");
+	LPRINTF(" Test Results: Error count = %d \r\n", err_cnt);
+	LPRINTF("**********************************\r\n");
+	/* Destroy the RPMsg endpoint */
+	rpmsg_destroy_ept(&lept);
+	LPRINTF("Quitting application .. Echo test end\r\n");
+
+	return 0;
+}
+
+int main(int argc, char *argv[])
+{
+	void *platform;
+	struct rpmsg_device *rpdev;
+	int ret;
+
+	/* Initialize platform */
+	ret = platform_init(argc, argv, &platform);
+	if (ret) {
+		LPERROR("Failed to initialize platform.\r\n");
+		ret = -1;
+	} else {
+		rpdev = platform_create_rpmsg_vdev(platform, 0,
+						  VIRTIO_DEV_MASTER,
+						  NULL,
+						  rpmsg_name_service_bind_cb);
+		if (!rpdev) {
+			LPERROR("Failed to create rpmsg virtio device.\r\n");
+			ret = -1;
+		} else {
+			app(rpdev, platform);
+			platform_release_rpmsg_vdev(rpdev, platform);
+			ret = 0;
+		}
+	}
+
+	LPRINTF("Stopping application...\r\n");
+	platform_cleanup(platform);
+
+	return ret;
+}
+
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -83,11 +83,22 @@ struct rpmsg_endpoint {
 /**
  * struct rpmsg_device_ops - RPMsg device operations
  * @send_offchannel_raw: send RPMsg data
+ * @hold_rx_buffer: hold RPMsg RX buffer
+ * @release_rx_buffer: release RPMsg RX buffer
+ * @get_tx_payload_buffer: get RPMsg TX buffer
+ * @send_offchannel_nocopy: send RPMsg data without copy
  */
 struct rpmsg_device_ops {
 	int (*send_offchannel_raw)(struct rpmsg_device *rdev,
 				   uint32_t src, uint32_t dst,
 				   const void *data, int len, int wait);
+	void (*hold_rx_buffer)(struct rpmsg_device *rdev, void *rxbuf);
+	void (*release_rx_buffer)(struct rpmsg_device *rdev, void *rxbuf);
+	void *(*get_tx_payload_buffer)(struct rpmsg_device *rdev,
+				       uint32_t *len, int wait);
+	int (*send_offchannel_nocopy)(struct rpmsg_device *rdev,
+				      uint32_t src, uint32_t dst,
+				       const void *data, int len);
 };
 
 /**
@@ -263,6 +274,166 @@ static inline int rpmsg_trysend_offchannel(struct rpmsg_endpoint *ept,
 					   const void *data, int len)
 {
 	return rpmsg_send_offchannel_raw(ept, src, dst, data, len, false);
+}
+
+/**
+ * @brief Holds the rx buffer for usage outside the receive callback.
+ *
+ * Calling this function prevents the RPMsg receive buffer from being released
+ * back to the pool of shmem buffers. This API can only be called at rx
+ * callback context (rpmsg_rx_cb_t). With this API, the application doesn't
+ * need to copy the message in rx callback. Instead, the rx buffer base address
+ * is saved in application context and further processed in application
+ * process. After the message is processed, the application can release the rx
+ * buffer for future reuse in vring by calling the rpmsg_release_rx_buffer()
+ * function.
+ *
+ * @param: ept The rpmsg endpoint
+ * @param: rxbuf RX buffer with message payload
+ *
+ * @see rpmsg_release_rx_buffer
+ */
+void rpmsg_hold_rx_buffer(struct rpmsg_endpoint *ept, void *rxbuf);
+
+/**
+ * @brief Releases the rx buffer for future reuse in vring.
+ *
+ * This API can be called at process context when the message in rx buffer is
+ * processed.
+ *
+ * @ept: the rpmsg endpoint
+ * @rxbuf: rx buffer with message payload
+ *
+ * @see rpmsg_hold_rx_buffer
+ */
+void rpmsg_release_rx_buffer(struct rpmsg_endpoint *ept, void *rxbuf);
+
+/**
+ * @brief Gets the tx buffer for message payload.
+ *
+ * This API can only be called at process context to get the tx buffer in vring.
+ * By this way, the application can directly put its message into the vring tx
+ * buffer without copy from an application buffer.
+ * It is the application responsibility to correctly fill the allocated tx
+ * buffer by data and passing correct parameters to the rpmsg_send_nocopy() or
+ * rpmsg_sendto_nocopy() function to perform data no-copy-send mechanism.
+ *
+ * @ept:  Pointer to rpmsg endpoint
+ * @len:  Pointer to store tx buffer size
+ * @wait: Boolean, wait or not for buffer to become available
+ *
+ * @return The tx buffer address on success and NULL on failure
+ *
+ * @see rpmsg_send_offchannel_nocopy
+ * @see rpmsg_sendto_nocopy
+ * @see rpmsg_send_nocopy
+ */
+void *rpmsg_get_tx_payload_buffer(struct rpmsg_endpoint *ept,
+				  uint32_t *len, int wait);
+
+/**
+ * rpmsg_send_offchannel_nocopy() - send a message in tx buffer reserved by
+ * rpmsg_get_tx_payload_buffer() across to the remote processor.
+ *
+ * This function sends buf of length len to the remote dst address,
+ * and uses src as the source address.
+ * The message will be sent to the remote processor which the ept
+ * endpoint belongs to.
+ * The application has to take the responsibility for:
+ *  1. tx buffer reserved (rpmsg_get_tx_payload_buffer() )
+ *  2. filling the data to be sent into the pre-allocated tx buffer
+ *  3. not exceeding the buffer size when filling the data
+ *  4. data cache coherency
+ *
+ * After the rpmsg_send_offchannel_nocopy() function is issued the tx buffer is
+ * no more owned by the sending task and must not be touched anymore unless the
+ * rpmsg_send_offchannel_nocopy() function fails and returns an error. In that
+ * case application should try to re-issue the rpmsg_send_offchannel_nocopy()
+ * again.
+ *
+ * @ept:  The rpmsg endpoint
+ * @src:  The rpmsg endpoint local address
+ * @dst:  The rpmsg endpoint remote address
+ * @data: TX buffer with message filled
+ * @len:  Length of payload
+ *
+ * @return number of bytes it has sent or negative error value on failure.
+ *
+ * @see rpmsg_get_tx_payload_buffer
+ * @see rpmsg_sendto_nocopy
+ * @see rpmsg_send_nocopy
+ */
+int rpmsg_send_offchannel_nocopy(struct rpmsg_endpoint *ept, uint32_t src,
+				 uint32_t dst, const void *data, int len);
+
+/**
+ * @brief rpmsg_sendto_nocopy() - sends a message in tx buffer allocated by
+ * rpmsg_get_tx_payload_buffer() across to the remote processor, specify dst.
+ *
+ * This function sends buf of length len to the remote dst address.
+ * The message will be sent to the remote processor which the ept
+ * endpoint belongs to, using ept's source address.
+ * The application has to take the responsibility for:
+ *  1. tx buffer allocation (rpmsg_get_tx_payload_buffer() )
+ *  2. filling the data to be sent into the pre-allocated tx buffer
+ *  3. not exceeding the buffer size when filling the data
+ *  4. data cache coherency
+ *
+ * After the rpmsg_sendto_nocopy() function is issued the tx buffer is no more
+ * owned by the sending task and must not be touched anymore unless the
+ * rpmsg_sendto_nocopy() function fails and returns an error. In that case the
+ * application should try to re-issue the rpmsg_sendto_nocopy() again.
+ *
+ * @ept:  The rpmsg endpoint
+ * @data: TX buffer with message filled
+ * @len:  Length of payload
+ * @dst:  Destination address
+ *
+ * @return number of bytes it has sent or negative error value on failure.
+ *
+ * @see rpmsg_get_tx_payload_buffer
+ * @see rpmsg_send_offchannel_nocopy
+ * @see rpmsg_send_nocopy
+ */
+static inline int rpmsg_sendto_nocopy(struct rpmsg_endpoint *ept,
+				      const void *data, int len, uint32_t dst)
+{
+	return rpmsg_send_offchannel_nocopy(ept, ept->addr, dst, data, len);
+}
+
+/**
+ * rpmsg_send_nocopy() - send a message in tx buffer reserved by
+ * rpmsg_get_tx_payload_buffer() across to the remote processor.
+ *
+ * This function sends buf of length len on the ept endpoint.
+ * The message will be sent to the remote processor which the ept
+ * endpoint belongs to, using ept's source and destination addresses.
+ * The application has to take the responsibility for:
+ *  1. tx buffer reserved (rpmsg_get_tx_payload_buffer() )
+ *  2. filling the data to be sent into the pre-allocated tx buffer
+ *  3. not exceeding the buffer size when filling the data
+ *  4. data cache coherency
+ *
+ * After the rpmsg_send_nocopy() function is issued the tx buffer is no more
+ * owned by the sending task and must not be touched anymore unless the
+ * rpmsg_send_nocopy() function fails and returns an error. In that case the
+ * application should try to re-issue the rpmsg_send_nocopy() again.
+ *
+ * @ept:  The rpmsg endpoint
+ * @data: TX buffer with message filled
+ * @len:  Length of payload
+ *
+ * @return number of bytes it has sent or negative error value on failure.
+ *
+ * @see rpmsg_get_tx_payload_buffer
+ * @see rpmsg_send_offchannel_nocopy
+ * @see rpmsg_sendto_nocopy
+ */
+static inline int rpmsg_send_nocopy(struct rpmsg_endpoint *ept,
+				    const void *data, int len)
+{
+	return rpmsg_send_offchannel_nocopy(ept, ept->addr,
+					    ept->dest_addr, data, len);
 }
 
 /**
--- a/lib/rpmsg/rpmsg_internal.h
+++ b/lib/rpmsg/rpmsg_internal.h
@@ -35,7 +35,12 @@ extern "C" {
 	} while (0)
 #endif
 
+#define RPMSG_BUF_HELD (1U << 31) /* Flag to suggest to hold the buffer */
+
+#define RPMSG_LOCATE_HDR(p) \
+	((struct rpmsg_hdr *)((unsigned char *)(p) - sizeof(struct rpmsg_hdr)))
 #define RPMSG_LOCATE_DATA(p) ((unsigned char *)(p) + sizeof(struct rpmsg_hdr))
+
 /**
  * enum rpmsg_ns_flags - dynamic name service announcement flags
  *
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -147,6 +147,7 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
 			data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
 							RPMSG_BUFFER_SIZE);
 			*len = RPMSG_BUFFER_SIZE;
+			*idx = 0;
 		}
 	}
 #endif /*!VIRTIO_SLAVE_ONLY*/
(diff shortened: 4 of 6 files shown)


xiaoxiang781216:
@wjliang, this patch bring back @MarekNovakNXP's zero copy transfer, please review it.

lib/include/openamp/rpmsg.h
```diff
@@ -123,7 +134,7 @@ struct rpmsg_device {
  * Returns number of bytes it has sent or negative error value on failure.
  */
 int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
-			      uint32_t dst, const void *data, int size,
+			      uint32_t dst, const void *data, int len,
```
wjliang: No need to change the input arg name.

lib/include/openamp/rpmsg.h
```diff
@@ -81,12 +81,23 @@ struct rpmsg_endpoint {
...
+	void (*release_rx_buffer)(struct rpmsg_device *rdev, void *rxbuf);
+	void *(*get_tx_payload_buffer)(struct rpmsg_device *rdev,
+				       unsigned int *len, int wait);
+	int (*send_offchannel_nocopy)(struct rpmsg_device *rdev,
+				      uint32_t src, uint32_t dst,
+				       const void *data, int len);
 	int (*send_offchannel_raw)(struct rpmsg_device *rdev,
 				   uint32_t src, uint32_t dst,
-				   const void *data, int size, int wait);
+				   const void *data, int len, int wait);
```
wjliang: No need to change the input arg name

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -109,14 +108,17 @@ static int rpmsg_virtio_enqueue_buffer(struct rpmsg_virtio_device *rvdev,
 
 		/* Initialize buffer node */
 		vqbuf.buf = buffer;
-		vqbuf.len = len;
+		vqbuf.len = RPMSG_BUFFER_SIZE;
```
wjliang: you fixed the rpmsg buffer size here, but it is actually up to the allocator, it is not necessary to have fixed size buffer.

wjliang:
thanks for you patch, could you share your use case to have this feature?
From your implementation, you add restrictions to the internal implementation, e.g. you use fixed buffer size, this will limit the future extension. And thus, want to know the need to bring back this feature.

lib/include/openamp/rpmsg.h
xiaoxiang781216: > No need to change the input arg name.

I make this change for consistency, because all other send functions(e.g. rpmsg_send) use len. Does it make sense?

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: > thanks for you patch, could you share your use case to have this feature?

We have several drivers use this feature:
1.rpmsg uart driver for console/gps/modem
2.rpmsg syslog driver for printf/log redirection
3.rpmsg netdev driver to talk with onchip mac layer(modem/wifi/ethernet)
4.rpmsg hostfs driver for file API redirection
5.rpmsg usrsock driver for sock API redirection
6.rpmsg clk driver for clk API redirection
Since these drivers:
1.Transfer the huge or variable length packet
2.The data transfer is the critical performance path
zero-copy could avoid one extra memory copy and then boost the performance.

> From your implementation, you add restrictions to the internal implementation, e.g. you use fixed buffer size, this will limit the future extension. And thus, want to know the need to bring back this feature.

Ok, I will refine the implementation to save the length in some reserved field as before.

xiaoxiang781216: > you fixed the rpmsg buffer size here, but it is actually up to the allocator, it is not necessary to have fixed size buffer. You will need to also remember the buffer size.

Ok, I will save the length in some reserved field as before.

xiaoxiang781216:
> thanks for you patch, could you share your use case to have this feature?

We have several drivers use this feature:
1.rpmsg uart driver for console/gps/modem
2.rpmsg syslog driver for printf/log redirection
3.rpmsg netdev driver to talk with onchip mac layer(modem/wifi/ethernet)
4.rpmsg hostfs driver for file API redirection
5.rpmsg usrsock driver for sock API redirection
6.rpmsg clk driver for clk API redirection
Since these drivers:
1.Transfer the huge or variable length packet
2.The data transfer is the critical performance path
zero-copy could avoid one extra memory copy and then boost the performance.

> From your implementation, you add restrictions to the internal implementation, e.g. you use fixed buffer size, this will limit the future extension. And thus, want to know the need to bring back this feature.

Ok, I will refine the implementation to save the length in some reserved field as before.

wjliang:
> Ok, I will refine the implementation to save the length in some reserved field as before.

My concern is these rpmsg APIs may only work with single buffer. E.g. it will not work with chained buffers.
However, the existing RPMsg implementation only use single buffer. Maybe we can restrict those new APIs to single buffer only.

Also please also post the zero-copy to kernel RPMsg, as you are using the reserved fields from the RPMsg header.

Thanks for you use cases list, I just have some questions:
* RPMsg netdev, have you considered virtio net?
* this feature looks like it is for either a system with a very small memory, or for high message passing performance. For huge data, just curious (I haven't done performance measurement), don't you want the payload to be page aligned and cacheable in some cases?

xiaoxiang781216:
> > Ok, I will refine the implementation to save the length in some reserved field as before.
> 
> My concern is these rpmsg APIs may only work with single buffer. E.g. it will not work with chained buffers.
> However, the existing RPMsg implementation only use single buffer. Maybe we can restrict those new APIs to single buffer only.
> 

If we extend the rpmsg to use the chained buffer, we can extend this API too.  Actually, each offical OpenAMP release isn't compatible with the previous one, so I think it isn't a big issue to change this API for the chained buffer.

> Also please also post the zero-copy to kernel RPMsg, as you are using the reserved fields from the RPMsg header.
> 

I will provide one, but I think this change is different from:
https://github.com/OpenAMP/open-amp/pull/155
https://github.com/OpenAMP/open-amp/pull/160
Since the reserved fields is reused as the temporary storage internally, it can work even the kernel side:
1.Reject the zero copy transfer enhancement
2.Kernel use the reserved to pass the side band info to remote

> Thanks for you use cases list, I just have some questions:
> 
> * RPMsg netdev, have you considered virtio net?

Yes, I have, but:
1.viritio support on OpenAMP is still not mature
2.netdev is used between two NuttX instance(not NuttX<->Linux), because NuttX:
  1.Have  the own and unique net driver model which is incompatible with Linux Kernel
  2.NuttX don't support the virtio net driver at all.
  So I have to define the netdev protocol on top of RPMsg.
If the communication is between NuttX<->Linux, VirtioIO net maybe a better opiton.

> * this feature looks like it is for either a system with a very small memory, or for high message passing performance. For huge data, just curious (I haven't done performance measurement), don't you want the payload to be page aligned and cacheable in some cases?

I expect that the message embedded directly in rpmsg buffer is around 1KB, the huge data(several ten/hundred KB) should pass through as a pointer, because:
1.Many service run on rpmsg send the small packet, ony few of them need send the huge data in one transfer. It isn't efficient to reserve the big buffer for the rare usecase.
2.If the service need send the huge data, normally it's will use the special memory manage strategy(e.g. ION) and is impossible to push the data into the buffer directly.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: @wjliang, after review the patch carefully again, I think the change is right for the current implementation.

> you fixed the rpmsg buffer size here, but it is actually up to the allocator,

Yes, rpmsg_virtio_shm_pool support the allocation with different size, but rpmsg master always allocate the fixed size(RPMSG_BUFFER_SIZE) now, so this limitation doesn't lead by this patch, but by the implementation. If master decide to change the buffer size in the future:
1.Negotiate individual buffer size dynamically(https://github.com/OpenAMP/open-amp/pull/155), we can update as:
  vqbuf.len = rvdev->config.rxbuf_size;
2.Support the variable buffer length, we can:
   a.Reserve 32bit space before the buffer, and save the length here, or
   b.Add a new length array, save the length into it by buffer index
We even extend the len of rpmsg_get_tx_payload_buffer as in/out parameter, so the caller can pass in the required size.

> it is not necessary to have fixed size buffer. You will need to also remember the buffer size.

As mention before, this patch is compatible with the current implementation but still easily modify to cover the extension.

xiaoxiang781216:
Rebase to the latest mainline, no function change.

lib/rpmsg/rpmsg_virtio.c
wjliang: > Yes, rpmsg_virtio_shm_pool support the allocation with different size, but rpmsg master always allocate the fixed size(RPMSG_BUFFER_SIZE) now, so this limitation doesn't lead by this patch, but by the implementation. If master decide to change the buffer size in the future:
> 1.Negotiate individual buffer size dynamically(#155), we can update as:
> vqbuf.len = rvdev->config.rxbuf_size;

I think the best is the buffer size is allocated based on the need of message/data, it can be, have default size, if the data is larger, the allocator can decide to enlarge the buffer. I prefer not to add more limitation to the existing implementation.
Even this feature has to assume the buffer size, it is better to fix the size in the new functions rather than change the existing function. As we can consider to conditionally enable the new feature.

xiaoxiang781216: Ok, please the review the new patch I just send out, which don't add more limitation to the existing implemenation.

xiaoxiang781216:
@wjliang the new patch keep len paramter, please

xiaoxiang781216:
@wjliang fix the build break.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -262,96 +267,154 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
...
-	status = metal_io_block_write(io,
-				      metal_io_virt_to_offset(io,
-				      RPMSG_LOCATE_DATA(buffer)),
-				      data, size);
-	RPMSG_ASSERT(status == size, "failed to write buffer\r\n");
 	metal_mutex_acquire(&rdev->lock);
 
 	/* Enqueue buffer on virtqueue. */
-	status = rpmsg_virtio_enqueue_buffer(rvdev, buffer, buff_len, idx);
+	status = rpmsg_virtio_enqueue_buffer(rvdev, hdr, 0, idx);
```
wjliang: should you use len instead of "0"? you can set len default value if len is not set here.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -109,14 +111,16 @@ static int rpmsg_virtio_enqueue_buffer(struct rpmsg_virtio_device *rvdev,
 
 		/* Initialize buffer node */
 		vqbuf.buf = buffer;
-		vqbuf.len = len;
+		vqbuf.len = len ? len : RPMSG_BUFFER_SIZE;
```
wjliang: Just in case of non-zero copy, len "0" indicates error, prefer to set the len in this function's caller. And only limit to use default len in the zero copy functions.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: Your mean 'len' parameter pass from caller? We can't use it here since virito layer need the whole buffer length, not the real playload length.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: Ok.

lib/rpmsg/rpmsg_virtio.c
wjliang: > Your mean 'len' parameter pass from caller? We can't use it here since virito layer need the whole buffer length, not the real playload length.

As both zero copy and non-zero copy calling this function, for non-zero copy case, it can get the buffer length.

xiaoxiang781216: Ok, the new patch remove the dependence between zero copy and non-zero copy. Please take a look.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -262,14 +263,128 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
...
+	uint32_t len;
+
+	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
+	rp_hdr = RPMSG_LOCATE_HDR(rxbuf);
+	/* The reserved field contains buffer index */
+	idx = rp_hdr->reserved;
+
+	metal_mutex_acquire(&rdev->lock);
+	/* Return buffer on virtqueue. */
+	len = virtqueue_get_buffer_length(rvdev->rvq, idx);
```
wjliang: if it is the virtio master, you cannot keep the idx for returning the buffer.
It will be ok for slave. On the other hand, I think it ok to assuming that the buffer size are always the same for master RX.
I am ok with this change. Please add comments to this line to indicate you can only use the idx to locate the old buffer if it is  virtio slave,  on the master side, the buffer sizes on the RX are the same.

xiaoxiang781216: > if it is the virtio master, you cannot keep the idx for returning the buffer.

It's depend on where the buffer come from:
1.The idx is invalid if the buffer come from share memory buffer pool
2.The idx is valid if the buffer come from tx used queue

> It will be ok for slave. On the other hand, I think it ok to assuming that the buffer size are always the same for master RX.

Yes, because these two points:
1.idx isn't always valid for master side
2.the buffer size could be assumed same in master side
The previous patch remove the buffer length argument.

> I am ok with this change. Please add comments to this line to indicate you can only use the idx to locate the old buffer if it is virtio slave, on the master side, the buffer sizes on the RX are the same.

Ok.

wjliang: > The previous patch remove the buffer length argument.

I think the len argument should be there, just if it is slave, as it is not the one to allocate the buffer, it cannot control the length of the buffer, it can get it from the index. For master, it depends on allocator, and i think usually, it is a fixed size.

lib/include/openamp/rpmsg.h
arnop2: i suggest to put this a separate commit

wjliang:
> Concerning the patch itself: seem correct. Just one remark:
> Something may be missing to release all buffers on application crash or on service destruction ( local or remote). But this would means that reserved buffers should be associated to the ept...

Good point, just in some cases of crashing, one end may not even able to do any cleaning.
Do you think it good to make this as another OpenAMP issue for recovery from failure?

lib/include/openamp/rpmsg.h
xiaoxiang781216: Ok, I will split this change to another patch

xiaoxiang781216:
> For me this patch answers to a limited use cases vs big data need:
> 
> * DMA transfer for copies ( cyclic mode)
> * big buffers
> * buffer reuses (data processing on same buffer).
>   ...

Since many above cases normally involve the dedicated memory management, I expect:
1.Use rpmsg to pass the pointer to these special memory block or
2.Develop the driver on virtio layer directly to remove the buffer management limitation imposed by rpmsg.

>   I don't want to block it but just a concerns vs a global solution base on a service on top of rpmsg...
>   i suppose that you define some shared buffers for the Sound Open Firmware use cases, and rpmsg are used for control, right?

Yes, the audio data buffer is preallocated after the hardware parameter is lock down and recycle in the whole playback/capture stage. The buffer location exchange when the state change from stop/pause to run and read/write directly by DSP, from that the rpmsg is just used to notify the buffer position, no real audo data send/receive through IPC.
 
> Concerning the patch itself: seem correct. Just one remark:
> Something may be missing to release all buffers on application crash or on service destruction ( local or remote). But this would means that reserved buffers should be associated to the ept...

Yes, it's a problem if application get a buffer but crash before return back. But since the memory protection is very weak for most OpenAMP usage, the action can be taken is reboot MCU/DSP if some application crash.

xiaoxiang781216:
@arnop2, the new pull request split the change into two patches.

arnop2:
protection is very weak for most OpenAMP usage, the action can be taken is to reboot MCU/DSP if some application crash, so what we need here is to teardown all resource and restart if the peer happen reboot.
This could be OK for a MCU/DSP dedicated for a feature. But OpenAMP can also run on the main MPU or  MCU processor... with multi services on top of RPMsg. In this case you would have to stop all the services and reinit the rpmsg.

xiaoxiang781216:
> protection is very weak for most OpenAMP usage, the action can be taken is to reboot MCU/DSP if some application crash, so what we need here is to teardown all resource and restart if the peer happen reboot.
> This could be OK for a MCU/DSP dedicated for a feature. But OpenAMP can also run on the main MPU or MCU processor... with multi services on top of RPMsg. In this case you would have to stop all the services and reinit the rpmsg.

Ok, we need take more time to consider how to handle the crash case gracefully, maybe like @wjliang said, open a new issue to track the problem.

xiaoxiang781216:
@wjliang how about this patch?

arnopo:
@xiaoxiang781216 
As discussed patch seems ok, but it introduce new interfaces. 
To be merged we need a way to test it en ensure non regression in future. 
So please could you provide associated [exemple](https://github.com/OpenAMP/open-amp/tree/master/apps/examples).

xiaoxiang781216:
> @xiaoxiang781216
> As discussed patch seems ok, but it introduce new interfaces.
> To be merged we need a way to test it en ensure non regression in future.
> So please could you provide associated [exemple](https://github.com/OpenAMP/open-amp/tree/master/apps/examples).

Sure, I will find some free time in the next couple week to add a new example.

arnopo:
Hi @xiaoxiang781216 
We plan a new release next month, would be nice to integrate this PR with associated example in this release.
do you think you will find time to work on it?

xiaoxiang781216:
> Hi @xiaoxiang781216
> We plan a new release next month, would be nice to integrate this PR with associated example in this release.
> do you think you will find time to work on it?

Sure, I will provide some example before end of March.

arnopo:
> > Hi @xiaoxiang781216
> > We plan a new release next month, would be nice to integrate this PR with associated example in this release.
> > do you think you will find time to work on it?
> 
> Sure, I will provide some example before end of March.

Great!

xiaoxiang781216:
@arnopo I adapter ping demo to utilize zero copy transfer, please take a look.

apps/tests/msg/rpmsg-ping.c
```diff
@@ -134,6 +125,14 @@ int app (struct rpmsg_device *rdev, void *priv)
 
 	LPRINTF("RPMSG endpoint is binded with remote.\r\n");
 	for (i = 0, size = PAYLOAD_MIN_SIZE; i < num_payloads; i++, size++) {
+		struct _payload *i_payload;
+
```
arnopo, line 129: i would prefer that you duplicate the exemple, instead of modifying it.
Create a rpmsg_ping_nc.c for the master and a rpmsg_echo_nc.c  in application for the salve

arnopo:
This PR need few reworks but LGTM after that.

FYI some additional updates are needed related to coding rules put in place since last release.
Please have a look to the [checkpatch result](https://github.com/OpenAMP/open-amp/pull/162/checks?check_run_id=619426784)
No error but some warnings that should be fixed to respect coding rules we just put in place ( I need to find a way to a status failed on warnings)
gitlint --commits  e2ebdda75394b06004a7377dac70e5d0af49f823-HEAD also reports 2 issues on you commit messages. 

Thanks!

lib/include/openamp/rpmsg.h
```diff
@@ -268,6 +279,165 @@ static inline int rpmsg_trysend_offchannel(struct rpmsg_endpoint *ept,
...
+ * @see rpmsg_get_tx_payload_buffer
+ * @see rpmsg_send_offchannel_nocopy
+ * @see rpmsg_sendto_nocopy
+ */
+static inline int rpmsg_send_nocopy(struct rpmsg_endpoint *ept, const void *data, int len)
+{
+	return rpmsg_send_offchannel_nocopy(ept, ept->addr,
+					    ept->dest_addr, data, len);
+}
+
```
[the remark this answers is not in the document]
edmooring: I just tried building rpmsg-echo.out from the proposed change and it added less than 1K to the text and data. I'm not sure it's a problem.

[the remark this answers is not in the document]
arnopo: For the context: one target of the 2018 work resulting in the 2018.10 release was to be able to generate a library ( open-AMP + lbmetal) with a foot print less or close to 3K, at least for the slave implementation.

A test on a zephyr sample gives me for the release 2020.04 ( slave only)
 - open-amp footprint  2610
 - lib-metal footprint 968
With this PR the Open-AMP foot print is increased to 3058.
Even if the measure depends on context this triggers me an Alarm...
So we will have to discuss how to limit the library foot print, to be usable for processor 
with low memory.
Anyway i propose to discuss this point in another thread

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -418,8 +533,15 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 
 		metal_mutex_acquire(&rdev->lock);
 
-		/* Return used buffers. */
-		rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
+		/* Check whether callback wants to hold buffer */
+		if (rp_hdr->reserved & RPMSG_BUF_HELD) {
+			/* 'rp_hdr->reserved' field is now used as storage for
+			 * 'idx' to release buffer later */
```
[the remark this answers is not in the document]
xiaoxiang781216: Yes, it's hard to find a better location to save this information. But I suppose that the flags is only used inside virtio/rpmsg even the new flag is introduced in the future. So we can consume the new flag in rpmsg_virtio_rx_callback immediately, or save into other location.

lib/rpmsg/rpmsg_internal.h
```diff
@@ -35,6 +35,9 @@ extern "C" {
 	} while (0)
 #endif
 
+#define RPMSG_BUF_HELD (1U << 31) /* Flag to suggest to hold the buffer */
```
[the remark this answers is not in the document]
xiaoxiang781216: The resolved field is used in two phase:
1.

> This PR need few reworks but LGTM after that.
> 
> FYI some additional updates are needed related to coding rules put in place since last release.
> Please have a look to the [checkpatch result](https://github.com/OpenAMP/open-amp/pull/162/checks?check_run_id=619426784)
> No error but some warnings that should be fixed to respect coding rules we just put in place ( I need to find a way to a status failed on warnings)
> gitlint --commits e2ebdda75394b06004a7377dac70e5d0af49f823-HEAD also reports 2 issues on you commit messages.
> 
> Thanks!

Sure, I will provide the update to fix the style issue.

xiaoxiang781216:
> This PR need few reworks but LGTM after that.
> 
> FYI some additional updates are needed related to coding rules put in place since last release.
> Please have a look to the [checkpatch result](https://github.com/OpenAMP/open-amp/pull/162/checks?check_run_id=619426784)
> No error but some warnings that should be fixed to respect coding rules we just put in place ( I need to find a way to a status failed on warnings)
> gitlint --commits e2ebdda75394b06004a7377dac70e5d0af49f823-HEAD also reports 2 issues on you commit messages.
> 
> Thanks!

Sure, I will provide the update to fix the style issue.

apps/tests/msg/rpmsg-ping.c
xiaoxiang781216: OK.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -260,6 +261,118 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
...
+						uint32_t *len, int wait)
+{
+	struct rpmsg_virtio_device *rvdev;
+	struct rpmsg_hdr *rp_hdr;
+	uint16_t idx;
+	int tick_count;
+
+	/* Get the associated remote device for channel. */
+	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
+
```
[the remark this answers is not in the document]
xiaoxiang781216: Should this be done by remoteproc_create_virtio?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -260,6 +261,118 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
...
+	len = virtqueue_get_buffer_length(rvdev->svq, idx);
+	status = rpmsg_virtio_enqueue_buffer(rvdev, hdr, len, idx);
+	RPMSG_ASSERT(status == VQUEUE_SUCCESS, "failed to enqueue buffer\r\n");
+	/* Let the other side know that there is a job to process. */
+	virtqueue_kick(rvdev->svq);
+
+	metal_mutex_release(&rdev->lock);
+
+	return len;
+}
```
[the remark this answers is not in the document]
xiaoxiang781216: Ok.

xiaoxiang781216:
@arnopo and @edmooring address all issue.

lib/include/openamp/rpmsg.h
```diff
@@ -81,12 +81,19 @@ struct rpmsg_endpoint {
...
+ * @hold_rx_buffer: hold RPMsg RX buffer
+ * @release_rx_buffer: release RPMsg RX buffer
+ * @get_tx_payload_buffer: get RPMsg TX buffer
+ * @send_offchannel_nocopy: send RPMsg data without copy
  */
 struct rpmsg_device_ops {
-	int (*send_offchannel_raw)(struct rpmsg_device *rdev,
-				   uint32_t src, uint32_t dst,
-				   const void *data, int len, int wait);
+	void (*hold_rx_buffer)(struct rpmsg_device *rdev, void *rxbuf);
```
arnopo, line 90: keep send_offchannel_raw ops. we have to preserve this API for sustainability. then to limit lib footprint impact. it will more simple to split legacy and no copy support.

apps/tests/msg/rpmsg-nocopy-ping.c
```diff
@@ -0,0 +1,199 @@
...
+			ret = 0;
+		}
+	}
+
+	LPRINTF("Stopping application...\r\n");
+	platform_cleanup(platform);
+
+	return ret;
+}
+
```
arnopo, line 199: Test is functional on my side, thanks

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -260,39 +261,50 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
...
-	struct rpmsg_hdr rp_hdr;
-	void *buffer = NULL;
+	struct rpmsg_hdr *rp_hdr;
+	uint16_t idx;
+	uint32_t len;
+
+	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
+	rp_hdr = RPMSG_LOCATE_HDR(rxbuf);
+	/* The reserved field contains buffer index */
+	idx = rp_hdr->reserved;
```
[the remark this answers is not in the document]
xiaoxiang781216: No, rpmsg_virtio_send_offchannel_nocopy doesn't protect too.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -355,7 +356,7 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 	hdr = RPMSG_LOCATE_HDR(data);
 	/* The reserved field contains buffer index */
 	idx = hdr->reserved;
-
+#if 0
```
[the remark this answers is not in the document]
xiaoxiang781216: The test can't pass the test after I change the rpmsg_virtio_send_offchannel_raw to reuse the nocopy API, I am debugging this problem, so please ignore this PR until I find the root cause.

vishnu-banavath:
Arnaud Pouliquen, just checking my understanding here. How likely or unlikely will these be merged? I'm asking this question because, based on the decision we can plan our work or come up with any other alternate solution.

Thanks

arnopo:
Hello,

Integrating the zero copy feature is in our plan.
But it is too late to integrate this PR in the coming release ( planned this month). We hope to integrate the feature for the next one
For the current PR some works are remaining:

  *   Fix segmentation fault issue.
  *   Limit the code size increase associated to the introduction of this PR. We have to find the best way to add the feature while limiting the impact of the footprint.

Any way, If you have any alternative solution to propose in a PR, or if you want to contribute to this one, don’t hesitate.

Regards
Arnaud

lib/rpmsg/rpmsg_virtio.c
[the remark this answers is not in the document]
kr-satish: I tried these patches and found a problem which I think also is the root cause for your segment fault.

Following line, inside the function "rpmsg_virtio_send_offchannel_nocopy" causes the problem:
buff_len = virtqueue_get_buffer_length(rvdev->svq, idx);

virtqueue_get_buffer_length reads buffer length from the buffer descriptor. When a buffer is allocated by rpmsg_virtio_shm_pool_get_buffer (inside function rpmsg_virtio_get_tx_buffer) instead of from used-ring, reading the length of the buffer from the buffer descriptor is not correct. As a result, whenever a Master sends the first message, the length of the buffer return by this function is observed to be ZERO. Which leads to a segment fault in subsequent message sends.

As a temporary fix, you can replace the above line with the following line. There should not be a segment fault after this change.
buff_len = RPMSG_BUFFER_SIZE;

I am also putting a comment on the review to trek this mistake.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -260,92 +261,119 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
...
-	status = metal_io_block_write(io,
-				      metal_io_virt_to_offset(io,
-						RPMSG_LOCATE_DATA(buffer)),
-				      data, len);
-	RPMSG_ASSERT(status == len, "failed to write buffer\r\n");
 	metal_mutex_acquire(&rdev->lock);
 
 	/* Enqueue buffer on virtqueue. */
-	status = rpmsg_virtio_enqueue_buffer(rvdev, buffer, buff_len, idx);
+	buff_len = virtqueue_get_buffer_length(rvdev->svq, idx);
```
kr-satish, line 376: Always reading the length of the buffer from the buffer descriptor is not correct. When a buffer is allocated for the first time by the function rpmsg_virtio_shm_pool_get_buffer, the buffer descriptor does not exist.

This is the cause of the segment fault.

lib/rpmsg/rpmsg_virtio.c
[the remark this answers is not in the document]
xiaoxiang781216: Thanks, @kr-satish, I will try your suggestion soon.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -396,6 +468,8 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 	metal_mutex_release(&rdev->lock);
 
 	while (rp_hdr) {
+		rp_hdr->reserved = 0;
```
kr-satish, line 504: There is a bug here. Consider the following scenario:-
a. Master sends a message to the remote.
b. The remote callback function is called.
c. Remote user holds the buffer but also decides to release the buffer before callback returns.

The bug is that in the above scenario, the rpmsg_virtio_release_rx_buffer function always free buffer with index zero because of this line.

This is a valid scenario, and I think the above line should be changed to:-
rp_hdr->reserved = idx;

And the function rpmsg_virtio_hold_rx_buffer should not overwrite this information by taking a bitwise OR as shown below:-
-	rp_hdr->reserved = RPMSG_BUF_HELD;
+	rp_hdr->reserved |= RPMSG_BUF_HELD;

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -147,6 +147,7 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
 			data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
```
kr-satish, line 148: @xiaoxiang781216 @arnopo 

In the case of master, the function rpmsg_virtio_get_tx_buffer keeps carving out a new buffer from shared memory if no buffer is found in the used-ring. If the remote receiver continuously holds buffers on the other side, the master will keep creating new buffers. I don't see a protection mechanism placed anywhere in the code which disallows more buffer creation than the available buffer descriptors. I have produced this condition quite easily by holding buffers at the remote side indefinitely. Eventually, when buffers are released by the remote, the indexes all went crazy which leads to a segment fault.

From remote to master communication there is no problem because all buffers are created by the master for remote at the time of rpmsg device initialization. And the buffers remain the same as the number of descriptors.

I don't understand the reason, why for Master the code workflow of buffer creation is different? Ideally for master also, all buffers should be created at the time of initialization. Then there is no need to create new buffers inside this function.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: @kr-satish good finding, I will fix it in the update.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: > @xiaoxiang781216 @arnopo
> 
> In the case of master, the function rpmsg_virtio_get_tx_buffer keeps carving out a new buffer from shared memory if no buffer is found in the used-ring. If the remote receiver continuously holds buffers on the other side, the master will keep creating new buffers. I don't see a protection mechanism placed anywhere in the code which disallows more buffer creation than the available buffer descriptors. I have produced this condition quite easily by holding buffers at the remote side indefinitely. Eventually, when buffers are released by the remote, the indexes all went crazy which leads to a segment fault.
> 

Yes, this is an issue, I will try to find a method to fix it. Thanks for report.

> From remote to master communication there is no problem because all buffers are created by the master for remote at the time of rpmsg device initialization. And the buffers remain the same as the number of descriptors.
> 

> I don't understand the reason, why for Master the code workflow of buffer creation is different? Ideally for master also, all buffers should be created at the time of initialization. Then there is no need to create new buffers inside this function.

Yes, the master could allocate all buffers in the initialization, but master has to manage these buffer with link list manually, I guss this is one of reason why the allocation delay to rpmsg_virtio_get_tx_buffer.

xiaoxiang781216:
@arnopo the sgement fault is fixed. @kr-satish, thanks for your help.

kr-satish:
> @arnopo the sgement fault is fixed. @kr-satish, thanks for your help.

@xiaoxiang781216 Thanks for the patches.

arnopo:
From my point of view the feature and associated patch are Ok.
Now the remaining things is to limit the increase of the library footprint that is also a strong requirement for low memory systems.
I'm going to make an assessment of the memory footprint based on Zephyr sample to evaluate the impact.
i suppose that the addition of new ops in rpmsg_device_ops will be the main contributor of the footprint increase...

If you have some suggestions to minimize the impact don't hesitate...

arnopo:
I performed measurements for some zehyr samples, building for the NXP expresso and STm32MP1 platforms.
The compiler is GCC with the optimization flag -Os. 

I note a library increase of about 10% for the ROM and 16 bytes for the RAM.
That's seems to me very reasonable.

- #### on: https://github.com/zephyrproject-rtos/zephyr/tree/master/samples/subsys/ipc/openamp
	**target**: lpcxpresso54114_m4 (master)
##### w/o zero copy:
	libmetal:	994 bytes 
	openamp:	2378 bytes
##### w zero copy: 
	libmetal:	994 bytes 
	openamp:	2664 bytes (+286)

- ####  Application: https://github.com/zephyrproject-rtos/zephyr/tree/master/samples/subsys/ipc/openamp/remote
	**target**: lpcxpresso54114_m0 (slave)
##### w/o zero copy: 
	libmetal:	806 bytes 
	openamp:	2300 bytes

##### w zero copy: 
	libmetal:	806 bytes 
	openamp:	2552 bytes (+252)

- #### Application: /https://github.com/zephyrproject-rtos/zephyr/tree/master/samples/subsys/ipc/openamp_rsc_table
	**target**: stm32mp157c_dk2(slave)
##### w/o zero copy: 
	libmetal:	968 bytes 
	openamp:	2696 bytes

##### w zero copy: 
	libmetal:	968 bytes 
	openamp:	2868 bytes (+172)

xiaoxiang781216:
The absolute number is  from 172 bytes to 252 bytes. Should we still need add a macro to enable/disable this feature?

arnopo:
> The absolute number is from 172 bytes to 252 bytes. Should we still need add a macro to enable/disable this feature?

I tested using macros, which saves about half the extra footprint but adds complexity.
From my point of view, I'll leave it like that and if optimization is needed later, we'll take care of it if necessary.

arnopo (approved):

arnopo:
Hi @galak 
As the footprint is also a concern for the Zephyr project, could you give your opinion?
Is the increase  of the memory footprint, is reasonable from your point of view?

xiaoxiang781216:
@arnopo how about we merge this PR now? so it get more time to test before release.

arnopo:
> @arnopo how about we merge this PR now? so it get more time to test before release.

I waiting @edmooring and @galak  feedback first. As everyone is busy on there own stuff we need to  allow a reasonable time for review.

arnopo:
Hi @Galak, @edmooring 
Gentle reminder :-)

edmooring (approved):
If @galak is OK with this, I think we're good to go.

arnopo:
Hi @xiaoxiang781216 
I merged the series  with 2 fixes of minor "check" point  reported by the check-patch.
Thanks a lot for this major contribution!

PR #163, opened 2019-02-27
assert only when status < 0 in rpmsg_virtio_rx_callback

xiaoxiang781216:
since the successful return value from many function is greater than zero

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -411,7 +411,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
 					 rp_hdr->len, rp_hdr->src, ept->priv);
 
-			RPMSG_ASSERT(status == RPMSG_SUCCESS,
+			RPMSG_ASSERT(status >= 0,
 				     "unexpected callback status\r\n");
 		}
 
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -45,6 +45,7 @@ extern "C" {
 struct rpmsg_endpoint;
 struct rpmsg_device;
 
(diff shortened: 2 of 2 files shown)


xiaoxiang781216:
@wjliang please review this minor fix.

xiaoxiang781216:

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -411,7 +411,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
 					 rp_hdr->len, rp_hdr->src, ept->priv);
 
-			RPMSG_ASSERT(status == RPMSG_SUCCESS,
+			RPMSG_ASSERT(status >= RPMSG_SUCCESS,
```
wjliang: Could you explain why the status will be > RPMSG_SUCCESS? Thanks.

xiaoxiang781216: For example, at the end of many cb implemenation, it normally send the response to another side by rpmsg_send, but rpmsg_send return the data length for success, negative error code for fail. Without this change, I have to:
int xxx_cb(...)
{
  int ret = rpmsg_send(...);
  return ret >= 0 ? 0 : ret;
}
Instead the more clean code:
int xxx_cb(...)
{
  return rpmsg_send(...);
}

wjliang: I see, just the rpmsg_send() in cb is more like application specific implementation. The cb() should return what's the cb expect to return.
From your change it is strange to see it still considers the return as successful even the return value is not RPMSG_SUCCESS. I think the rpmsg.h should add comments to explain what's the expected result.

xiaoxiang781216: Ok, the new patch add comment just before rpmsg_ept_cb and change RPMSG_SUCCESS to 0.

PR #164, opened 2019-02-27
zero out ept in rpmsg_destroy_ept

xiaoxiang781216:
and don't use zero as the local address, so
it's easier to detect the uninitialized case.

### DIFF
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -15,6 +15,7 @@
 #include <metal/compiler.h>
 #include <metal/mutex.h>
 #include <metal/list.h>
+#include <metal/sleep.h>
 #include <metal/utilities.h>
 #include <string.h>
 #include <stdbool.h>
@@ -25,22 +26,28 @@ extern "C" {
 #endif
 
 /* Configurable parameters */
-#define RPMSG_NAME_SIZE		(32)
-#define RPMSG_ADDR_BMP_SIZE	(128)
+#define RPMSG_NAME_SIZE			(32)
+#define RPMSG_ADDR_BMP_SIZE		(128)
 
-#define RPMSG_NS_EPT_ADDR	(0x35)
-#define RPMSG_ADDR_ANY		0xFFFFFFFF
+#define RPMSG_NS_EPT_ADDR		(0x35)
+#define RPMSG_ADDR_ANY			0xFFFFFFFF
+
+/* Total tick count for 15secs - 1usec tick. */
+#define RPMSG_TICK_COUNT		15000000
+
+/* Time to wait - In multiple of 1 msecs. */
+#define RPMSG_TICKS_PER_INTERVAL	1000
 
 /* Error macros. */
-#define RPMSG_SUCCESS		0
-#define RPMSG_ERROR_BASE	-2000
-#define RPMSG_ERR_NO_MEM	(RPMSG_ERROR_BASE - 1)
-#define RPMSG_ERR_NO_BUFF	(RPMSG_ERROR_BASE - 2)
-#define RPMSG_ERR_PARAM		(RPMSG_ERROR_BASE - 3)
-#define RPMSG_ERR_DEV_STATE	(RPMSG_ERROR_BASE - 4)
-#define RPMSG_ERR_BUFF_SIZE	(RPMSG_ERROR_BASE - 5)
-#define RPMSG_ERR_INIT		(RPMSG_ERROR_BASE - 6)
-#define RPMSG_ERR_ADDR		(RPMSG_ERROR_BASE - 7)
+#define RPMSG_SUCCESS			0
+#define RPMSG_ERROR_BASE		-2000
+#define RPMSG_ERR_NO_MEM		(RPMSG_ERROR_BASE - 1)
+#define RPMSG_ERR_NO_BUFF		(RPMSG_ERROR_BASE - 2)
+#define RPMSG_ERR_PARAM			(RPMSG_ERROR_BASE - 3)
+#define RPMSG_ERR_DEV_STATE		(RPMSG_ERROR_BASE - 4)
+#define RPMSG_ERR_BUFF_SIZE		(RPMSG_ERROR_BASE - 5)
+#define RPMSG_ERR_INIT			(RPMSG_ERROR_BASE - 6)
+#define RPMSG_ERR_ADDR			(RPMSG_ERROR_BASE - 7)
 
 struct rpmsg_endpoint;
 struct rpmsg_device;
@@ -160,10 +167,17 @@ int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
 static inline int rpmsg_send(struct rpmsg_endpoint *ept, const void *data,
 			     int len)
 {
-	if (!is_rpmsg_ept_ready(ept))
-		return RPMSG_ERR_ADDR;
-	return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
-					 len, true);
+	int tc = 0;
+
+	for (; tc < RPMSG_TICK_COUNT; tc += RPMSG_TICKS_PER_INTERVAL) {
+		if (is_rpmsg_ept_ready(ept))
+			return rpmsg_send_offchannel_raw(ept, ept->addr,
+							 ept->dest_addr,
+							 data, len, true);
+		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
+	}
+
+	return RPMSG_ERR_ADDR;
 }
 
 /**
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -111,6 +111,20 @@ struct rpmsg_device {
 };
 
 /**
+ * is_rpmsg_ept_ready - check if the rpmsg endpoint ready to send
+ *
+ * @ept: pointer to rpmsg endpoint
+ *
+ * Returns 1 if the rpmsg endpoint has both local addr and destination
+ * addr set, 0 otherwise
+ */
+static inline unsigned int is_rpmsg_ept_ready(struct rpmsg_endpoint *ept)
+{
+	return (ept->dest_addr != RPMSG_ADDR_ANY) &&
+		(ept->addr != 0) && (ept->addr != RPMSG_ADDR_ANY);
+}
+
+/**
  * rpmsg_send_offchannel_raw() - send a message across to the remote processor,
  * specifying source and destination address.
  * @ept: the rpmsg endpoint
@@ -146,7 +160,7 @@ int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
 static inline int rpmsg_send(struct rpmsg_endpoint *ept, const void *data,
 			     int len)
 {
-	if (ept->dest_addr == RPMSG_ADDR_ANY)
+	if (!is_rpmsg_ept_ready(ept))
 		return RPMSG_ERR_ADDR;
 	return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
 					 len, true);
@@ -216,7 +230,7 @@ static inline int rpmsg_send_offchannel(struct rpmsg_endpoint *ept,
 static inline int rpmsg_trysend(struct rpmsg_endpoint *ept, const void *data,
 				int len)
 {
-	if (ept->dest_addr == RPMSG_ADDR_ANY)
+	if (!is_rpmsg_ept_ready(ept))
 		return RPMSG_ERR_ADDR;
 	return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
 					 len, false);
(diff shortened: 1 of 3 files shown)


xiaoxiang781216:
@wjliang, since ept could be embeded in some big structure and then normally zero by zalloc/memset now, it's important that is_rpmsg_ept_ready return false before rpmsg_create_ept get called.
note: in the previous verson, we can't embed rpmsg_channel directly, so we can detect this case by comparing channel pointer equals NULL.

xiaoxiang781216:

lib/include/openamp/rpmsg.h
```diff
@@ -344,7 +344,7 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept);
 static inline unsigned int is_rpmsg_ept_ready(struct rpmsg_endpoint *ept)
 {
 	return (ept->dest_addr != RPMSG_ADDR_ANY) &&
-		(ept->addr != RPMSG_ADDR_ANY);
+		(ept->addr != 0) && (ept->addr != RPMSG_ADDR_ANY);
```
wjliang: endpoint address can be 0.

lib/rpmsg/rpmsg.c
```diff
@@ -27,7 +27,7 @@ static uint32_t rpmsg_get_address(unsigned long *bitmap, int size)
 	unsigned int addr = RPMSG_ADDR_ANY;
 	unsigned int nextbit;
 
-	nextbit = metal_bitmap_next_clear_bit(bitmap, 0, size);
+	nextbit = metal_bitmap_next_clear_bit(bitmap, 1, size);
```
wjliang: endpoint address can be 0.

lib/rpmsg/rpmsg.c
```diff
@@ -262,4 +262,5 @@ void rpmsg_destroy_ept(struct rpmsg_endpoint *ept)
 	metal_mutex_acquire(&rdev->lock);
 	rpmsg_unregister_endpoint(ept);
 	metal_mutex_release(&rdev->lock);
+	memset(ept, 0, sizeof(*ept));
```
wjliang: can set the endpoint address to ANY.
If RPMsg endpoint is statically defined, how about using ANY address for initialization?

lib/rpmsg/rpmsg.c
xiaoxiang781216: The adress is allocated from 1 by the above change, so the LOCAL address is impossible to zero now.

lib/include/openamp/rpmsg.h
xiaoxiang781216: > endpoint address can be 0. It is possible for existing RPMsg user to use OpenAMP implementation on one end, and use non RPMsg implementation on the other end.

Yes, that's why I just check the local adress(not dest_addr) equals zero, since the other end may allocate zero as a normal address by non RPMsg impelementation. 
But with the change at line 30, OpenAMP doesn't allocate zero as the local address anymore.

lib/rpmsg/rpmsg.c
xiaoxiang781216: > can set the endpoint address to ANY.
> If RPMsg endpoint is statically defined, how about using ANY address for initialization?

Yes, we can require the user set the address to ANY after allocation and before rpmsg_create_ept, but the user may(often?) forget to do it,  it's better to use zero here:
1.global/static variable is zero by compiler/linker automatically
2.It's common practice to memset(or zalloc) the allocated object
so no any special initializition is required anymore.

lib/include/openamp/rpmsg.h
wjliang: I have  seen Linux kernel reserved some endpoint address for RPMsg services (0 to 1024), but Linux kernel side doesn't how it expects those reserved services endpoint to setup. And thus, I have some concerns on we have some OpenAMP specific reservation due to implementation convenience.

I am fine with this change, want to see other existing users' input.

wjliang:
@arnop2, @galak, please review

lib/rpmsg/rpmsg.c
arnop2: With you patch the address 0 is forbidden. So that means that the "well know" service will be forbidden if associated to address 0. I understand the simplification from a user point of view but the rpmsg_init_ept helper should answer to this.

lib/rpmsg/rpmsg.c
arnop2: Please don't  clean the endpoint here, better to do it in  rpmsg_init_ept. 
for instance  if you create an ept with specific addresses  you can want to reuse the ept later ( for instance suspend resume usecases).

xiaoxiang781216: > Please don't clean the endpoint here, better to do it in rpmsg_init_ept.
> for instance if you create an ept with specific addresses you can want to reuse the ept later ( for instance suspend resume usecases).

But, if the user call rpmsg_destroy_ept, should they call rpmsg_create_ept again?
Actually, I want to block this type of error too(call rpmsg_destroy_ept, but not call rpmsg_create_ept).

lib/rpmsg/rpmsg.c
xiaoxiang781216: > With you patch the address 0 is forbidden. So that means that the "well know" service will be forbidden if associated to address 0. I understand the simplification from a user point of view but the rpmsg_init_ept helper should answer to this.

Ok, I will add a check in rpmsg_create_ept to return RPMSG_ERR_PARAM for zero local address.

xiaoxiang781216:
The new patch check zero in rpmsg_create_ept to make the convention more consistent.

lib/rpmsg/rpmsg.c
arnop2: > > Please don't clean the endpoint here, better to do it in rpmsg_init_ept.
> > for instance if you create an ept with specific addresses you can want to reuse the ept later ( for instance suspend resume usecases).
> 
> But, if the user call rpmsg_destroy_ept, should they call rpmsg_create_ept again?
Nothing should prevent this. For instance if application would update the local address or the call backs.

> Actually, I want to block this type of error too(call rpmsg_destroy_ept, but not call rpmsg_create_ept yet).
> And rpmsg_destroy_ept has more side effect than zero out memory:
> 1.Send RPMSG_NS_DESTROY message
> 2.Remove the allocation bit in rpmsg_unregister_endpoint
> Both these actions need recover by call rpmsg_create_ept, then we need pass the address to it anyway.
Why this should happen? except a bug in application...if application call rpmsg_init_ept to create the endpoint this should not occurs. My concern is that this add a restriction in OpenAMP to facilitate application robustness... 

Please tell me if I'm wrong, but here's what you want to solve is the problem when an application calls rpmsg_create_ept without calling rpmsg_init_ept first.
In this case, the ept structure is set to 0 and since the local address is 0, there is no address negotiation.

In this case what about testing the ept->cb != NULL instead?  this should prevent a bad initialization in  rpmsg_destroy_ept and rpmsg_destroy_ept function...
Today this callback is mandatory ( no protection in rpmsg_virtio_rx_callback), but no test on it...

xiaoxiang781216: > > > Please don't clean the endpoint here, better to do it in rpmsg_init_ept.
> > > for instance if you create an ept with specific addresses you can want to reuse the ept later ( for instance suspend resume usecases).

memory clean only happen in rpmsg_destroy_ept, I can't understand why resume/suspend will call rpmsg_destroy_ept. Even suspend really need call rpmsg_destroy_ept, resume must call rpmsg_create_ept again to recover the side effect done by rpmsg_destroy_ept to make the endpoint work again. As you can see both rpmsg_create_ept and rpmsg_init_ept need the caller pass the local and remote address, the caller should save the address somewhere anyway.

> > 
> > 
> > But, if the user call rpmsg_destroy_ept, should they call rpmsg_create_ept again?
> > Nothing should prevent this. For instance if application would update the local address or the call backs.

But, application shouldn't directly modify the local address since rpmsg_device track the allocation in bitmap field, the unexpected modification make the tracking inaccuracy.

> 
> > Actually, I want to block this type of error too(call rpmsg_destroy_ept, but not call rpmsg_create_ept yet).
> > And rpmsg_destroy_ept has more side effect than zero out memory:
> > 1.Send RPMSG_NS_DESTROY message
> > 2.Remove the allocation bit in rpmsg_unregister_endpoint
> > Both these actions need recover by call rpmsg_create_ept, then we need pass the address to it anyway.
> > Why this should happen? except a bug in application...if application call rpmsg_init_ept to create the endpoint this should not occurs. My concern is that this add a restriction in OpenAMP to facilitate application robustness...
> 
> Please tell me if I'm wrong, but here's what you want to solve is the problem when an application calls rpmsg_create_ept without calling rpmsg_init_ept first.

Since rpmsg_create_ept call rpmsg_init_ept internally, why we need call rpmsg_init_ept then rpmsg_create_ept? is it enough to only call rpmsg_create_ept here?
After grep rpmsg_create_ept and rpmsg_init_ept, you can see all applications just call rpmsg_create_ept, only two places call rpmsg_init_ept:
1.rpmsg_create_ept
2.rpmsg_init_vdev for NS endpoint
So what my patch try to resolve is:
1.rpmsg_endpoint allocate dynamiclly or static
2.before call rpmsg_create_ept
and then application call is_rpmsg_ept_ready, we should return fail here.

> In this case, the ept structure is set to 0 and since the local address is 0, there is no address negotiation.
> 
> In this case what about testing the ept->cb != NULL instead? this should prevent a bad initialization in rpmsg_destroy_ept and rpmsg_destroy_ept function...
> Today this callback is mandatory ( no protection in rpmsg_virtio_rx_callback), but no test on it...

Yes, we can check cb instead, but since is_rpmsg_ept_ready only check the address, I want to make the check more consistent.

arnop2: > > In this case, the ept structure is set to 0 and since the local address is 0, there is no address negotiation.
> > In this case what about testing the ept->cb != NULL instead? this should prevent a bad initialization in rpmsg_destroy_ept and rpmsg_destroy_ept function...
> > Today this callback is mandatory ( no protection in rpmsg_virtio_rx_callback), but no test on it...
> 
> Yes, we can check cb instead, but since is_rpmsg_ept_ready only check the address, I want to make the check more consistent.
No sure that this API is really used... @wjliang: you confirm? 
Have you a use case that justifies the call of this function with a bad ept, after an error returned by the rpmsg_create_ept?

I would like to be sure that we do not introduce restriction in OpenAMP just to help user to debug its application...

xiaoxiang781216: > > > In this case, the ept structure is set to 0 and since the local address is 0, there is no address negotiation.
> > > In this case what about testing the ept->cb != NULL instead? this should prevent a bad initialization in rpmsg_destroy_ept and rpmsg_destroy_ept function...
> > > Today this callback is mandatory ( no protection in rpmsg_virtio_rx_callback), but no test on it...
> > 
> > 
> > Yes, we can check cb instead, but since is_rpmsg_ept_ready only check the address, I want to make the check more consistent.
> > No sure that this API is really used... @wjliang: you confirm?

My next pull request will use this API, and in our rpmsg driver code also use this API too.

> > Have you a use case that justifies the call of this function with a bad ept, after an error returned by the rpmsg_create_ept?
> 

This patch just fix is_rpmsg_ept_ready return true if it get called before rpmsg_create_ept, since rpmsg_endpoint:
1.normally define as global or static variables
2.or directly embed in a big structure
In both case, rpmsg_endpoint normally initialize to zero, and can't check the validity by compare null pointer like the rpmsg_channel in old version.

> I would like to be sure that we do not introduce restriction in OpenAMP just to help user to debug its application...

The only restriction is that local zero address can't be used anymore, I think it isn't a big issue since:
1.Linux doesn't use zero address too(actually 0-1023 isn't used except 0x35)
2.Zero is too special to use as a valid address just like nobody think zero is a valid pointer
3.The remote side could still use zero address if they really want

arnop2: So is_rpmsg_ept_ready should be kept. 
Now  if you add the "ept->cb != NULL " ( that is missing anyway) in rpmsg_create_ept, this should prevent from a call of rpmsg_create_ept without ept initialization.
In case of error retuurned by rpmsg_create_ept, calling  is_rpmsg_ept_ready is not correct. Adding test on null address seems here over-protection.
On the other size:  if some users has implemented this address as a table index, this patch breaks the legacy
On Linux side:  0 to 1023 address seems not forbidden but reserved for pre-allocated address, means that client could create an endpoint with source address = 0. In this case this could also break the legacy.
For these reasons, I suggest not to add the 0 address restriction but only a check on the CB.
Anyway i'm not the maintainer so let @wjliang decide.

wjliang: I can see the point on @xiaoxiang781216 's purpose, "0" is generic default value for static initialization. And I also agree with @arnop2 to prefer not to add restriction.
This change will only impact the application needs dynamic endpoints allocation, for static endpoints, it can still be 0.  Unless, there are some application strictly expect the endpoint comes from the name service needs to be 0. And thus, I am fine with the idea, cannot see it  breaks. But as this does add restriction, I prefer to get all the reviewers agreement.

But please take @arnop2's suggestion not to set the endpoints address to 0 in destroy, as some user application already assume the previous allocated endpoint address are preserved in the endpoint data struct. You can memset it to 0 after you call destroy.

xiaoxiang781216: Ok, I will remove zero in the next patch

xiaoxiang781216:
@wjliang zero memory is removed in the new patch.

PR #169, opened 2019-03-25
remoteproc: add get_mem callback to remoteproc_ops

xiaoxiang781216:
1.change va from output to input/output
2.remoteproc_get_io_xxx fallback to mmap callback

### DIFF
--- a/lib/include/openamp/remoteproc.h
+++ b/lib/include/openamp/remoteproc.h
@@ -395,6 +395,7 @@ struct remoteproc {
  *        memory may not be off.
  * @shutdown: shutdown the remoteproc and release its resources.
  * @notify: notify the remote
+ * @get_mem: get remoteproc memory I/O region.
  */
 struct remoteproc_ops {
 	struct remoteproc *(*init)(struct remoteproc *rproc,
@@ -410,6 +411,27 @@ struct remoteproc_ops {
 	int (*stop)(struct remoteproc *rproc);
 	int (*shutdown)(struct remoteproc *rproc);
 	int (*notify)(struct remoteproc *rproc, uint32_t id);
+	/**
+	 * get_mem
+	 *
+	 * get remoteproc memory I/O region by either name, virtual
+	 * address, physical address or device address.
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 * @name - memory name
+	 * @pa - physical address
+	 * @da - device address
+	 * @va - virtual address
+	 * @size - memory size
+	 *
+	 * @returns remoteproc memory pointed by buf if success, otherwise NULL
+	 */
+	struct remoteproc_mem *(*get_mem)(struct remoteproc *rproc,
+					  const char *name,
+					  metal_phys_addr_t pa,
+					  metal_phys_addr_t da,
+					  void *va, size_t size,
+					  struct remoteproc_mem *buf);
 };
 
 /* Remoteproc error codes */
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -29,10 +29,11 @@ remoteproc_check_fw_format(const void *img_data, size_t img_len)
 		return NULL;
 }
 
+/* try the internal list added by remoteproc_add_mem first and then get_mem callback */
 static struct remoteproc_mem *
 remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 		   metal_phys_addr_t pa, metal_phys_addr_t da,
-		   void *va, size_t size)
+		   void *va, size_t size, struct remoteproc_mem *buf)
 {
 	struct metal_list *node;
 	struct remoteproc_mem *mem;
@@ -72,7 +73,11 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 			return NULL;
 		}
 	}
-	return NULL;
+
+	if (!rproc->ops->get_mem)
+		return NULL;
+
+	return rproc->ops->get_mem(rproc, name, pa, da, va, size, buf);
 }
 
 static metal_phys_addr_t
@@ -282,9 +287,10 @@ remoteproc_get_io_with_name(struct remoteproc *rproc,
 			    const char *name)
 {
 	struct remoteproc_mem *mem;
+	struct remoteproc_mem buf;
 
 	mem = remoteproc_get_mem(rproc, name,
-				 METAL_BAD_PHYS, METAL_BAD_PHYS, NULL, 0);
+				 METAL_BAD_PHYS, METAL_BAD_PHYS, NULL, 0, &buf);
 	if (mem)
 		return mem->io;
 	else
@@ -296,8 +302,9 @@ remoteproc_get_io_with_pa(struct remoteproc *rproc,
 			  metal_phys_addr_t pa)
 {
 	struct remoteproc_mem *mem;
+	struct remoteproc_mem buf;
 
-	mem = remoteproc_get_mem(rproc, NULL, pa, METAL_BAD_PHYS, NULL, 0);
+	mem = remoteproc_get_mem(rproc, NULL, pa, METAL_BAD_PHYS, NULL, 0, &buf);
 	if (mem)
 		return mem->io;
 	else
@@ -310,8 +317,9 @@ remoteproc_get_io_with_da(struct remoteproc *rproc,
 			  unsigned long *offset)
 {
 	struct remoteproc_mem *mem;
+	struct remoteproc_mem buf;
 
-	mem = remoteproc_get_mem(rproc, NULL, METAL_BAD_PHYS, da, NULL, 0);
+	mem = remoteproc_get_mem(rproc, NULL, METAL_BAD_PHYS, da, NULL, 0, &buf);
 	if (mem) {
 		struct metal_io_region *io;
 		metal_phys_addr_t pa;
@@ -329,9 +337,10 @@ struct metal_io_region *
 remoteproc_get_io_with_va(struct remoteproc *rproc, void *va)
 {
 	struct remoteproc_mem *mem;
+	struct remoteproc_mem buf;
 
 	mem = remoteproc_get_mem(rproc, NULL, METAL_BAD_PHYS, METAL_BAD_PHYS,
-				 va, 0);
+				 va, 0, &buf);
 	if (mem)
 		return mem->io;
 	else
@@ -346,6 +355,7 @@ void *remoteproc_mmap(struct remoteproc *rproc,
 	void *va = NULL;
 	metal_phys_addr_t lpa, lda;
 	struct remoteproc_mem *mem;
+	struct remoteproc_mem buf;
 
 	if (!rproc)
 		return NULL;
@@ -359,7 +369,7 @@ void *remoteproc_mmap(struct remoteproc *rproc,
 		lda =  *da;
 	else
 		lda = METAL_BAD_PHYS;
-	mem = remoteproc_get_mem(rproc, NULL, lpa, lda, NULL, size);
+	mem = remoteproc_get_mem(rproc, NULL, lpa, lda, NULL, size, &buf);
 	if (mem) {
 		if (lpa != METAL_BAD_PHYS)
 			lda = remoteproc_patoda(mem, lpa);


xiaoxiang781216:
@wjliang this patch ensure all address conversion fallback to mmap callback, so all remoteproc_get_io_with_xxx has the same behaviour as remoteproc_mmap.

wjliang:
if the metal_io_region is always required, why needs to have va to pa/da conversion in this API? This APIs are already used in apps, it will break the existing application.

You can consider another wrapper for va to da.

xiaoxiang781216:
> if the metal_io_region is always required, why needs to have va to pa/da conversion in this API?

1.metal_io_region is a returned value here. If we just have va, how can we get io region first?
2.remoteproc_set_rsc_table get io region through remoteproc_get_io_with_va. The old impelementation just search the registered remoteproc_mem list, which mean that mmap callback can't get the chance to hook the process and return/add the new remoteproc_mem as needed.
3. As the commit message said, we need make the behaviour consistent between remoteproc_get_io_with_[v|p|d]a and remoteproc_mmap:
  a.remoteproc_get_io_with_[v|p|d]a just search the registered remoteproc_mem list
  b.remoteproc_mmap search the registered remoteproc_mem list, but fallback to mmap callback if no entry can be found.
We need modify remoteproc_get_io_with_[v|p|d]a like as remoteproc_mmap.

> This APIs are already used in apps, it will break the existing application.

But I already modify all source code to ensure them work with the new prototype, I am tracking OpenAMP for more than three years, all client code need to adapter in each major OpenAMP release to make it work, so is it important to keep the source code compatiblity here?

>
> You can consider another wrapper for va to da.

Do you have the better suggestion? add remoteproc_mmap2 for new va parameter? I can adjust the change as needed if there is a better approach.

apps/examples/load_fw/zynqmp_r5_lcm_rproc_example.c
```diff
@@ -137,86 +137,79 @@ void r5_rproc_remove(struct remoteproc *rproc)
...
-	if (!da || !pa)
-		return NULL;
 	LPRINTF("%s: pa=0x%x, da=0x%x, size=0x%x, atrribute=0x%x\r\n",
 		__func__, *pa, *da, size, attribute);
-	lda = *da;
-	lpa = *pa;
 	if (!attribute)
 		attribute = NORM_SHARED_NCACHE | PRIV_RW_USER_RW;
-	if (lda <= 0x40000) {
+	if (*da <= 0x40000) {
```
bentheredonethat: check if da is invalid pointer before checking contents

xiaoxiang781216: don't check null pointer here since mmap is only called by remoteproc_mmap and this function already ensure va/pa/da never become null.

wjliang:
> 1.metal_io_region is a returned value here. If we just have va, how can we get io region first?

I am confused on this comment. If you already have va, it looks like you have already got the memory mapped. And thus, you will not need to call this function.

xiaoxiang781216:
> > 1.metal_io_region is a returned value here. If we just have va, how can we get io region first?
> 
> I am confused on this comment. If you already have va, it looks like you have already got the memory mapped. And thus, you will not need to call this function.

Yes, the platform layer has all memory mapping information, but OpenAMP may not.
Here is the case which is the root reason why this patch exist:
1.Tell OpenAMP the resource table location by remoteproc_set_rsc_table
2.remoteproc_set_rsc_table get the io region by remoteproc_get_io_with_va
But since remoteproc_mem list is empty, remoteproc_get_io_with_va fail here.
We didn't add remoteproc_mem item before, because:
1.Either mmap callback will add remoteproc_mem as needed
2.Or mmap callback save the mapping info locally and compute the returned value each time
In both case, remoteproc_get_io_with_va need fallback to mmap callback if remoteproc_mem list doesn't have the requested region.

edmooring:
This pull request will not go in. It changes an existing API and will break other people's code.
If this were proposed as a new function, with a commit message that explains the reason it has been added, and documentation for the function, I would consider it.

xiaoxiang781216:
@edmooring @arnopo there are at least two reasons why I think it's valuable to change remoteproc_mmap prototype:
1.It' design is asymmetric:
     a.Caller can convert pa to da/va pair.
     b.Caller can also convert da to pa/va pair
     c.But caller can't convert va to da/pa pair
  The third conversion is also very usefull in some case. For example, remoteproc_get_io_with_xxx can only query the registered region by remoteproc_get_mem. But with the new desgin, all of them call remoteproc_mmap which:
    a.Not only query the registered region like before
    b.But also fallback to mmap callback
  The second method give the platform a chance to do the mapping dynamically which is very important for the complex embeded system.
2.The new prototype extend to the old prototype, so it's very easly to adapter the old code to the new API as the PR show that only one or two line change from caller. For the platform side, if the developer don't want to support va to pa/da conversion, he/she can simply return error. BTW, since most people don't use remoteproc_mmap directly, I don't expect there are many compatiblity issue.
Last, I am tracking OpenAMP for more than four years, the breaking change happen almost in each offical release except the recent one(the incompatibility issue is much bigger than what I change here). So I supose that the goal of OpenAMP is better design and high quality(just like Linux), is it right?

xiaoxiang781216:
All checkpatch warning already exist in the original code base:
```
-------------------------------------------------------------
[1/1] Check commit - 99ff6c586a4783d91411a8497c3a873c72a00230
-------------------------------------------------------------
-:268: ERROR:CODE_INDENT: code indent should use tabs where possible
#268: FILE: apps/machine/zynqmp/zynqmp_linux_r5_proc.c:148:
+ metal_phys_addr_t *da, void **va, size_t size,$

-:268: WARNING:LEADING_SPACE: please, no spaces at the start of a line
#268: FILE: apps/machine/zynqmp/zynqmp_linux_r5_proc.c:148:
+ metal_phys_addr_t *da, void **va, size_t size,$

-:269: ERROR:CODE_INDENT: code indent should use tabs where possible
#269: FILE: apps/machine/zynqmp/zynqmp_linux_r5_proc.c:149:
+ unsigned int attribute, struct metal_io_region **io)$
```
it's better to fix by a new PR.

xiaoxiang781216:
> > @edmooring @arnopo there are at least two reasons why I think it's valuable to change remoteproc_mmap prototype:
> > agree with @edmooring this break the API

Ok, I can provide the compatable interface in next revision if the stable API is important. Let's focus on other issue first.

> > 1.It' design is asymmetric:
> > a.Caller can convert pa to da/va pair.
> > b.Caller can also convert da to pa/va pair
> > c.But caller can't convert va to da/pa pair
> 
> remoteproc_mmap as been designed to get a va based on a name, a da or a pa, not to convert
> va to pa. this is similar to the Linux mmap.
> So as proposed @wjliang better to add a new service than break the existing API.
> 
> > The third conversion is also very usefull in some case. For example, remoteproc_get_io_with_xxx can only query the registered region by remoteproc_get_mem. But with the new desgin, all of them call remoteproc_mmap which:
> > a.Not only query the registered region like before
> > b.But also fallback to mmap callback
> > The second method give the platform a chance to do the mapping dynamically which is very important for the complex embeded system.
> 
> Your need is not clear to me, is your need ids to make Va to pa or va to da conversion?
> In this case as proposed @wjliang the better solution is to add an API for this, even if the libmetal already offers you this API.
> 

Here is my case: we have the complex va<->pa<->da mapping, kernel already maintain the mapping table to support:
1.dma transfer(va->pa/da)
2.program the hardware by io remap(pa->va)
It doesn't make sense to duplicate the same complex table inside OpenAMP again by remoteproc_add_mem. So remoteproc::mems field is empty in our case and but remoteproc_ops::mmap point to a function which call OS API to query the mapping:
https://github.com/apache/incubator-nuttx/blob/801b9d6e5f8376e1aa2484756915218768083053/drivers/rptun/rptun.c#L223
Now, let's look at the remoteproc design especially remoteproc_ops::mmap, we can:
1.Easily implement the dynamical query  pa->va/da or da->va/pa
2.But impossible to implement the dynamical query va->pa/da
This is why I decide to extend mmap to cover the complex case.

> > 2.The new prototype extend to the old prototype, so it's very easly to adapter the old code to the new API as the PR show that only one or two line change from caller. For the platform side, if the developer don't want to support va to pa/da conversion, he/she can simply return error. BTW, since most people don't use remoteproc_mmap directly, I don't expect there are many compatiblity issue.
> 
> Here you point a principle of the open source community. Several or many users, with different platform/system/environment. one our objective is that user has not to update its code after each release. So you can understand that we can not adapt the API because it is more simple for you to use it in one way. We have also to take into account all other users.
> 

Ok, if other concern get resolved, I will update the patch:
1.Add a new callback function mmap2 to  remoteproc_ops
2.Add a new public function remoteproc_mmap2 which:
  a.Almost same as the current remoteproc_mmap, but
  b.Add the new argument void **va
  c.Try remoteproc_ops::mmap2 first
  d.Fallback to remoteproc_ops::mmap
3.remoteproc_mmap simply forward to remoteproc_mmap2

> > Last, I am tracking OpenAMP for more than four years, the breaking change happen almost in each offical release except the recent one(the incompatibility issue is much bigger than what I change here). So I supose that the goal of OpenAMP is better design and high quality(just like Linux), is it right?
> 
> Yes our objective is to make it stable and to ensure compatibility with the Linux implementation.
> The v2018.10 release was a special release that uncorrelates (as much as possible) the rpmsg and the remoteproc. Since this release we are applying same kind of politic than most of the open source project:
> 
> * avoid to break the legacy when possible.
> * extend the API instead of modifying it.
> * made rpmsg and resource table evolution in Linux and OpenAMP in parallel.

xiaoxiang781216:
@arnopo and @edmooring keep the compability now.

arnopo:
I'm trying to understand your need, but still not crystal clear.
what i understood here is that you have already mapped your memory in you system, and you want to provide the information to OpenAMP instead of registering remoteproc_mem.

The aim of the mmap function is to map a physical or a device address to a virtual address.
So memory map a virtual address look strange for me.
Regarding your need, look like that what is missing is a new fallback ops to provide the io region in remoteproc_get_io_with_xxx functions if no mems is found. 
something like 
struct metal_io_region *(get_io) (struct remoteproc *rproc,
		     metal_phys_addr_t *pa, metal_phys_addr_t *da,
		     void *va, size_t size);
Would this answer to your need?

xiaoxiang781216:
> I'm trying to understand your need, but still not crystal clear.
> what i understood here is that you have already mapped your memory in you system, and you want to provide the information to OpenAMP instead of registering remoteproc_mem.
> 

Yes, it's exactly the situation we want to resolve.

> The aim of the mmap function is to map a physical or a device address to a virtual address.
> So memory map a virtual address look strange for me.

Yes, mmap normally means to map the virtual memory to physic memory, but remoteproc_mmap is different because:
1.remoteproc_mmap don't have remoteproc_munmap counterpart
2.remoteproc_mmap call remoteproc_get_mem internally, so it's possible to get result without do any real mapping and caller don't know whether remoteproc_ops->mmap get called.

> Regarding your need, look like that what is missing is a new fallback ops to provide the io region in remoteproc_get_io_with_xxx functions if no mems is found.
> something like
> struct metal_io_region *(get_io) (struct remoteproc *rproc,
> metal_phys_addr_t *pa, metal_phys_addr_t *da,
> void *va, size_t size);
> Would this answer to your need?

Yes, it can resolve the problem too. So do you prefer this method?

arnopo:
Yes I would prefer this method, 
But perhaps let's @edmooring giving his feedback first, before going toward this solution

edmooring:
This looks like a good approach to me. @arnopo what is the effect on code size on a Cortex-M?

arnopo:
> This looks like a good approach to me. @arnopo what is the effect on code size on a Cortex-M?
This code has impact only a master processor in charge of loading a co-processor firmware. We can suppose that this processor has enough memory resource to be able to load a firmware.
I don't see impact on the co-processor foot print if the remoteproc is not used. Indeed the remoteproc is de-correlated from the rpmsg protocol.

bentheredonethat:
just curious if from reading this, would it be reasonable in either pull request or in another to introduce a remoteproc unmap API?

it would not necessarily conflict with pre-existing applications but only wire in if supplied by user and provide the virtual address unmap that is also normally done in the kernel-case

xiaoxiang781216:
@arnopo done.

arnopo:
> just curious if from reading this, would it be reasonable in either pull request or in another to introduce a remoteproc unmap API?
> 
> it would not necessarily conflict with pre-existing applications but only wire in if supplied by user and provide the virtual address unmap that is also normally done in the kernel-case

Hi @bentheredonethat 
sorry missing your comment. yes that makes sense. should be part of another PR

lib/remoteproc/remoteproc.c
```diff
@@ -32,7 +32,7 @@ remoteproc_check_fw_format(const void *img_data, size_t img_len)
 static struct remoteproc_mem *
 remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 		   metal_phys_addr_t pa, metal_phys_addr_t da,
-		   void *va, size_t size)
+		   void *va, size_t size, struct remoteproc_mem *buf)
```
arnopo, line 35: Could you explain why you need to add remoteproc_mem parameter as the structure is already returned by the function?
look to me that you need a copy of the remoteproc_mem structure.

why not having remote proc ops like this:
```
int (*get_mem)(struct remoteproc *rproc, const char *name,
		       metal_phys_addr_t pa, metal_phys_addr_t da,
		       void *va, size_t size, struct remoteproc_mem **mem);

```
and then you could call the ops like this
```
		ret = rproc->ops->get_mem(rproc, name, pa, da, va, size, &mem);
		if (ret == 0)
			return mem;

```

xiaoxiang781216: Because the porting layer normally has the different representation of memory region. If we use the second prototype, teh porting layer have to maintain another list which's type is remoteproc_mem and waste the memory.
It's also a common usage in the standard library design, for example:
https://en.cppreference.com/w/c/chrono/gmtime
```
struct tm *gmtime_r(const time_t *timep, struct tm *result);
```

arnopo: Please, could you provide link to associated porting layer and user code that i understand how you use this?

If you want to propose this kind of API i think you also need to implement this for get_mem ops to allow a porting layer relying on a list. With an appropriate description of the ops that describes the expected behavior.

xiaoxiang781216: > Please, could you provide link to associated porting layer and user code that i understand how you use this?
> 

Here(note I haven't convert the implementation from mmap to get_mem yet):
https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L210-L257
The mapping is done in these two function:
https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L647-L668
https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L670-L691
The table structure:
https://github.com/apache/incubator-nuttx/blob/master/include/nuttx/rptun/rptun.h#L261-L266

> If you want to propose this kind of API i think you also need to implement this for get_mem ops to allow a porting layer relying on a list. With an appropriate description of the ops that describes the expected behavior.

Done.

lib/include/openamp/remoteproc.h
```diff
@@ -410,6 +414,9 @@ struct remoteproc_ops {
 	int (*stop)(struct remoteproc *rproc);
 	int (*shutdown)(struct remoteproc *rproc);
 	int (*notify)(struct remoteproc *rproc, uint32_t id);
+	int (*get_mem)(struct remoteproc *rproc, const char *name,
+		       metal_phys_addr_t pa, metal_phys_addr_t da,
+		       void *va, size_t size, struct remoteproc_mem *mem);
```
arnopo, line 419: >> If you want to propose this kind of API i think you also need to implement this for get_mem ops to allow a porting layer relying on a list. With an appropriate description of the ops that describes the expected behavior.

>Done.
my point was also about get_mem itself
```
 -	int (*get_mem)(struct remoteproc *rproc, const char *name,
 +	struct remoteproc_mem *(*get_mem)(struct remoteproc *rproc, const char *name,
		       metal_phys_addr_t pa, metal_phys_addr_t da,
		       void *va, size_t size, struct remoteproc_mem *mem);

```
This would allow the use case where get_mem returns its own pointer.

lib/remoteproc/remoteproc.c
arnopo: Thanks for the links! that help to clarify my understanding of the use case!

arnopo (changes requested):

lib/include/openamp/remoteproc.h
xiaoxiang781216: Yes, the implementation could ignore mem argument and return the internal managed remoteproc_mem directly, but in this case it's more simpler to call remoteproc_add_mem in the initialize phase and reuse the common code instead.

xiaoxiang781216: Done.

xiaoxiang781216:
@arnopo could you merge this patch? then I can update the related stuff in NuttX.

lib/include/openamp/remoteproc.h
```diff
@@ -395,6 +395,13 @@ struct remoteproc {
  *        memory may not be off.
  * @shutdown: shutdown the remoteproc and release its resources.
  * @notify: notify the remote
+ * @get_mem: get remoteproc memory I/O region by either name, virtual address,
+ *           physical address or device address.
+ *           Note:
+ *           1.remoteproc_get_io_with_xxx try the internal list added by
+ *             remoteproc_add_mem first and then get_mem.
```
arnopo, line 402: Put this comment in  remoteproc_get_mem as this is related to remoteproc_get_mem  behavior not the op

lib/include/openamp/remoteproc.h
```diff
@@ -395,6 +395,13 @@ struct remoteproc {
...
+ *           2.get_mem could return either internal managed instance or
+ *             the 'buf' argument as needed.
```
arnopo, line 404: No clear that the buff is optional... proposal inspirated from https://en.cppreference.com/w/c/chrono/gmtime:
`if "buf" parameter is not specified the result is stored in static remoteproc_mem structure and a pointer to that static storage is returned. Else uses provided storage "buf" for the result and return the associated pointer.`

arnopo (changes requested):

lib/include/openamp/remoteproc.h
xiaoxiang781216: Done.

lib/include/openamp/remoteproc.h
xiaoxiang781216: buf isn't optional here. the caller always pass buf, but the implemenation could select wether to use this buffer or return an internal instance.

arnopo (approved):

lib/include/openamp/remoteproc.h
```diff
@@ -395,6 +395,10 @@ struct remoteproc {
  *        memory may not be off.
  * @shutdown: shutdown the remoteproc and release its resources.
  * @notify: notify the remote
+ * @get_mem: get remoteproc memory I/O region by either name, virtual address,
+ *           physical address or device address.
+ *           Note: could return either internal managed instance or the 'buf'
+ *           argument as needed.
```
edmooring, line 401: This is not enough documentation. 

It needs to specify return values, and error return values, as well as describing the values for each input parameter and what the default value should be if it is ignored. 

The circumstances under which the 'buf' argument could be set also need to be documented.

lib/remoteproc/remoteproc.c
```diff
@@ -29,10 +29,11 @@ remoteproc_check_fw_format(const void *img_data, size_t img_len)
 		return NULL;
 }
 
+/* try the internal list added by remoteproc_add_mem first and then get_mem callback */
 static struct remoteproc_mem *
 remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 		   metal_phys_addr_t pa, metal_phys_addr_t da,
-		   void *va, size_t size)
+		   void *va, size_t size, struct remoteproc_mem *buf)
```
edmooring, line 36: 'buf' is unused in this function, except in the (so far non-existent) case where the get_mem operation is provided. This may cause warnings from static analyzers. Maybe it would be better to always set it to match the return value.

edmooring (changes requested):

lib/include/openamp/remoteproc.h
xiaoxiang781216: > This is not enough documentation.
> 
> It needs to specify return values, and error return values, as well as describing the values for each input parameter and what the default value should be if it is ignored.
> 

But all other functions don't follow your suggestioin. It's better to create a new patch to fix all place instead.

> The circumstances under which the 'buf' argument could be set also need to be documented.

'buf' always pass to the callback, but the implementation could fill and return 'buf', or ignore it and return the internal managed instance.

lib/remoteproc/remoteproc.c
xiaoxiang781216: > 'buf' is unused in this function, except in the (so far non-existent) case where the get_mem operation is provided. This may cause warnings from static analyzers.

Could you point me the warning output? 

> Maybe it would be better to always set it to match the return value.

But it require we copy the whole struct from 'mem' to 'buf', which is unnecessary.

xiaoxiang781216: Anyway, I add the follow code like others:
```
(void)buf;
```
which should address your concern.

lib/include/openamp/remoteproc.h
xiaoxiang781216: @edmooring does my reply address your concern?

edmooring: > @edmooring does my reply address your concern?

It does not. 

Newly-added APIs should have good documentation from the start. If the documentation for older APIs should be improved, that can be addressed separately.

I have thought about it more, and the implementation should always fill in the 'buf' argument. That way an application developer knows they can always trust its value.

xiaoxiang781216: Done.

edmooring (approved):
Looks good to me now.

xiaoxiang781216:
@edmooring has approved PR, @arnopo could you merge it?

PR #173, opened 2019-05-21
compilation warnings and IAR c-stat reports

arnopo:
This series proposes corrections for few compilation issues and issues reported by IAR c-stat tools  (alignment with MISRA C 2012)
Only one c-stat report as not been fixed, as seems related to a false positive:
rsc_table_parser.c	52	Array pointer `rsc_table' is accessed with index [8,8] which may be out of array bounds [0,UNKNOWN]
rsc_table_parser.c	52	array pointer `rsc_table' is accessed with index [9,9] which may be out of array bounds [0,UNKNOWN] 
This seems not possible as 'reserved' 2 word array is defined in the resource table header.

### DIFF
--- a/lib/proxy/rpmsg_retarget.c
+++ b/lib/proxy/rpmsg_retarget.c
@@ -162,7 +162,7 @@ int _open(const char *filename, int flags, int mode)
 	struct rpmsg_rpc_syscall *syscall;
 	struct rpmsg_rpc_syscall resp;
 	int filename_len = strlen(filename) + 1;
-	int payload_size = sizeof(*syscall) + filename_len;
+	unsigned int payload_size = sizeof(*syscall) + filename_len;
 	unsigned char tmpbuf[MAX_BUF_LEN];
 	int ret;
 
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -37,10 +37,17 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 	struct metal_list *node;
 	struct remoteproc_mem *mem;
 
+	/*
+	 * Check name length to avoid overflow. This test has to be kept for
+	 * MISRA compliance
+	 */
+	if (name && strlen(name) > RPROC_MAX_NAME_LEN)
+		return NULL;
+
 	metal_list_for_each(&rproc->mems, node) {
 		mem = metal_container_of(node, struct remoteproc_mem, node);
 		if (name) {
-			if (!strncmp(name, mem->name, sizeof(mem->name)))
+			if (!strncmp(name, mem->name, strlen(name)))
 				return mem;
 		} else if (pa != METAL_BAD_PHYS) {
 			metal_phys_addr_t pa_start, pa_end;
@@ -170,7 +177,7 @@ struct remoteproc *remoteproc_init(struct remoteproc *rproc,
 
 int remoteproc_remove(struct remoteproc *rproc)
 {
-	int ret;
+	int ret = 0;
 
 	if (rproc) {
 		metal_mutex_acquire(&rproc->lock);
@@ -495,7 +502,6 @@ int remoteproc_load(struct remoteproc *rproc, const char *path,
 	metal_log(METAL_LOG_DEBUG, "%s: load executable data\r\n", __func__);
 	offset = 0;
 	len = 0;
-	ret = -RPROC_EINVAL;
 	while(1) {
 		unsigned char padding;
 		size_t nmemsize;
@@ -693,7 +699,6 @@ int remoteproc_load_noblock(struct remoteproc *rproc,
 		if (!loader) {
 			metal_log(METAL_LOG_ERROR,
 			       "load failure: failed to identify image.\r\n");
-			ret = -RPROC_EINVAL;
 			metal_mutex_release(&rproc->lock);
 			return -RPROC_EINVAL;
 		}
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -214,8 +214,6 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
 		/* TODO: clarify metal_cpu_yield usage*/
 		metal_cpu_yield();
 	}
-
-	return false;
 }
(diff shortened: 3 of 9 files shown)


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -310,9 +308,10 @@ static int rpmsg_virtio_send_offchannel_raw(struct rpmsg_device *rdev,
 		/* Lock the device to enable exclusive access to virtqueues */
 		metal_mutex_acquire(&rdev->lock);
 		avail_size = _rpmsg_virtio_get_buffer_size(rvdev);
-		if (size <= avail_size)
-			buffer = rpmsg_virtio_get_tx_buffer(rvdev, &buff_len,
-							    &idx);
+		if (size > avail_size)
+			return RPMSG_ERR_BUFF_SIZE;
```
wjliang: will need to call metal_mutex_release(&rdev->lock); before returning

lib/proxy/rpmsg_retarget.c
```diff
@@ -133,7 +133,6 @@ int rpmsg_rpc_send(struct rpmsg_rpc_data *rpc,
 		if (rpc->poll)
 			rpc->poll(rpc->poll_arg);
 	}
-	return ret;
```
wjliang: will you get compilation warning if you remove the return? it looks like you will need to have return as the return type is int.

lib/virtio/virtqueue.c
```diff
@@ -513,6 +517,7 @@ static void vq_ring_init(struct virtqueue *vq, void *ring_mem, int alignment)
 
 #ifndef VIRTIO_SLAVE_ONLY
 	if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
+		int i;
```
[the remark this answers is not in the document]
arnop2: ok i will add an empty line,

arnop2:
Any other comment on this pull request?
FYI at least error is generated in Zephyr CI test due to  lib/remoteproc/elf_loader.c warning should be fixed
(https://app.shippable.com/github/zephyrproject-rtos/zephyr/runs/44297/5/console)

lib/remoteproc/remoteproc.c
```diff
@@ -36,11 +36,16 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 {
 	struct metal_list *node;
 	struct remoteproc_mem *mem;
+	unsigned int str_size;
 
 	metal_list_for_each(&rproc->mems, node) {
 		mem = metal_container_of(node, struct remoteproc_mem, node);
 		if (name) {
-			if (!strncmp(name, mem->name, sizeof(mem->name)))
+			if (strlen(name) < sizeof(mem->name))
```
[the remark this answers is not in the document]
edmooring: @arnop2 strncmp() will stop the comparison as soon as it hits the null terminator in either string, or when the specified number of characters have been compared. Therefore it is safe to use in this situation. 
There is another bug here, namely that mem->name can be non-null terminated.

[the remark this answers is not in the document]
arnop2: @edmooring  Could you tell me if your comment is related to the the last version of the code using strcmp? In this case, please,  could you comment in code instead of replying of the conversation which is outdated? easier to follow :)
If name is not terminated by a null character and  sizeof (mem->name) > sizeof(name) you can access to the memory following the one reserved for name buffer.  This is the point highlighted by MISRA. Something strange (for my unsderstanding) is that MISRA accepts the use of strcmp instead...

lib/remoteproc/remoteproc.c
```diff
@@ -36,11 +36,12 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 {
 	struct metal_list *node;
 	struct remoteproc_mem *mem;
+	unsigned int str_size;
 
 	metal_list_for_each(&rproc->mems, node) {
 		mem = metal_container_of(node, struct remoteproc_mem, node);
 		if (name) {
-			if (!strncmp(name, mem->name, sizeof(mem->name)))
+			if (!strcmp(name, mem->name))
```
edmooring: This change is problematic, since it is currently possible that mem->name is not null-terminated. Linux remoteproc solves this by forcibly null-terminating the string, possibly truncating it. strncmp() is safer here, because it will terminate without running past the end of mem->name.

arnop2: The issue here seems not mem->name but "name" parameter. The name size is considered as unknown by MISRA. this generates a High level error:
  Overrun in strncmp with string `mem->name'. Length passed to strncmp is 32, but string is of length <unknown size>	LIB-strncmp-overrun	High	openamp\open-amp\lib\remoteproc\remoteproc.c:44
    remoteproc_get_mem	openamp\open-amp\lib\remoteproc\remoteproc.c
      if (name) is true	openamp\open-amp\lib\remoteproc\remoteproc.c:43
      if (!strncmp(name, mem->name, sizeof((mem->name))))	openamp\open-amp\lib\remoteproc\remoteproc.c:44
      LIB-strncmp-overrun	openamp\open-amp\lib\remoteproc\remoteproc.c:44
 
Following implementation matches with MISRA compliance. Could you tell me if this should be ok for you?
  if (name && strlen(name) <= sizeof(mem->name))  {
       if (!strncmp(name, mem->name, strlen(name)))
		return mem;

edmooring: I believe that the error reported here is erroneous:) strncmp() will compare at most sizeof(mem->name) bytes, stopping as soon as it sees a null terminator in either string.

That being said, I would have no objection to your alternative implementation, as long as you also add a comment explaining why the code is there. Otherwise, we risk someone removing it as unnecessary in the future.

arnop2: As CPU can access to a unexpected memory region, i understand the report. Now it is a read access could not be a High severity report... But anyway, i'm not 100% happy with my fix which is more a compromize than an elegant solution.  I also notice that there is no check on name parameter size... I 'm trying to find a alternative...

arnop2:
@edmooring 
Comment added for the strlen() workaround, as you requested

arnopo:
@edmooring, @wjliang 
Could you review this PR, at least the first 2 commits should be part of the next release.
Thanks!

edmooring:
Reviewed and approved.

PR #184, opened 2019-11-15
remoteproc: make all elf_* functions static except elf_identify

xiaoxiang781216:
it's always the best practice to reduce the naming scope as small as possible


arnopo:
Hi  @xiaoxiang781216,

I suppose that @wjliang exposes this API to applications for some purpose.
Do you have any reason to believe that this is not the case?
This could break legacy...

xiaoxiang781216:
> Hi @xiaoxiang781216,
> 
> I suppose that @wjliang exposes this API to applications for some purpose.
> Do you have any reason to believe that this is not the case?
> This could break legacy...

First, I search the whole tree doesn't find any caller outside elf_loader.c.
Second, the 3rd party still can call them through elf_ops->xxx.
Last, elf_load is very easy to conflict with RTOS which support loader elf from file system, actually we hit the double definition linker error now, and think other people will hit the similar error sooner or later because more and more RTOS support ELF.

xiaoxiang781216:
@arnopo do you have any more concern about this change? From my point of view, it's even better to make elf_ops and elf_identify as internal function and provide a generate function(e.g. get_loader). Then the exteranl user call get_loader to parse any format without any code change. Actually, we have a toolchain which output COFF format.
As you are the maintainer of openamp, you should take care what's the better desgin and imrpove it, instead of keeping the bad desgin spread the code base.

arnopo:
Same here this would break the legacy. 
We can image a user who implements it own remoteproc for muti reasons such as authenticated or encryted firmware, specific sections...
As we are not 100% sure that it is not use it seems more reasonable to keep it.

xiaoxiang781216:
> Same here this would break the legacy.
> We can image a user who implements it own remoteproc for muti reasons such as authenticated or encryted firmware, specific sections...
> As we are not 100% sure that it is not use it seems more reasonable to keep it.

So, is this mean that:
1.All functions prototype can't change anymore
2.All struct field can't remove anymore
3.All function can't change the behaviour
4.We can only add new function/struct or field
I am fine with the policy, but how can I get the same right like you to reject PR if anyone(even you) fail to follow the same princple? Because I already saw the breaking change happen many time before. I am afraid that you just ask the contributor to follow the strict rule, but your guy ignore the compatibility rule and make the change as you need.

arnopo:
> > Same here this would break the legacy.
> > We can image a user who implements it own remoteproc for muti reasons such as authenticated or encryted firmware, specific sections...
> > As we are not 100% sure that it is not use it seems more reasonable to keep it.
> 
> So, is this mean that:
> 1.All functions prototype can't change anymore
> 2.All struct field can't remove anymore
> 3.All function can't change the behaviour
> 4.We can only add new function/struct or field
These are not a strict rules, we have to be pragmatic. But since we don't know users implementation and if we want to propose a stable library, yes we have to follow as far as possible this rule.

I can also add 2 other rules based on some companies concerns:
- avoid to increase the library footprint (rom and ram) as it is  used on coprocessors with low memory
-  Limit latency introduced by the library.

> I am fine with the policy, but how can I get the same right like you to reject PR if anyone(even you) fail to follow the same princple? Because I already saw the breaking change happen many time before. I am afraid that you just ask the contributor to follow the strict rule, but your guy ignore the compatibility rule and make the change as you need.

It's a community, so you have the same right than me to challenge the PR. Ed and Myself are acting as maintainer but any reviewer are welcome, especially when the PR is sent by myself or Ed.

As a PR is always sent for an update in openamp or libmetal,  by subscribing to the mailing list you will see modifications.

xiaoxiang781216:
So, if I point out anyone's(or even your) patch break the rule. The patch shouldn't get merged before the patch fix the problem even you/Ed approve the patch or think the patch may benefit what you are tryng to resolve.

PR #187, opened 2020-01-09
Checkpatch script

arnopo:
Introduce checkpatch script, for openamp coding rules check.
These scripts have been cherry-picked from zephyr project
some additional patches fix the code according to the do_checkpatch result


wjliang:
thanks @arnopo to introduce checking tool. Will you have a following up patch to use it in the CI?
The checkpatch shell script is of Apache license, while the spelling.txt is of GPLv2 license. @galak or @edmooring can confirm if GPLv2 license is OK for the tools which will not be linked to the library.

arnopo:
> thanks @arnopo to introduce checking tool. Will you have a following up patch to use it in the CI?
> The checkpatch shell script is of Apache license, while the spelling.txt is of GPLv2 license. @galak or @edmooring can confirm if GPLv2 license is OK for the tools which will not be linked to the library.
FYI This scripts are a copy past of the zephyr project so reuse the licenses defined for zephyr.

PR #192, opened 2020-02-14
Pack the vring_desc and vring_avail structs

supergaute:
This is done in order to avoid a compiler warning from
the -Wcast-align flag.

### DIFF
--- a/lib/include/openamp/virtio_ring.h
+++ b/lib/include/openamp/virtio_ring.h
@@ -9,6 +9,8 @@
 #ifndef VIRTIO_RING_H
 #define	VIRTIO_RING_H
 
+#include <metal/compiler.h>
+
 #if defined __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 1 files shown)


lib/include/openamp/virtio_ring.h
```diff
@@ -43,13 +46,14 @@ struct vring_desc {
...
-};
+}METAL_PACKED_END;
 
+METAL_PACKED_BEGIN
 struct vring_avail {
 	uint16_t flags;
 	uint16_t idx;
 	uint16_t ring[0];
-};
+}METAL_PACKED_END;
```
arnopo: Today no warning on other structures but could you to extend patch for vring_used_elem and vring_used and vring  structure.
thanks

supergaute: Done

arnopo:
@supergaute 
just see that your sign-off is missing in commit message, please add it
Thanks,

PR #195, opened 2020-02-17
Adjustable rpmsg size

supergaute:
Add option to adjust RPMSG_BUFFER_SIZE during CMake build

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -74,5 +74,9 @@ endif (WITH_ZEPHYR)
 
 option (WITH_LIBMETAL_FIND "Check Libmetal library can be found" ON)
 
+if (DEFINED RPMSG_BUFFER_SIZE)
+  add_definitions( -DRPMSG_BUFFER_SIZE=${RPMSG_BUFFER_SIZE} )
+endif (DEFINED RPMSG_BUFFER_SIZE)
+
 message ("-- C_FLAGS : ${CMAKE_C_FLAGS}")
 # vim: expandtab:ts=2:sw=2:smartindent


supergaute:
Can you take a look at this? Having this change in the upstream repo would be very nice for our project.

xiaoxiang781216:
> Can you take a look at this? Having this change in the upstream repo would be very nice for our project.

The RPMSG_BUFFER_SIZE need align with kernel definiton, both side(OpenAMP and Linux kernel) need negotiate the new buffer size), you can see the patch here:
https://github.com/OpenAMP/open-amp/pull/155

arnopo:
> > Can you take a look at this? Having this change in the upstream repo would be very nice for our project.
> 
> The RPMSG_BUFFER_SIZE need align with kernel definiton, both side(OpenAMP and Linux kernel) need negotiate the new buffer size), you can see the patch here:
> #155

There are also discussions around resource table evolution to address the point.
Waiting conclusion this patch makes sense for MPOV,  for solution based on OpenAMP on both  processors.

cmake/options.cmake
```diff
@@ -74,5 +74,9 @@ endif (WITH_ZEPHYR)
 
 option (WITH_LIBMETAL_FIND "Check Libmetal library can be found" ON)
 
+if (DEFINED RPMSG_BUFFER_SIZE)
+  add_definitions( -DRPMSG_BUFFER_SIZE=${RPMSG_BUFFER_SIZE} )
```
edmooring: This needs a documentation update somewhere to tell the user that they can change the default, and that changing the default will make it incompatible with the Linux kernel implementation. I think that this change or #155 is major enough to bring up in the next openamp-rp call, or on the mailing list.

arnopo: Hi @edmooring,
Agree on the principle for the Documentation. We have to discuss the way to provide a documentation accessible to contributors.
Concerning this commit itself, it is not a major change. The RPMSG_BUFFER_SIZE can already be customized (https://github.com/OpenAMP/open-amp/blob/master/lib/include/openamp/rpmsg_virtio.h#L25)
So this patch only exports the definition, to be able to define the size in the make command line instead of updating the makefiles.

edmooring: Hi @arnopo 
For now, putting it in README.md, near the instructions for building OpenAMP for the various platforms would be fine with me.

arnopo: That makes sense, as some other cmake options have also to be defined (e.g WITH_VIRTIO_SLAVE, WITH_VIRTIO_MASTER ) , i would propose to have a dedicated thread for this.

arnopo:
Use and restriction of the RPMSG_BUFFER_SIZE option will de documented in 
https://github.com/OpenAMP/open-amp/pull/202

PR #196, opened 2020-02-19
examples: Remove stray local change from Linux platform_info.c

edmooring:
This change caused a compilation failure when compiling the OpenAMP examples on Linux due to a missing include file and some experimental changes to the libmetal shared memory API.

Signed-off-by: Ed Mooring <ed.mooring@linaro.org>


edmooring:
> Please add a commit message explained the rational of this patch ( e.g. fix application and test sample compilation issue)
> Else LGTM
> I also found a new issue by running non reg on your patch: #197

Is the new commit message acceptable?

Will check #197 today.

PR #201, opened 2020-04-08
Create CI checkpatch test using github workflow

arnopo:
Create a first checkpatch test for continuous integration
this job runs the checkpatch action from
https://github.com/webispy/checkpatch-action


.github/workflows/Continuous-integration.yml
```diff
@@ -0,0 +1,26 @@
+name: open-amp lib Continuous Integration
```
arnopo, line 1: filename to update to suppress the upper case

PR #204, opened 2020-04-23
Fix the wrong address usage

xiaoxiang781216:


xiaoxiang781216:
@arnopo @edmooring this patch series don't allocate the address between 0-1023 like kernel side. Also fix a potential bug that NS adress(0x35) will be allocated accidentally if user has more than 53 active channels.

apps/examples/echo/rpmsg-echo.c
```diff
@@ -58,7 +58,7 @@ int app(struct rpmsg_device *rdev, void *priv)
 	LPRINTF("Try to create rpmsg endpoint.\r\n");
 
 	ret = rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
-			       0, RPMSG_ADDR_ANY, rpmsg_endpoint_cb,
+			       1024, RPMSG_ADDR_ANY, rpmsg_endpoint_cb,
```
arnopo, line 61: if you have to update the example you break the legacy. 
Could you explain the rational? reserve memory from 0 to 1023 is  a design choice for Linux.  Today i don't see any reason to prohibit these addresses for OpenAMP, only 0x35 is reserved.
So to have a stable API i would prefer not integrate this commit if it is not an absolute necessity.

lib/include/openamp/rpmsg.h
```diff
@@ -146,8 +146,8 @@ int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
  */
 static inline unsigned int is_rpmsg_ept_ready(struct rpmsg_endpoint *ept)
 {
-	return (ept->dest_addr != RPMSG_ADDR_ANY) &&
-		(ept->addr != RPMSG_ADDR_ANY);
+	return ept && ept->rdev && ept->dest_addr &&
+		ept->dest_addr != RPMSG_ADDR_ANY;
```
arnopo, line 150: Same remark here, this break the legacy

lib/include/openamp/rpmsg.h
```diff
@@ -147,8 +168,17 @@ int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
...
+	int tc = 0;
+
+	for (; tc < RPMSG_TICK_COUNT; tc += RPMSG_TICKS_PER_INTERVAL) {
+		if (is_rpmsg_ept_ready(ept))
+			return rpmsg_send_offchannel_raw(ept, ept->addr,
+							 ept->dest_addr,
+							 data, len, true);
+		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
+	}
+
```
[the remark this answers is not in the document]
xiaoxiang781216: > This patch add an additional check on each send.
> Then what's happen if case of timeout in your application? you have another loop to try?

No, this loop will wait for 15 seconds, I don't think that more wait can help here.

> Could you explain the rational of this?

The patch is designed work with: https://github.com/OpenAMP/open-amp/pull/160
the caller can do:
```
rpmsg_create_ept
rpmsg_send
```
just like the socket API.

> i don't see here a real need except for NS_ACK, and only to try to simplify the application code.

It also work with your favor solution(the dummy server message). Sometime, it's very natual that the client need send the initial message to server just after the channel get created. Without this patch, you have to wait the sever send the first dummy back which make the code more complex than needed. And become impossible if the two phase handshake done inside the rpmsg except we provide the new callback to user.

> And you mention in previous commit: rpmsg_send_offchannel_raw already checks the destination address

Yes, but rpmsg_send_offchannel_raw return failure without the retry logic.

lib/include/openamp/rpmsg.h
```diff
@@ -271,25 +271,20 @@ static inline int rpmsg_trysend_offchannel(struct rpmsg_endpoint *ept,
...
-	ept->dest_addr = dest;
-	ept->cb = cb;
-	ept->ns_unbind_cb = ns_unbind_cb;
-}
+void rpmsg_init_ept(struct rpmsg_endpoint *ept,
+		    struct rpmsg_device *rdev,
+		    const char *name,
+		    uint32_t src, uint32_t dest,
+		    rpmsg_ept_cb cb,
+		    rpmsg_ns_unbind_cb ns_unbind_cb);
```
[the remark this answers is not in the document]
xiaoxiang781216: But I can't understand how the user can call the old API without problem: the user can't receive any callback without calling rpmsg_register_endpoint. It isn't only perfect but is wrong! rpmsg_init_ept is designed for the well known server, right? I suppose that all well known server can send/receive message immediately after calling this API.

apps/examples/echo/rpmsg-echo.c
xiaoxiang781216: > if you have to update the example you break the legacy.

Yes, because the example use the reserved address which encourage people use these special address without any thought.

> Could you explain the rational? reserve memory from 0 to 1023 is a design choice for Linux. Today i don't see any reason to prohibit these addresses for OpenAMP, only 0x35 is reserved.

From the original design, this region is reserved for the well known service. Both side(Linux and RTOS) need follow the convention: any private protocol shouldn't use the port inside this region to keep the compatibility. The reserved port may allocate to the well know service in the furture.

> So to have a stable API i would prefer not integrate this commit if it is not an absolute necessity.

But API implementation is wrong, we should fix the wrong behaviour as early as we can. The compability isn't a good reason to keep the bad behaviour and encourage the people do the wrong thing.

lib/include/openamp/rpmsg.h
xiaoxiang781216: If the demo code want to use 1-1023, the code should call rpmsg_init_ept instead. Here is a summary what I understand:
1.Well known service call rpmsg_init_ept with address inside 1-1023
2.Other service call rpmsg_create_ept with address >= 1024
3.It's prefer to call rpmsg_create_ept with RPMSG_ADDR_ANY(~0)
From my point, the quality of example code is very pool because it demo the bad practice:
1.Use the reserved address
2.Use the fixed address

xiaoxiang781216:
> Following commits fix some issue:
> [c310169](https://github.com/OpenAMP/open-amp/commit/c310169009dce571c3eaa8587f4355771485a631)
> [fbcf487](https://github.com/OpenAMP/open-amp/commit/fbcf487abebeaa75b43f9733dafa053c785e3311)
> => could you send them in a separate PR?
>

Here is the PR:
https://github.com/OpenAMP/open-amp/pull/205
 
> The other patches break the legacy and to be accepted we need justifications that the library can not work properly without them. In this case an approach that consists in extending the exiting API should be privileged.

lib/include/openamp/rpmsg.h
arnopo: Right use RPMSG_ADDR_ANY for the source address in demo is a best usage than fixing the endpoint. But we have to support both. So a good compromize would be to  update only a part of the demos. For instance the echo demo.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
xiaoxiang781216: The desgin is following the same principle inside rpmsg_virtio_send_offchannel_raw loop:
```
static int rpmsg_virtio_send_offchannel_raw(struct rpmsg_device *rdev,
					    uint32_t src, uint32_t dst,
					    const void *data,
					    int size, int wait)
{
	...
	if (wait)
		tick_count = RPMSG_TICK_COUNT / RPMSG_TICKS_PER_INTERVAL;
	else
		tick_count = 0;

	while (1) {
		...
		buffer = rpmsg_virtio_get_tx_buffer(rvdev, &buff_len, &idx);
		if (buffer || !tick_count)
			break;
		if (avail_size != 0)
			return RPMSG_ERR_BUFF_SIZE;
		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
		tick_count--;
	}
```
Does this mean that we shold drop the above loop too? From the design, rpmsg_send should block and wait the resource available, only rpmsg_try_send can return fail directly wihtout wait.

apps/examples/echo/rpmsg-echo.c
xiaoxiang781216: > > From the original design, this region is reserved for the well known service. Both side(Linux and RTOS) need follow the convention: any private protocol shouldn't use the port inside this region to keep the compatibility. The reserved port may allocate to the well know service in the furture. How can you handle this case in the compability way once it happen?
> 

What I mean the original desgin is the Linux Kernel implemenation. It's the golden reference implemntation and other implmenation should follow it's convention.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
xiaoxiang781216: > > But I can't understand how the user can call the old API without problem: the user can't receive any callback without calling rpmsg_register_endpoint.
> 
> Yes it is the normal behavior. you have to register callback for communication.
> 
> > It isn't only perfect but is wrong! rpmsg_init_ept is designed for the well known server, right? I suppose that all well known server can send/receive message immediately after calling this API, but it is impossible without this patch. even worse the user can't call rpmsg_register_endpoint by self because it is declared in a private header file.
> 
> No, rpmsg_init_ept is designed to initialize the ept structure before registering it. But i agree that something is

rpmsg_create_ept initilize all rpmsg_endpoint's fields, why need we call rpmsg_init_ept before that? Does it make sense to initialize the same fields twice?

> not good here.
> The normal API for end point management is rpmsg_create_ept. I can not see a reason why rpmsg_init_ept is exposed. For me this function should be declared in rpmsg_internal.h.
> Any reason that you need to use it, instead of rpmsg_create_ept?

From my interception it's better to:
1.call rpmsg_init_ept for the well known service(both src/dest is fixed) and skip all NS interaction
2.call rpmsg_create_ept for the dynamic serice(either src/dest is any) and utilize NS to find the peer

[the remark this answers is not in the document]
arnopo: > From my interception it's better to:
> 1.call rpmsg_init_ept for the well known service(both src/dest is fixed) and skip all NS interaction
> 2.call rpmsg_create_ept for the dynamic serice(either src/dest is any) and utilize NS to find the peer

In both case rpmsg_create_ept have to be called, this is what has been implemented since the 2018.10 release. This function handles fixed and non fixed addresses, NS is sent only if destination address is set to RPMSG_ADDR_ANY.
That's why  rpmsg_register_endpoint is not exposed as API  and why rpmsg_init_ept should probably not be exposed.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
arnopo: > The desgin is following the same principle inside rpmsg_virtio_send_offchannel_raw loop:
> 
> ```
> static int rpmsg_virtio_send_offchannel_raw(struct rpmsg_device *rdev,
> 					    uint32_t src, uint32_t dst,
> 					    const void *data,
> 					    int size, int wait)
> {
> 	...
> 	if (wait)
> 		tick_count = RPMSG_TICK_COUNT / RPMSG_TICKS_PER_INTERVAL;
> 	else
> 		tick_count = 0;
> 
> 	while (1) {
> 		...
> 		buffer = rpmsg_virtio_get_tx_buffer(rvdev, &buff_len, &idx);
> 		metal_mutex_release(&rdev->lock);
> 		if (buffer || !tick_count)
> 			break;
> 		if (avail_size != 0)
> 			return RPMSG_ERR_BUFF_SIZE;
> 		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
> 		tick_count--;
> 	}
> ```
> 
> Does this mean that we shold drop the above loop too? From the design, rpmsg_send should block and wait the resource available, only rpmsg_try_send can return fail directly wihtout wait.

No, the loop has to be kept here, blocking fro rpmsg_send, not blocking for rpmsg_try_send ( same behavior than Linux API).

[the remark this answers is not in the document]
xiaoxiang781216: If so, why you consider my patch useless? It also handle the case where the resouce unavailable.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
xiaoxiang781216: Anyway I just want to make OpenAMP API design become better. If you refuse my change for rpmsg_init_ept I am fine, I will drop in the next update, but for other bug fix

lib/include/openamp/rpmsg.h
xiaoxiang781216: I will provide a patch change ALL 0 to RPMSG_ADDR_ANY.

apps/examples/echo/rpmsg-echo.c
arnopo: This is you opinion...
Another paradigm could be that we respect the Linux convention as Linux does not complain about the remote OpenAMP address...

Anyway I think the good place to continue this discussion is the  OpenAMP-RP meeting. 
It is an interesting topic to discuss to have more opinions on this topic.
@edmooring, any opinion on this?

lib/include/openamp/rpmsg.h
xiaoxiang781216: Here is the PR:
https://github.com/OpenAMP/open-amp/pull/207

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
arnopo: Simply  because you have the API to do this in application, so i don't see a strong reason to integrate it in rpmsg_send.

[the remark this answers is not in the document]
xiaoxiang781216: If so, we just need provide rpmsg_try_send I don't see any reason to add rpmsg_send which is very easy to implement on top of rpmsg_try_send. As I said before, it's a common issue all people will hit immediately after he/she create a rpmsg endpoint proactively. The library need provide a solution to help people mitigate the problem, instead let user find the touch bug and repeat the code again and again.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
arnopo: I don't think that this point is close. I'm agree with you there is something wrong with rpmsg_init_ept. This is for me a case where we have probably to fix the API.
Firsdt of all i would understand why you propose to duplicate the endpoint API by updating rpmsg_init_ept instead of using rpmsg_create_ept? Any usecase that you can not support using rpmsg_create_ept?

apps/examples/echo/rpmsg-echo.c
xiaoxiang781216: One day, if Linux kernel decide to assgin the additional number in this region to some new well known service, your program will stop working suddenly. the code can work today, doesn't mean that it's implmentation is right.
Actually, the bug is already in the code base now, if you create the channel more than 0x35, you wil hit the brick. In the furture, you will hit more brick:)

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
xiaoxiang781216: I already said before, because rpmsg_create_ept will send NS message, but NS message isn't necessary and may have the bad side effect on kernel side for the well known service because both side already know the peer address. Please imagine how the kernle will response if you create ”NS" endpoint by rpmsg_create_ept.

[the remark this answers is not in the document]
xiaoxiang781216: If you think more carefully and spend more time to read OpenAMP code, you can understand the difference between rpmsg_init_ept and rpmsg_create_ept. Anyway, I don't want to spend my time to improve OpenAMP anymore. I just want to fix the bug, so I will always drop this patch in the next update.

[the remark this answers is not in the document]
xiaoxiang781216: "NS" channel is created by rpmsg_init_ept, so it's natural to use it for all well known service. Anyway, rpmsg_init_ept in the current form is broken at least as a public API.

xiaoxiang781216:
@arnopo could merge the unquestioned PR first:
https://github.com/OpenAMP/open-amp/pull/207
https://github.com/OpenAMP/open-amp/pull/206
https://github.com/OpenAMP/open-amp/pull/205
then I can update this PR.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
arnopo: > If so, we just need provide rpmsg_try_send I don't see any reason to add rpmsg_send which is very easy to implement on top of rpmsg_try_send.

rpmsg_try_send returns an error if no TX buffer available, rpmsg_send is blocked waiting a TX_buffer. No relation chip with your usecase

> As I said before, it's a common issue all people will hit immediately after he/she create a rpmsg endpoint proactively. The library need provide a solution to help people mitigate the problem, instead let user find the touch bug and repeat the code again and again.

As rpmsg_send returns an error if remote address is not defined it is more a bug in application that does not treat error and /or does not use is_rpmsg_ept_ready as show in examples.
So the issue here seems more the use of the API than the API itself. 
If user code is inspirated from the examples they should use is_rpmsg_ept_ready. 
But If you have implemented the NS ack for your customer you probably also increases this bad usage, by masking the rpmsg current protocol.

the solution to help user seems more the documentation or example.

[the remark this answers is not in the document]
xiaoxiang781216: If you search examples folder, all client side has this code:
```
	/* Create RPMsg endpoint */
	ret = rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
			       RPMSG_ADDR_ANY, RPMSG_ADDR_ANY,
			       rpmsg_endpoint_cb, rpmsg_service_unbind);
	if (ret) {
		LPERROR("Failed to create RPMsg endpoint.\r\n");
		return ret;
	}

	while (!is_rpmsg_ept_ready(&lept))
		platform_poll(priv);
```
As OpenAMP maintainer, do you have any responsbility to improve this situation in the next release? I don't think the code like this can survive in Linux kernel. Actually from my point, this patch is just the first step, the next is:
1.Ensure all system under libmetal/lib/lib implement conditon
2.Replace the poll loop with the event trigger just like Linux kernel
But anyway, I don't make any suggestion like this anymore since I don't want to waste my time to dissuss how to improve OpenAMP.

arnopo:
> @arnopo could merge the unquestioned PR first:
> #207
> #206
> #205
> then I can update this PR.

This patch will be merged  when both myself and @edmooring  have approved them

Then this PR is coming unreadable as there are several feature in one PR, so many parallel discussion.. Would be nice to have one PR per feature.
And the most  important, we have all to stay constructive to find solutions.

Here is a sum-up of my current vision on the 3 mains points discussed:

- commit: rpmsg: shouldn't allocate 0-1023 address in rpmsg_create_ept 
I'd rather not prevent needs that could never happen.
 - this can never happen.
 - this will probably generate more impact such as a the support of a new service
 - waiting this we could generate regression on exiting code.
So it seems to me, more reasonable to postpone and address this point when a real case  will come.
Anyway  as i proposed let's discuss the point in bi weekly meeting to have more opinion on this.

- commit: rpmsg: wait ept ready in rpmsg_send:
 I understand your point of view and your request  to simplify the user code. But I have 2 concerns about your implementation:

  - Is this need is not linked to the NS ack feature? In current OpenAMP implementation as a first message is needed from the remote ept , could we consider that  the start of the send can be conditioned by the first message received ( ept callback trigger)?
  - Some other companies complain about latency introduced in rpmsg transfer. So I have also to take this into account, and try to not increase message processing as far as possible.
So what i request here is to understand if  there is a use case in current implementation,  that justify it, or if this commit should be discussed with the NS ack. Then to avoid to impact other plaforms, a compromise could be to add a new API.

- commit rpmsg: merge rpmsg_register_endpoint into rpmsg_init_ept
I think here there is a misunderstanding between you and me. To solve this, i sent you a personal mail that we review code together that we align our comprehension of the code  and find a solution.

xiaoxiang781216:
> > @arnopo could merge the unquestioned PR first:
> > #207
> > #206
> > #205
> > then I can update this PR.
> 
> This patch will be merged when both myself and @edmooring have approved them
> 
> Then this PR is coming unreadable as there are several feature in one PR, so many parallel discussion.. Would be nice to have one PR per feature.

Ok, once #207, #206 and #205 merge, I will split the PR into serveral one.

> And the most important, we have all to stay constructive to find solutions.
> 
> Here is a sum-up of my current vision on the 3 mains points discussed:
> 
> * commit: rpmsg: shouldn't allocate 0-1023 address in rpmsg_create_ept
>   I'd rather not prevent needs that could never happen.
> * this can never happen.
> * this will probably generate more impact such as a the support of a new service
> * waiting this we could generate regression on exiting code.
>   So it seems to me, more reasonable to postpone and address this point when a real case  will come.

Since rpmsg is the protocol between the different OS, it is no reason to break the protocol without any justification. Do you think htat it's good decision OpenAMP will stop working just because the rpmsg community define new standard service or some vendor want to define the private well known service?

>   Anyway  as i proposed let's discuss the point in bi weekly meeting to have more opinion on this.

Ok, it's better to vote on the meeting.

> * commit: rpmsg: wait ept ready in rpmsg_send:
>   I understand your point of view and your request  to simplify the user code. But I have 2 concerns about your implementation:
>   
>   * Is this need is not linked to the NS ack feature?

The loop already exist in all example now, but NS ack feature isn't mereged yet, so why you think that this patch link to that feature? This patch can move the loop to common place and allow the furture optimation more easlier.

> In current OpenAMP implementation as a first message is needed from the remote ept , could we consider that  the start of the send can be conditioned by the first message received ( ept callback trigger)?

Yes, this is a solution. But could you think about why all examples select the busy loop method? If you try to convert an example which require the client to send the message immediately after the endpoint get created, you can understand 

>   * Some other companies complain about latency introduced in rpmsg transfer. So I have also to take this into account, and try to not increase message processing as far as possible.

Hi, this patch is a step to address this issue. If you let's caller add the loop by them self, how you can resolve the problem in the furture? Sleeping inside the loop contribute the latency, the best method is to wait/signal condition.

>     So what i request here is to understand if  there is a use case in current implementation,  that justify it, or if this commit should be discussed with the NS ack. Then to avoid to impact other plaforms, a compromise could be to add a new API.
> * commit rpmsg: merge rpmsg_register_endpoint into rpmsg_init_ept
>   I think here there is a misunderstanding between you and me. To solve this, i sent you a personal mail that we review code together that we align our comprehension of the code  and find a solution.

xiaoxiang781216:
@arnopo and @edmooring I create the new PR for the controversial patch:
https://github.com/OpenAMP/open-amp/pull/208
https://github.com/OpenAMP/open-amp/pull/209
So the patch left in this PR can be merged first.

lib/rpmsg/rpmsg.c
```diff
@@ -226,15 +228,16 @@ int rpmsg_create_ept(struct rpmsg_endpoint *ept, struct rpmsg_device *rdev,
...
 	rpmsg_init_ept(ept, name, addr, dest, cb, unbind_cb);
 	rpmsg_register_endpoint(rdev, ept);
+	metal_mutex_release(&rdev->lock);
 
-	if (rdev->support_ns && ept->dest_addr == RPMSG_ADDR_ANY) {
-		/* Send NS announcement to remote processor */
-		metal_mutex_release(&rdev->lock);
+	/* Send NS announcement to remote processor */
+	if (ept->name[0] && rdev->support_ns &&
+	    ept->dest_addr == RPMSG_ADDR_ANY)
```
arnopo, line 235: this is not related to the commit message, please send this fix in a separate commit.

lib/include/openamp/rpmsg.h
```diff
@@ -110,6 +110,19 @@ struct rpmsg_device {
...
+ * @ept: pointer to rpmsg endpoint
+ *
+ * Returns 1 if the rpmsg endpoint has both local addr and destination
+ * addr set, 0 otherwise
+ */
+static inline unsigned int is_rpmsg_ept_ready(struct rpmsg_endpoint *ept)
+{
+	return ept && ept->rdev && ept->dest_addr != RPMSG_ADDR_ANY;
+}
+
```
arnopo, line 125: Please could you not move the position of the declaration in file, that a diff show that you only update the return test here.

lib/rpmsg/rpmsg.c
```diff
@@ -132,7 +132,7 @@ int rpmsg_send_ns_message(struct rpmsg_endpoint *ept, unsigned long flags)
 	ns_msg.flags = flags;
 	ns_msg.addr = ept->addr;
 	strncpy(ns_msg.name, ept->name, sizeof(ns_msg.name));
-	ret = rpmsg_send_offchannel_raw(ept, ept->addr,
+	ret = rpmsg_send_offchannel_raw(ept, RPMSG_NS_EPT_ADDR,
```
arnopo, line 135: Any rational to do this?
Looking at Linux implementation the source address seems the endpoint address not the NS endpoint address

lib/rpmsg/rpmsg.c
xiaoxiang781216: Ok

lib/include/openamp/rpmsg.h
xiaoxiang781216: Ok, I will split the movement to another patch

lib/rpmsg/rpmsg.c
xiaoxiang781216: Yes, you are right. I will drop this patch.

lib/include/openamp/rpmsg.h
arnopo: For this series , it is sufficient to put the function back to its original place that the diff shows that only the condition as been reworked.
Then LGTM for the rest of the series

xiaoxiang781216: But, you can view the individual patches in commits tab or git log. I want to keep the patchset as small as possible:
https://github.com/apache/incubator-nuttx/tree/master/openamp

lib/rpmsg/rpmsg.c
```diff
@@ -176,18 +176,14 @@ struct rpmsg_endpoint *rpmsg_get_endpoint(struct rpmsg_device *rdev,
 
 static void rpmsg_unregister_endpoint(struct rpmsg_endpoint *ept)
 {
-	struct rpmsg_device *rdev;
-
-	if (!ept)
-		return;
-
-	rdev = ept->rdev;
+	struct rpmsg_device *rdev = ept->rdev;
```
arnopo, line 179: not related to the commit
Please put this in https://github.com/OpenAMP/open-amp/pull/204/commits/fe6412ba6a0cd6237b7d0ae16e6616d9fbbc5c1a and rename the commit in something like rpmsg_unregister_endpoint management cleanup

lib/include/openamp/rpmsg.h
xiaoxiang781216: Since you are arguing to add poll inside rpmsg_send, I prefer to merge this simple patch first, so the downstream project can redue the patch need to apply even the poll patch can't upstream.

lib/rpmsg/rpmsg.c
xiaoxiang781216: Done.

lib/include/openamp/rpmsg.h
arnopo: Doing this we will simplify your downstream but perhaps impact another one... Not a good practice, if we accept this one, then we would have to accept future patches from another platform that could not match with your expectation.
And  regarding https://github.com/apache/incubator-nuttx/blob/master/openamp/0004-rpmsg-wait-ept-ready-in-rpmsg_send.patch seems to me that it is minor in term of impact in your downstream.

xiaoxiang781216: But, can you please tell me what will impact by this patch? And it is also good thing to put the simple inline function at the begining like other header files:
https://github.com/OpenAMP/open-amp/blob/master/lib/include/openamp/remoteproc.h#L432-L448
https://github.com/OpenAMP/open-amp/blob/master/lib/include/openamp/rpmsg_virtio.h#L65-L110

arnopo: As explained in my previous comment, this commit seems appears to be for  your convenience only.  Do I accept it to simplify your patches on top of OpenAMP and potentially impact some other platforms which would have to rework their own patch on top of OpenAMP because this one introduce a conflict? I can not find here a good reason to do this as it does not  answer to a real need.

PR #205, opened 2020-05-05
Minor fix

xiaoxiang781216:


xiaoxiang781216:
@arnopo could we merge the patch which doesn't have any concern?

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -128,20 +128,43 @@ static uint32_t rproc_virtio_negotiate_features(struct virtio_device *vdev,
...
-	(void)vdev;
-	(void)offset;
-	(void)src;
-	(void)length;
+	struct remoteproc_virtio *rpvdev;
+	struct fw_rsc_vdev *vdev_rsc;
+	struct metal_io_region *io;
+	char *config;
+
+	if (offset + length > vdev->config_len || offset + length < length)
```
arnopo, line 157: My compiler complains here and in rproc_virtio_read_config :
warning: comparison between signed and unsigned integer expressions
=> length is signed while others params are unsigned.

Seems that there are some miss alignment of the length declaration between rpmsg, virtio and virtio queue.
It seems reasonable to me to update virtio_dispatch struct to define length as unsigned.

This makes me think that updating rpmsg to declare length as unsigned would be cleaner and no generate a regression  for legacy code. But this is another point...

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -222,6 +245,7 @@ rproc_virtio_create_vdev(unsigned int role, unsigned int notifyid,
 	vdev->notifyid = notifyid;
 	vdev->role = role;
 	vdev->reset_cb = rst_cb;
+	vdev->config_len = vdev_rsc->config_len;
```
arnopo, line 248: seems no need of  vdev->config_len field as information is in vdev_rsc that is accessible in rproc_virtio_write_config and rproc_virtio_read_config.

lib/remoteproc/remoteproc_virtio.c
arnopo, line 157: offset + length < length ?
i don't understand this test... seems similar to  length < 0?

xiaoxiang781216:
> The last one is not a fix but a change request.
> It is simpler for the review to have one PR per feature. If someone is not agree with the config implementation this will block the 2 fixes than LGTM.

No, it is a bug fix, do you think is the empy implementation of rproc_virtio_read_config and rproc_virtio_write_config right? But, anyway, I will send a new PR. I don't want to spend the time argue this.

lib/remoteproc/remoteproc_virtio.c
xiaoxiang781216: > My compiler complains here and in rproc_virtio_read_config :
> warning: comparison between signed and unsigned integer expressions
> => length is signed while others params are unsigned.
> 

Ok, I will read the compiler output, but could you add -Werror to CI? then contributor can know something happen quickly.

> Seems that there are some miss alignment of the length declaration between rpmsg, virtio and virtio queue.
> It seems reasonable to me to update virtio_dispatch struct to define length as unsigned.
> 
> This makes me think that updating rpmsg to declare length as unsigned would be cleaner and no generate a regression for legacy code. But this is another point...

xiaoxiang781216: it can catch the overflow error. Kernel has the same code too.

lib/remoteproc/remoteproc_virtio.c
xiaoxiang781216: Yes, fix.

xiaoxiang781216:
Here is the new PR for config read/write:
https://github.com/OpenAMP/open-amp/pull/206/files

lib/remoteproc/remoteproc_virtio.c
xiaoxiang781216: Yes, removed from new PR

lib/remoteproc/remoteproc_virtio.c
xiaoxiang781216: Anyway, I removed to avoid the warning, we can trust caller in embeded system anyway.

PR #207, opened 2020-05-11
apps: replace all zero address to RPMSG_ADDR_ANY

xiaoxiang781216:
the example should demo the best practice for the endpoint creation

### DIFF
--- a/apps/tests/msg/rpmsg-flood-ping.c
+++ b/apps/tests/msg/rpmsg-flood-ping.c
@@ -12,7 +12,7 @@
 #include "platform_info.h"
 #include "rpmsg-ping.h"
 
-#define APP_EPT_ADDR    0
+#define APP_EPT_ADDR    1024
 #define LPRINTF(format, ...) printf(format, ##__VA_ARGS__)
 #define LPERROR(format, ...) LPRINTF("ERROR: " format, ##__VA_ARGS__)
 
--- a/apps/examples/echo/rpmsg-echo.c
+++ b/apps/examples/echo/rpmsg-echo.c
@@ -58,7 +58,8 @@ int app(struct rpmsg_device *rdev, void *priv)
 	LPRINTF("Try to create rpmsg endpoint.\r\n");
 
 	ret = rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
-			       0, RPMSG_ADDR_ANY, rpmsg_endpoint_cb,
+			       RPMSG_ADDR_ANY, RPMSG_ADDR_ANY,
+			       rpmsg_endpoint_cb,
 			       rpmsg_service_unbind);
 	if (ret) {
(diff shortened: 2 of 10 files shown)


apps/tests/msg/rpmsg-flood-ping.c
```diff
@@ -92,7 +91,7 @@ static void rpmsg_name_service_bind_cb(struct rpmsg_device *rdev,
 		LPERROR("Unexpected name service %s.\r\n", name);
 	else
 		(void)rpmsg_create_ept(&lept, rdev, RPMSG_SERVICE_NAME,
-				       APP_EPT_ADDR, dest,
+				       RPMSG_ADDR_ANY, dest,
```
arnopo, line 94: Could you keep this test unchanged? the idea here is to keep one test with a fixed source address to also test a use case with a fixed source address.

xiaoxiang781216: Ok

xiaoxiang781216:
@arnopo please merge this PR if you don't have more concern.

apps/tests/msg/rpmsg-flood-ping.c
```diff
@@ -12,7 +12,7 @@
 #include "platform_info.h"
 #include "rpmsg-ping.h"
 
-#define APP_EPT_ADDR    0
+#define APP_EPT_ADDR    1024
```
arnopo, line 15: I applied your patch and test it.
This address is not valid.  Current addressing is based on a 128 bit field, so only address 0 to 127 are supported.
as consequence test fails on endpoint creation.
Please use an address in 0-127 range.

Here is a bash script to test it:

dirs_to_check=$PWD
echo "################### build libmetal #####################"
cd libmetal
rm -r build
cmake . -Bbuild || exit 1
cd build
make || exit 1

cd $dirs_to_check

echo "################### build openamp #####################"
cd open-amp
rm -r build
cmake . -DWITH_APPS=on -DWITH_PROXY=on -Bbuild -DCMAKE_INCLUDE_PATH="../libmetal/build/lib/include" -DCMAKE_LIBRARY_PATH="../libmetal/build/lib" || exit 1
cd build
make install DESTDIR=./bin || exit 1

cd $dirs_to_check
cd open-amp
cd build

echo "################### run flood test #####################"
LD_LIBRARY_PATH=./bin/usr/local/lib:../../libmetal/build/lib/  ./bin/usr/local/bin/rpmsg-echo-shared &
sleep 1
LD_LIBRARY_PATH=./bin/usr/local/lib:../../libmetal/build/lib/  ./bin/usr/local/bin/msg-test-rpmsg-flood-ping-shared 1

xiaoxiang781216: Yes, this change need PR:
https://github.com/OpenAMP/open-amp/pull/209
I think the new implementation for rpmsg_create_ept can make both side happy.

PR #208, opened 2020-05-15
rpmsg: move rpmsg_init_ept to lib/rpmsg/rpmsg_internal.h

xiaoxiang781216:
and merge rpmsg_register_endpoint into rpmsg_init_ept so the
service which use preserved address could call rpmsg_init_ept

### DIFF
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -266,33 +266,6 @@ static inline int rpmsg_trysend_offchannel(struct rpmsg_endpoint *ept,
 }
 
 /**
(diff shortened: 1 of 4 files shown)


xiaoxiang781216:
rpmsg_init_ept is used for well known service, so it's better to put into the internal header file.

arnopo:
Thanks for the patch, but we have first to define the process to make an API deprecated with ED and the OpenAMP members. We will apply the policy on this API.

arnopo:
fixed by https://github.com/OpenAMP/open-amp/pull/221

PR #213, opened 2020-06-19
cmake: set the policy CMP0077 to NEW

tejlmand:
emoves the following warning when using newer CMake versions:
```
Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
--help-policy CMP0077" for policy details.  Use the cmake_policy command
to set the policy and suppress this warning.`
```

Both OLD and NEW behavior will use the value defined earlier in the
CMake process, however the OLD behavior will also add an entry to the
CMakeCache.txt but ignore that value on sub-sequent invocations.
As this may lead to confusion on users looking into the CMakeCache.txt,
then the best is to not get such values into the cache, and thus use NEW
behavior.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,10 @@ if (POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif()
 
+if (POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
 # The version number
 set (OPENAMP_VERSION_MAJOR 1)
 set (OPENAMP_VERSION_MINOR 0)


arnopo:
Hi @tejlmand
Look good to me. I tested with cmake 3.5 and 3.17.
i also tried to test compatibility with cmake 2.6 but it fails for some other reasons in Libmetal (e.g not able to find librt on my Ubuntu 18.04).
Do you test with cmake 2.6 in your environment?
i would take to opportunity to verify that is the minimum supported version of cmake is still v2.6  
@edmooring : seems that the support of the v2.6 as minimum version comes from Xilinx.
Could you check also on your side?

Thanks,
Arnaud

tejlmand:
There were a question from @arnopo in https://github.com/OpenAMP/libmetal/pull/119 which I commented here:
https://github.com/OpenAMP/libmetal/pull/119#issuecomment-648614285

basically, this issue is seen because Zephyr sets:
https://github.com/zephyrproject-rtos/open-amp/blob/master/CMakeLists.txt#L9-L25
before including OpenAMP.

As Zephyr sets those variable first, then it basically means that CMP0077 most likely is preferred to be set to NEW.
However CMP0077 is not available to CMake <3.12 users.
Note: older CMake users might actually not realize that OLD behavior is already hiding their configuration.

tejlmand:
DNM added in title to allow for some time to discuss.

arnopo:
I am not a cmake expert so difficult to me to see real impact if this policy is set to NEW. 

For cmake version <= 3.12 in theory no impact, so the risk is for cmake  >3.12

I don't see any regression using cmake 3.17 for examples compiled on Linux platform with policy set to NEW.
i suppose that it is also OK on zephyr...
But what about other environments...

By default  it seems reasonable to set it to OLD as you propose, but on the other side the OLD behavior can be set to deprecated in future versions  of cmake. So seems more  reliable to set it to NEW.

@edmooring what is you feeling on this? 

Could also nice to have more feedback from other users.

tejlmand:
@arnopo I finally found the time to investigate this a bit more.
Both OLD and NEW behavior will behave the same, in the sense that the both will honor the current value of the option, if specified earlier in the build system, such as here: https://github.com/zephyrproject-rtos/open-amp/blob/master/CMakeLists.txt#L9

The difference between OLD and NEW is that OLD will add an entry in the CMakeCache.txt, which is set to either ON / OFF, irregardless of the previous value, so in fact the value from the CMakeCache.txt is effectively ignored.

Using the CMP0077 NEW ensures the variable will never enter the CMakeCache.txt and thus remove any risk of confusing users who examins the CMakeCache.txt.

tejlmand:
This also mean that there is no strong arguments in this PR for requiring a later CMake version.

arnopo:
> @arnopo I finally found the time to investigate this a bit more.
> Both OLD and NEW behavior will behave the same, in the sense that the both will honor the current value of the option, if specified earlier in the build system, such as here: https://github.com/zephyrproject-rtos/open-amp/blob/master/CMakeLists.txt#L9
> 
> The difference between OLD and NEW is that OLD will add an entry in the CMakeCache.txt, which is set to either ON / OFF, irregardless of the previous value, so in fact the value from the CMakeCache.txt is effectively ignored.
> 
> Using the CMP0077 NEW ensures the variable will never enter the CMakeCache.txt and thus remove any risk of confusing users who examins the CMakeCache.txt.
> 
> Ready for merge.

Thanks for the investigation.  LGTM to set it  to NEW

arnopo:
> This also mean that there is no strong arguments in this PR for requiring a later CMake version.

yes not relationship with this PR. In  reviewing you commit i just wondered if we we still support cmake from version 2.6

tejlmand:
> > This also mean that there is no strong arguments in this PR for requiring a later CMake version.
> 
> yes not relationship with this PR. In reviewing you commit i just wondered if we we still support cmake from version 2.6

A quick look through the CMake code didn't catch my attention on anything that requires a newer CMake version, so in that sense there is not a strong reason to bump the minimal required.
But Kitware has CMake 3.0.2 as oldest version on in the online documentation:
https://cmake.org/cmake/help/v3.0/
so simply because of that, it could be wise to bump to `3.0.2`

Any Zephyr user will already have `3.13.1` or newer, as that is the minimal version in Zephyr:
https://github.com/zephyrproject-rtos/zephyr/blob/master/cmake/app/boilerplate.cmake#L29

PR #216, opened 2020-07-20
Ensure C++ compatibility of headers

sleiner:
Merging this ensures that all open-amp headers can be included in a C++ translation unit without the compiler throwing errors.

This was tested with GCC 7.3.1 for arm64 with `std=c++17`.


sleiner:
Even though checkpath shows as successful, it actually did throw an error:
```
-:21: ERROR:POINTER_LOCATION: "(foo*)" should be "(foo *)"
#21: FILE: lib/include/openamp/virtqueue.h:208:
+ vqs = (struct virtqueue*)metal_allocate_memory(vq_size);

- total: 1 errors, 0 warnings, 8 lines checked

NOTE: For some of the reported defects, checkpatch may be able to
mechanically convert to the typical style using --fix or --fix-inplace.

Your patch has style problems, please review.
```

I'm going to fix this.

Issue #220, opened 2020-10-05
rpmsg: supress broken  rpmsg_init_ept API


arnopo:
The rpmsg_init_ept does not seem to be usable by application but is used internally. The proposal is to mark this API deprecated  API and in at least two full release, to remove this API as the standard API to use is  rpmsg_create_ept.
In parallel, the online function must be moved to rpmsg_internal.h

Issue entered in reference to :

- https://github.com/OpenAMP/open-amp/pull/204#discussion_r423218842

- https://github.com/OpenAMP/open-amp/pull/208#pullrequestreview-420375651

arnopo:
Hi @Ed,
Proposal is to use similar mechanism as one used in zephyr (https://docs.zephyrproject.org/latest/development_process/api_lifecycle.html)
Introducing the deprecated attribute ( supported by gcc and IAR) allows to continue to use the API, but with a warning at compilation time:
` warning: ‘rpmsg_init_ept’ is deprecated [-Wdeprecated-declarations]`

To deprecate a function we just need to update the function declaration in the API header:
```
-static inline void rpmsg_init_ept(struct rpmsg_endpoint *ept,
+__deprecated static inline void rpmsg_init_ept(struct rpmsg_endpoint *ept,

```
and the "-Wno-deprecated-declarations" compilation option can disable the warning on need.

What is your feeling about this solution?
i plan to send associated PR tomorrow if this looks good to you.

edmooring:
Hi @arnopo,
This seems to me to be a good solution.

PR #221, opened 2020-10-06
rpmsg: set rpmsg_init_ept API as deprecated

arnopo:
This PR is related to the issue OpenAMP/open-amp#220

The rpmsg_init_ept does not seem to be usable by application
but is used internally. This patch mark this API deprecated
and in at least two full release, to remove this API as the
standard API to use is rpmsg_create_ept.
In addition, move the rpmsg_init_ept to rpmsg_internal.h and rename it.

### DIFF
--- a/lib/rpmsg/rpmsg_internal.h
+++ b/lib/rpmsg/rpmsg_internal.h
@@ -87,6 +87,33 @@ struct rpmsg_ns_msg {
 	uint32_t flags;
 } METAL_PACKED_END;
 
+/**
+ * rpmsg_initialize_ept - initialize rpmsg endpoint
+ *
+ * Initialize an RPMsg endpoint with a name, source address,
+ * remoteproc address, endpoint callback, and destroy endpoint callback.
+ *
+ * @ept: pointer to rpmsg endpoint
+ * @name: service name associated to the endpoint
+ * @src: local address of the endpoint
+ * @dest: target address of the endpoint
+ * @cb: endpoint callback
+ * @ns_unbind_cb: end point service unbind callback, called when remote ept is
+ *                destroyed.
+ */
+static inline void rpmsg_initialize_ept(struct rpmsg_endpoint *ept,
+					const char *name,
+					uint32_t src, uint32_t dest,
+					rpmsg_ept_cb cb,
+					rpmsg_ns_unbind_cb ns_unbind_cb)
+{
+	strncpy(ept->name, name ? name : "", sizeof(ept->name));
+	ept->addr = src;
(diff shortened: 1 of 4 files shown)


lib/rpmsg/rpmsg_internal.h
```diff
@@ -87,6 +87,33 @@ struct rpmsg_ns_msg {
...
+					const char *name,
+					uint32_t src, uint32_t dest,
+					rpmsg_ept_cb cb,
+					rpmsg_ns_unbind_cb ns_unbind_cb)
+{
+	strncpy(ept->name, name ? name : "", sizeof(ept->name));
+	ept->addr = src;
+	ept->dest_addr = dest;
+	ept->cb = cb;
+	ept->ns_unbind_cb = ns_unbind_cb;
```
xiaoxiang781216, line 114: can you call rpmsg_register_endpoint here? ept is useless without this call.

arnopo: I saw your proposal in your PR. I hesitated because rpmsg_initialize_ept is now only a helper function to initialize the ept.

I kept it that way to minimize the impact of the code, in case someone complains about using this API. Even though I don't think the API is usable.
So I would prefer to minimize the update in this step and then, when we will remove the API (in at least two versions), we could merge rpmsg_initialize_ept into rpmsg_register_endpoint. 
Do you see any potential issue to do it  in this way ?

@edmooring : any opinion on this point?

xiaoxiang781216: It's better to make the new API work like rpmsg_init_ept at the beginning, or don't add it at all. It's hard to change the behaviour late without breaking the compatibility.

arnopo: No sure to understand your point as there is no new API. The rpmsg_initialize_ept  is in rpmsg_internal.h not exposed to application. 
So no compatibility issue just a refactoring of some internal functions. 
AT the end,  this is probably more a question of reverting a patch, so both solutions are ok for me , just waiting @edmooring  feedback before deciding.

edmooring: I agree with Arnaud about minimizing the change in this step. Keep the existing behavior and deprecate the API.

edmooring (approved):
LGTM.

lib/rpmsg/rpmsg_internal.h
arnopo: Ok, let do this in 2 steps, we will optimize the code when we will suppress the API after the deprecated period.

PR #224, opened 2020-10-09
Add <errno.h> explicitly to all files that use errno values.

edmooring:
When building OpenAMP apps for FreeRTOS on Xilinx hardware, there
were compile errors due to EINVAL not being defined. On other
platforms, <errno.h> was implicitly included by OpenAMP header
files. This change explicitly includes it where it is needed
in the code.

### DIFF
--- a/apps/system/linux/machine/generic/platform_info.c
+++ b/apps/system/linux/machine/generic/platform_info.c
@@ -26,6 +26,7 @@
 #include <metal/utilities.h>
 #include <openamp/remoteproc.h>
 #include <openamp/rpmsg_virtio.h>
+#include <errno.h>
 #include <poll.h>
 #include <string.h>
 #include <stdio.h>
--- a/apps/examples/load_fw/common.h
+++ b/apps/examples/load_fw/common.h
@@ -14,6 +14,7 @@
 #include <metal/utilities.h>
 #include <openamp/remoteproc.h>
 #include <openamp/remoteproc_loader.h>
+#include <errno.h>
 #include <stdarg.h>
 #include <stdio.h>
 /* Xilinx headers */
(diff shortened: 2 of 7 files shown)


apps/system/linux/machine/generic/platform_info.c
```diff
@@ -26,6 +26,7 @@
 #include <metal/utilities.h>
 #include <openamp/remoteproc.h>
 #include <openamp/rpmsg_virtio.h>
+#include <errno.h>
```
arnopo, line 29: This one is also needed? seems related to Linux not FreeRTOS

arnopo (commented):
LGTM just a question regarding apps/system/linux/machine/generic/platform_info.c

apps/system/linux/machine/generic/platform_info.c
edmooring: I decided to add it to this file so that everything that uses an errno value directly includes the relevant header. This makes it clear that the use of the errno constants is intended.

apps/examples/load_fw/load_fw.c
```diff
@@ -8,6 +8,7 @@
  */
 
 #include <platform_info.h>
+#include <errno.h>
```
arnopo, line 11: for load_fw should be in common.h with other includes

apps/system/linux/machine/generic/platform_info.c
arnopo: Right!  i crosschecked the code, i did not seen any instance that you not address

apps/examples/load_fw/load_fw.c
edmooring: Done.

arnopo (approved):

PR #228, opened 2020-10-16
CI tests: add build test actions

arnopo:
Add build tests for Linux, arm platform and zephyr.
No build test for freeRTOS as the openamp library is relying on libmetal and so is FreeRTOS agnostic  

Notice that for Zephyr,the openamp and libmetal repos have been forked.
As consequence for the build the associated openamp and libmetal modules are updated with the current version including new patches and some Zephyr samples are compiled idirectly in the zephyr  environment.

### DIFF
--- a/.github/actions/build_ci/README.md
+++ b/.github/actions/build_ci/README.md
@@ -1,0 +1,20 @@
+# OpenAMP lib build check docker action
+
+This action test builds for a specified target.
+
+## Inputs
+
+### `target`
+
+**Required** the build target. Default `"linux"`.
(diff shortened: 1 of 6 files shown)


.github/actions/build_ci/README.md
```diff
@@ -0,0 +1,21 @@
+# Libmetal build check docker action
```
edmooring, line 1: Should this be "OpenAMP" instead of "Libmetal"?

edmooring (approved):
Other than that, LGTM.

Issue #231, opened 2020-10-21
apps: platform_release_rpmsg_vdev not implemented


arnopo:
platform_release_rpmsg_vdev fonction is void in applications. Need to free resources (e.g virtio) allocated in platform_create_rpmsg_vdev.

List of files potentially concerned:
apps/system/linux/machine/generic/platform_info.c
apps/system/linux/machine/generic/platform_info.h
apps/machine/zynq7/platform_info.c
apps/machine/zynq7/platform_info.h
apps/machine/zynqmp_r5/platform_info.c
apps/machine/zynqmp_r5/platform_info.h
apps/machine/zynqmp/platform_info.c
apps/machine/zynqmp/platform_info.h

bentheredonethat:
working on branch for this now

PR #233, opened 2020-10-21
Upstream issue 231

bentheredonethat:
as per https://github.com/OpenAMP/open-amp/issues/231 
update platform_release_rpmsg_vdev  to cleanup openamp environment

### DIFF
--- a/apps/examples/echo/rpmsg-echo.c
+++ b/apps/examples/echo/rpmsg-echo.c
@@ -104,7 +104,7 @@ int main(int argc, char *argv[])
 			ret = -1;
 		} else {
 			app(rpdev, platform);
-			platform_release_rpmsg_vdev(rpdev);
+			platform_release_rpmsg_vdev(rpdev, platform);
 			ret = 0;
 		}
 	}
--- a/apps/machine/microblaze_generic/platform_info.c
+++ b/apps/machine/microblaze_generic/platform_info.c
@@ -193,6 +193,18 @@ int platform_poll(void *priv)
 	return 0;
 }
 
+void platform_release_rpmsg_vdev(struct rpmsg_device *rpdev, void *platform)
+{
+	struct rpmsg_virtio_device *rpvdev;
+	struct remoteproc *rproc;
+
+	rpvdev = metal_container_of(rpdev, struct rpmsg_virtio_device, rdev);
+	rproc = platform;
+
+	rpmsg_deinit_vdev(rpvdev);
+	remoteproc_remove_virtio(rproc, rpvdev->vdev);
+}
+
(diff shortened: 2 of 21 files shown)


apps/examples/echo/rpmsg-echo.c
```diff
@@ -85,7 +85,8 @@ int app(struct rpmsg_device *rdev, void *priv)
 int main(int argc, char *argv[])
 {
 	void *platform;
-	struct rpmsg_device *rpdev;
+	struct rpmsg_device *rpdev = NULL;
+	struct rpmsg_virtio_device *rpvdev;
```
arnopo, line 89: The examples should be as agnostic as possible about how the RPmsg is sent. So i would prefer to to not introduce  rpmsg_virtio_device *rpvdev in exemples.
intead do you try to use following method to retrieve the rpmsg_virtio_device  struct in platform_release_rpmsg_vdev?
`struct rpmsg_virtio_device *rvdev = metal_container_of(rpdev, struct rpmsg_virtio_device, rdev);`

apps/machine/microblaze_generic/platform_info.c
```diff
@@ -193,10 +193,10 @@ int platform_poll(void *priv)
 	return 0;
 }
 
-void platform_release_rpmsg_vdev(struct rpmsg_device *rpdev)
+void platform_release_rpmsg_vdev(struct rpmsg_virtio_device *rpvdev)
 {
-	if (rpdev->
-	(void)rpdev;
+	rproc_virtio_remove_vdev(rpvdev->vdev);
+	rpmsg_deinit_vdev(rpvdev);
```
arnopo, line 199: i think the order should be reversed  and using remoteproc_remove_virtio seems more clean.

```
rpmsg_deinit_vdev(rpvdev);
remoteproc_remove_virtio(rproc, rpvdev->vdev);

```

arnopo (changes requested):
Thanks for the proactivity on the issues
Please find few comments

apps/machine/microblaze_generic/platform_info.c
bentheredonethat: so we can rproc from the platform, but otherwise how do you suggest to get that in? also is not rproc cleaned up later on?

bentheredonethat: ok then we can add the demos' platform var which is of type rproc IIRC

bentheredonethat:
updated so only cleanup requires update in demos

arnopo (approved):
LGTM

edmooring (approved):
Tested on ZCU102.

Looks good.

edmooring (approved):
Tested on QEMU.
Looks good.

PR #241, opened 2020-11-01
rpmsg: virito: limit the buffer allocate from shared memory pool

xiaoxiang781216:
rpmsg_virtio_get_tx_buffer shouldn't allocate the number of
buffer bigger than the virtio ring length of sending
report here: https://github.com/OpenAMP/open-amp/pull/162#discussion_r513496611

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -143,7 +143,7 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
 #ifndef VIRTIO_SLAVE_ONLY
 	if (role == RPMSG_MASTER) {
 		data = virtqueue_get_buffer(rvdev->svq, len, idx);
-		if (!data) {
+		if (!data && rvdev->svq->vq_free_cnt) {
 			data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
 							RPMSG_BUFFER_SIZE);
 			*len = RPMSG_BUFFER_SIZE;


arnopo:
This issue seems quite old and related to the first introduction of the zero copy that introduce index (b16ca550f1d5734692d5204d148a1ed2ffb680c7)...

Here I would be agree with @kr-satish  https://github.com/OpenAMP/open-amp/pull/162#discussion_r513496611:
"all buffers should be created at the time of initialization".

I think in a first step we have to figure out if adding new buffer in TX  for master, if no more buffer is available, is a good strategy...
I had a look to Linux  implementation and i don't see such approach...

@wjliang @MarekNovakNXP and @eanjum: I would like to require your memory for this topic...
- As you have the legacy, do you know the reason of such implementation?
- Do you think that it is still needed, or could we cleanup this implementation?

xiaoxiang781216:
No, Linux use the similar strategy:
https://github.com/torvalds/linux/blob/master/drivers/rpmsg/virtio_rpmsg_bus.c#L447-L468
The difference is how to determine whether to allocate memory from the pool. Linux use last_sbuf, OpenAMP depend on the returned value from rpmsg_virtio_shm_pool_get_buffer before, now use vq_free_cnt.

arnopo (approved):
I re-parsed the code to better understood the impact of not limiting the allocation. I think you are right, allocating more buffer than the number of entry of the vrings is a bug. 
And i don't see any code that try to extend the TX vrings...
So LGTM

Thanks,

edmooring (approved):
This looks good. We really shouldn't be trying to allocate more buffers than there are entries in the ring.

PR #244, opened 2020-11-20
apps: rpc_demo: fix compilation warning

arnopo:
The GCC compiler complains about the printf of a size_t variable.
add 'z' length modifier to fix the warning.


edmooring (approved):
Looks good to me.

PR #247, opened 2020-12-03
Openamp bug fix on RPMSG_NS_DESTROY & rpmsg_send failed

GUIDINGLI:
openamp: rpmsg_send() shouldn't check buffer size when get buffer failed
    
    cause _rpmsg_virtio_get_buffer_size() always return size > 0,
    when role is RPMSG_MASTER.

   openamp: fix ns_unbind_cb error when meet RPMSG_NS_DESTROY

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -559,7 +559,7 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 			_ept->dest_addr = RPMSG_ADDR_ANY;
 		metal_mutex_release(&rdev->lock);
 		if (_ept && _ept->ns_unbind_cb)
-			_ept->ns_unbind_cb(ept);
+			_ept->ns_unbind_cb(_ept);
 	} else {
 		if (!_ept) {
 			/*
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -322,6 +322,7 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		metal_mutex_release(&rdev->lock);
 		if (rp_hdr || !tick_count)
 			break;
+
 		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
 		tick_count--;
 	}


GUIDINGLI:
@arnopo  Why the checks failed ?  This will check my Signed-off ?   What's the example?

arnopo (changes requested):
LGTM , just need to update the commits messages

arnopo:
> @arnopo Why the checks failed ? This will check my Signed-off ? What's the example?

The compliance check is waiting a sign-of format with this format: First Last <email>.
So could you please provide your first name and last name in your signed-off-by?

GUIDINGLI:
> > @arnopo Why the checks failed ? This will check my Signed-off ? What's the example?
> 
> The compliance check is waiting a sign-of format with this format: First Last .
> So could you please provide your first name and last name in your signed-off-by?

done

arnopo:
Compliance check still fails because your git config is not pro pertly configured  you probably have something like that:
$ git config -l
...
user.name=ligd
...
should be  user.name=Guiding Li

it Ok for this time but i would suggest to change this for future commits. This compliance test is use in several open source project.

arnopo (approved):

GUIDINGLI:
> Compliance check still fails because your git config is not pro pertly configured you probably have something like that:
> $ git config -l
> ...
> user.name=ligd
> ...
> should be user.name=Guiding Li
> 
> it Ok for this time but i would suggest to change this for future commits. This compliance test is use in several open source project.

This check is not reasonable I think.
If someone in some country don't have divided on  lastname firstname. 
Or if someone just have one short English name, like Mike David Lucy ...
What will he/she do ?

arnopo:
> This check is not reasonable I think.
This check is mainly to encourage contributor to use their real name and not a pseudo or anonymous contribution.
It is a good practice that help to identify the contributor of a patch.
If you have a look to some other opensource projects ( such as Linux or Zephy), you will find this kind of check mainly based on gitlint tool.

> If someone in some country don't have divided on lastname firstname.
> Or if someone just have one short English name, like Mike David Lucy ...
> What will he/she do ?
If we face the case we will just discuss together to find the best solution. At the end having this check fails does not mean that we do not accept the commit.

edmooring (approved):
Looks good to me.

PR #277, opened 2021-03-19
CI Fix compliance  test

arnopo:
A false positive is returned by the compliance check when the script crashes.

This pull request is a cherry-pick of the Zephyr project that already fixes the issue
- fix the script to return fail, in case of crash,
- fix the crash linked to junitparser library version compatibility.


edmooring (approved):
Looks good to me.

PR #281, opened 2021-04-15
Version update

arnopo:
As the openAMP is now quite stable, a proposition is to create the V1 release. and then to follow semver rules.

In this scope the pull request:
- update the readme to provide a reference to the semver
- create a VERSION file to make it easier to find and update the version.
- Add a new API to get openamp library version.


galak (commented):
Are the defines for version accessible to the user?

So one could possibly use them to ifdef code if an API changed for example?

arnopo:
> Are the defines for version accessible to the user?
> 
> So one could possibly use them to ifdef code if an API changed for example?
No the defines are not accessible.
I hesitated... we need the version to generate the dynamic library. To expose the define without duplicating the version definitions, the version.h should be generated by the cmake.  That would create compilation dependencies for user applications.
If you have a solution to propose don't hesitate.

arnopo:
Zephyr CI test fails because the new VERSION file is not copied in zephyr during test. I need to fix this

edmooring (approved):
Looks good to me.

arnopo:
@galak, @edmooring, all,

I added a patch on top of, to address the pre-processing point. 

If anyone has a better approach, please feel free. We can also just postpone this until after the release.

Here is an exemple of the "version_def.h" file generated in the install dir ( build/bin/usr/local/include/openamp/version_def.h).
```
/*
 * Copyright (c) 2021, STMicroelectronics.
 *
 * SPDX-License-Identifier: BSD-3-Clause
 */

/*
 * @file	version_def.h
 * @brief	generated version definition for OpenAMP library.
 */

#ifndef __OPENAMP_VERSION_DEF__H__
#define __OPENAMP_VERSION_DEF__H__

/*  values come from cmake/option.cmake */

#define OPENAMP_VERSION_MAJOR  0
#define OPENAMP_VERSION_MINOR  1
#define OPENAMP_VERSION_PATCH  0
#define OPENAMP_VERSION        "0.1.0"

#endif /* __OPENAMP_VERSION_DEF__H__ */
```
Same could be done in libmetal.

edmooring (approved):
Looks good to me.

PR #283, opened 2021-04-23
rpmsg: fix variable overflow

arnopo:
The idx variable size is 16-bit. Add a cast to fix "expression may overflow"
code analysis complain.


edmooring (approved):
Looks good to me.

PR #288, opened 2021-05-04
CI: Zephyr test: copy whole libmetal directory for test

arnopo:
When only the libmetal/lib is copied, some files such as VERSION or CmakeList.txt are not copied. This can lead to a test error or false positives.

This PR  fixes CI test that does not highlight issue resolved in the https://github.com/OpenAMP/libmetal/pull/154.


edmooring (approved):
Looks good to me.

Issue #292, opened 2021-06-03
VirtQueue should use metal_cache_* APIs


carlocaione:
On some platforms the VirtQueue and Vring structures could be stored in cacheable memory. When a data cache is present OpenAPM should take care of flushing / invalidating vring structures and control block when needed to keep consistency, this should be done using the `metal_cache_*` APIs.

arnopo:
As the cache management is platform dependent, the libmetal is a better place to handle the cache. 
The open-amp lib should not have to care about the cache memory management.
Regarding your patch in libmetal, I suppose that your need is for Zephyr, right?
Here is an example in libmetal for NuttX than could be adapted for Zephyr:
https://github.com/OpenAMP/libmetal/blob/master/lib/system/nuttx/io.c

carlocaione:
@arnopo agree but unfortunately I don't think that's enough. When also the vring control block is in cacheable memory (and shared between remote and master) we need to flush / invalidate also when accessing the struct that is in the shared memory, for example https://github.com/zephyrproject-rtos/open-amp/blob/master/open-amp/lib/virtio/virtqueue.c#L550

arnopo:
You are right! Thanks for pointing  this out to me.

That would means that we should have to use io access to update the virtqueues or to introduce a cmake option for the cache management.

Before we get into the details. It seems to me that you are the first one to point this out and ask for the use of cache memory for RPMsg communication.
Could you explain the rationale behind this request?
Do you observe some performance improvement using cacheable memory?

carlocaione:
> Could you explain the rationale behind this request?
> Do you observe some performance improvement using cacheable memory?

I'm not sure how much I can go into details but a client I'm working with has a platform where OpenAMP is allocated in cacheable memory for reasons not only related to performance improvement.

> That would means that we should have to use io access to update the virtqueues or to introduce a cmake option for the cache management.

Can you expand a bit of the cmake option? Is that an alternative to use `metal_cache_*` APIs to flush / invalidate the virtqueues when needed?

arnopo:
> > Could you explain the rationale behind this request?
> > Do you observe some performance improvement using cacheable memory?
> 
> I'm not sure how much I can go into details but a client I'm working with has a platform where OpenAMP is allocated in cacheable memory for reasons not only related to performance improvement.

So the rational is that they can not allocate in non-cacheable memrory, right? 
I just want to confirm with you that the need is because  your client doesn't know how to allocate non-cacheable memory. :)

> 
[...]

PR #293, opened 2021-06-09
virtqueue:  Introduce cache helpers

carlocaione:
On some platforms the VirtQueue and Vring structures could be stored in cacheable memory. When a data cache is present OpenAPM should take care of flushing / invalidating vring structures and control block when needed to keep consistency, this should be done using the metal_cache_* APIs.

This is an RFC and I'm open to any comment / suggestion on how to address this issue the proper way.

Fixes: #292

Signed-off-by: Carlo Caione <ccaione@baylibre.com>

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -74,6 +74,18 @@ if (NOT WITH_VIRTIO_SLAVE)
   add_definitions(-DVIRTIO_MASTER_ONLY)
 endif (NOT WITH_VIRTIO_SLAVE)
 
+option (WITH_DCACHE_VRINGS "Build with vrings cache operations enabled" OFF)
+
+if (WITH_DCACHE_VRINGS)
+  add_definitions(-DVIRTIO_CACHED_VRINGS)
+endif (WITH_DCACHE_VRINGS)
+
+option (WITH_DCACHE_BUFFERS "Build with vrings cache operations enabled" OFF)
+
+if (WITH_DCACHE_BUFFERS)
+  add_definitions(-DVIRTIO_CACHED_BUFFERS)
+endif (WITH_DCACHE_BUFFERS)
+
 # Set the complication flags
 set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
 
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -8,6 +8,7 @@
  */
 
 #include <metal/alloc.h>
+#include <metal/cache.h>
 #include <metal/sleep.h>
 #include <metal/utilities.h>
 #include <openamp/rpmsg_virtio.h>
@@ -102,6 +103,11 @@ static int rpmsg_virtio_enqueue_buffer(struct rpmsg_virtio_device *rvdev,
 				       uint16_t idx)
 {
 	unsigned int role = rpmsg_virtio_get_role(rvdev);
+
+#ifdef VIRTIO_CACHED_BUFFERS
+	metal_cache_flush(buffer, len);
+#endif /* VIRTIO_CACHED_BUFFERS */
+
 #ifndef VIRTIO_SLAVE_ONLY
 	if (role == RPMSG_MASTER) {
 		struct virtqueue_buf vqbuf;
@@ -191,6 +197,11 @@ static void *rpmsg_virtio_get_rx_buffer(struct rpmsg_virtio_device *rvdev,
 		    virtqueue_get_available_buffer(rvdev->rvq, idx, len);
 	}
 #endif /*!VIRTIO_MASTER_ONLY*/
+
+#ifdef VIRTIO_CACHED_BUFFERS
+	/* Invalidate the buffer before returning it */
+	metal_cache_invalidate(data, *len);
+#endif /* VIRTIO_CACHED_BUFFERS */
 
 	return data;
 }
--- a/lib/virtio/virtqueue.c
+++ b/lib/virtio/virtqueue.c
@@ -11,6 +11,7 @@
 #include <metal/atomic.h>
 #include <metal/log.h>
 #include <metal/alloc.h>
+#include <metal/cache.h>
 
 /* Prototype for internal functions. */
 static void vq_ring_init(struct virtqueue *, void *, int);
@@ -27,6 +28,14 @@ static int virtqueue_nused(struct virtqueue *vq);
 #ifndef VIRTIO_MASTER_ONLY
 static int virtqueue_navail(struct virtqueue *vq);
 #endif
+
+#ifdef VIRTIO_CACHED_VRINGS
+#define VRING_FLUSH(x)		metal_cache_flush(&x, sizeof(x))
+#define VRING_INVALIDATE(x)	metal_cache_invalidate(&x, sizeof(x))
+#else
+#define VRING_FLUSH(x)		do { } while (0)
+#define VRING_INVALIDATE(x)	do { } while (0)
+#endif /* VIRTIO_CACHED_VRINGS */
 
 /* Default implementation of P2V based on libmetal */
 static inline void *virtqueue_phys_to_virt(struct virtqueue *vq,
@@ -87,6 +96,11 @@ int virtqueue_create(struct virtio_device *virt_dev, unsigned short id,
 		/* Initialize vring control block in virtqueue. */
 		vq_ring_init(vq, ring->vaddr, ring->align);
 	}
+
+	/*
+	 * CACHE: nothing to be done here. Only desc.next is setup at this
+	 * stage but that is only written by master, so no need to flush it.
+	 */
 
 	return status;
 }
@@ -174,6 +188,9 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 	void *cookie;
 	uint16_t used_idx, desc_idx;
 
+	/* Used.idx is updated by slave, so we need to invalidate */
+	VRING_INVALIDATE(vq->vq_ring.used->idx);
+
 	if (!vq || vq->vq_used_cons_idx == vq->vq_ring.used->idx)
 		return NULL;
 
@@ -183,6 +200,9 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 	uep = &vq->vq_ring.used->ring[used_idx];
 
 	atomic_thread_fence(memory_order_seq_cst);
+
+	/* Used.ring is written by slave, invalidate it */
+	VRING_INVALIDATE(vq->vq_ring.used->ring[used_idx]);
 
 	desc_idx = (uint16_t)uep->id;
 	if (len)
@@ -202,6 +222,7 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 
 uint32_t virtqueue_get_buffer_length(struct virtqueue *vq, uint16_t idx)
 {
+	VRING_INVALIDATE(vq->vq_ring.desc[idx].len);
 	return vq->vq_ring.desc[idx].len;
 }
 
@@ -241,15 +262,23 @@ void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
 	void *buffer;
 
 	atomic_thread_fence(memory_order_seq_cst);
+
+	/* Avail.idx is updated by master, invalidate it */
+	VRING_INVALIDATE(vq->vq_ring.avail->idx);
 	if (vq->vq_available_idx == vq->vq_ring.avail->idx) {
 		return NULL;
 	}
 
 	VQUEUE_BUSY(vq);
 
+	/* Avail.ring is updated by master, invalidate it */
+	VRING_INVALIDATE(vq->vq_ring.avail->ring[head_idx]);
+
 	head_idx = vq->vq_available_idx++ & (vq->vq_nentries - 1);
 	*avail_idx = vq->vq_ring.avail->ring[head_idx];
 
+	/* Invalidate the desc entry written by master before accessing it */
+	VRING_INVALIDATE(vq->vq_ring.desc[*avail_idx]);
 	buffer = virtqueue_phys_to_virt(vq, vq->vq_ring.desc[*avail_idx].addr);
 	*len = vq->vq_ring.desc[*avail_idx].len;
 
@@ -279,14 +308,21 @@ int virtqueue_add_consumed_buffer(struct virtqueue *vq, uint16_t head_idx,
 
 	VQUEUE_BUSY(vq);
 
+	/* CACHE: used is never written by master, so it's safe to directly access it */
 	used_idx = vq->vq_ring.used->idx & (vq->vq_nentries - 1);
 	used_desc = &vq->vq_ring.used->ring[used_idx];
 	used_desc->id = head_idx;
 	used_desc->len = len;
 
+	/* We still need to flush it because this is read by master */
+	VRING_FLUSH(vq->vq_ring.used->ring[used_idx]);
+
 	atomic_thread_fence(memory_order_seq_cst);
 
 	vq->vq_ring.used->idx++;
+
+	/* Used.idx is read by master, so we need to flush it */
+	VRING_FLUSH(vq->vq_ring.used->idx);
 
 	/* Keep pending count until virtqueue_notify(). */
 	vq->vq_queued_cnt++;
@@ -323,22 +359,28 @@ void virtqueue_disable_cb(struct virtqueue *vq)
 		if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
 			vring_used_event(&vq->vq_ring) =
 			    vq->vq_used_cons_idx - vq->vq_nentries - 1;
+			VRING_FLUSH(vring_used_event(&vq->vq_ring));
 		}
 #endif /*VIRTIO_SLAVE_ONLY*/
 #ifndef VIRTIO_MASTER_ONLY
 		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE) {
 			vring_avail_event(&vq->vq_ring) =
 			    vq->vq_available_idx - vq->vq_nentries - 1;
+			VRING_FLUSH(vring_avail_event(&vq->vq_ring));
 		}
 #endif /*VIRTIO_MASTER_ONLY*/
 	} else {
 #ifndef VIRTIO_SLAVE_ONLY
-		if (vq->vq_dev->role == VIRTIO_DEV_MASTER)
+		if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
 			vq->vq_ring.avail->flags |= VRING_AVAIL_F_NO_INTERRUPT;
+			VRING_FLUSH(vq->vq_ring.avail->flags);
+		}
 #endif /*VIRTIO_SLAVE_ONLY*/
 #ifndef VIRTIO_MASTER_ONLY
-		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE)
+		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE) {
 			vq->vq_ring.used->flags |= VRING_USED_F_NO_NOTIFY;
+			VRING_FLUSH(vq->vq_ring.used->flags);
+		}
 #endif /*VIRTIO_MASTER_ONLY*/
 	}
 
@@ -375,6 +417,9 @@ void virtqueue_dump(struct virtqueue *vq)
 	if (!vq)
 		return;
 
+	VRING_INVALIDATE(vq->vq_ring.avail);
+	VRING_INVALIDATE(vq->vq_ring.used);
+
 	metal_log(METAL_LOG_DEBUG,
 		  "VQ: %s - size=%d; free=%d; queued=%d; "
 		  "desc_head_idx=%d; avail.idx=%d; used_cons_idx=%d; "
@@ -399,14 +444,24 @@ uint32_t virtqueue_get_desc_size(struct virtqueue *vq)
 	uint16_t avail_idx = 0;
 	uint32_t len = 0;
 
+	/* Avail.idx is updated by master, invalidate it */
+	VRING_INVALIDATE(vq->vq_ring.avail->idx);
+
 	if (vq->vq_available_idx == vq->vq_ring.avail->idx) {
 		return 0;
 	}
 
 	VQUEUE_BUSY(vq);
 
+	/* Avail.ring is updated by master, invalidate it */
+	VRING_INVALIDATE(vq->vq_ring.avail->ring[head_idx]);
+
 	head_idx = vq->vq_available_idx & (vq->vq_nentries - 1);
 	avail_idx = vq->vq_ring.avail->ring[head_idx];
+
+	/* Invalidate the desc entry written by master before accessing it */
+	VRING_INVALIDATE(vq->vq_ring.desc[avail_idx].len);
+
 	len = vq->vq_ring.desc[avail_idx].len;
 
 	VQUEUE_IDLE(vq);
@@ -440,6 +495,7 @@ static uint16_t vq_ring_add_buffer(struct virtqueue *vq,
 		VQASSERT(vq, idx != VQ_RING_DESC_CHAIN_END,
 			 "premature end of free desc chain");
 
+		/* CACHE: No need to invalidate desc because it is only written by master */
 		dp = &desc[idx];
 		dp->addr = virtqueue_virt_to_phys(vq, buf_list[i].buf);
 		dp->len = buf_list[i].len;
@@ -454,6 +510,13 @@ static uint16_t vq_ring_add_buffer(struct virtqueue *vq,
 		 */
 		if (i >= readable)
 			dp->flags |= VRING_DESC_F_WRITE;
+
+		/*
+		 * Instead of flushing the whole desc region, we flush only the
+		 * single entry hopefully saving some cycles
+		 */
+		VRING_FLUSH(desc[idx]);
+
 	}
 
 	return idx;
@@ -469,6 +532,7 @@ static void vq_ring_free_chain(struct virtqueue *vq, uint16_t desc_idx)
 	struct vring_desc *dp;
 	struct vq_desc_extra *dxp;
 
+	/* CACHE: desc is never written by slave, no need to invalidate */
 	VQ_RING_ASSERT_VALID_IDX(vq, desc_idx);
 	dp = &vq->vq_ring.desc[desc_idx];
 	dxp = &vq->vq_descx[desc_idx];
@@ -495,6 +559,8 @@ static void vq_ring_free_chain(struct virtqueue *vq, uint16_t desc_idx)
 	 * We must append the existing free chain, if any, to the end of
 	 * newly freed chain. If the virtqueue was completely used, then
 	 * head would be VQ_RING_DESC_CHAIN_END (ASSERTed above).
+	 *
+	 * CACHE: desc.next is never read by slave, no need to flush it.
 	 */
 	dp->next = vq->vq_desc_head_idx;
 	vq->vq_desc_head_idx = desc_idx;
@@ -541,13 +607,21 @@ static void vq_ring_update_avail(struct virtqueue *vq, uint16_t desc_idx)
 	 * deferring to virtqueue_notify() in the hopes that if the host is
 	 * currently running on another CPU, we can keep it processing the new
 	 * descriptor.
+	 *
+	 * CACHE: avail is never written by slave, so it is safe to not invalidate here
 	 */
 	avail_idx = vq->vq_ring.avail->idx & (vq->vq_nentries - 1);
 	vq->vq_ring.avail->ring[avail_idx] = desc_idx;
 
+	/* We still need to flush the ring */
+	VRING_FLUSH(vq->vq_ring.avail->ring[avail_idx]);
+
 	atomic_thread_fence(memory_order_seq_cst);
 
 	vq->vq_ring.avail->idx++;
+
+	/* And the index */
+	VRING_FLUSH(vq->vq_ring.avail->idx);
 
 	/* Keep pending count until virtqueue_notify(). */
 	vq->vq_queued_cnt++;
@@ -566,23 +640,31 @@ static int vq_ring_enable_interrupt(struct virtqueue *vq, uint16_t ndesc)
 	 */
 	if (vq->vq_dev->features & VIRTIO_RING_F_EVENT_IDX) {
 #ifndef VIRTIO_SLAVE_ONLY
-		if (vq->vq_dev->role == VIRTIO_DEV_MASTER)
+		if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
 			vring_used_event(&vq->vq_ring) =
 				vq->vq_used_cons_idx + ndesc;
+			VRING_FLUSH(vring_used_event(&vq->vq_ring));
+		}
 #endif /*VIRTIO_SLAVE_ONLY*/
 #ifndef VIRTIO_MASTER_ONLY
-		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE)
+		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE) {
 			vring_avail_event(&vq->vq_ring) =
 				vq->vq_available_idx + ndesc;
+			VRING_FLUSH(vring_avail_event(&vq->vq_ring));
+		}
 #endif /*VIRTIO_MASTER_ONLY*/
 	} else {
 #ifndef VIRTIO_SLAVE_ONLY
-		if (vq->vq_dev->role == VIRTIO_DEV_MASTER)
+		if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
 			vq->vq_ring.avail->flags &= ~VRING_AVAIL_F_NO_INTERRUPT;
+			VRING_FLUSH(vq->vq_ring.avail->flags);
+		}
 #endif /*VIRTIO_SLAVE_ONLY*/
 #ifndef VIRTIO_MASTER_ONLY
-		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE)
+		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE) {
 			vq->vq_ring.used->flags &= ~VRING_USED_F_NO_NOTIFY;
+			VRING_FLUSH(vq->vq_ring.used->flags);
+		}
 #endif /*VIRTIO_MASTER_ONLY*/
 	}
 
@@ -635,8 +717,10 @@ static int vq_ring_must_notify(struct virtqueue *vq)
 	if (vq->vq_dev->features & VIRTIO_RING_F_EVENT_IDX) {
 #ifndef VIRTIO_SLAVE_ONLY
 		if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
+			/* CACHE: no need to invalidate avail */
 			new_idx = vq->vq_ring.avail->idx;
 			prev_idx = new_idx - vq->vq_queued_cnt;
+			VRING_INVALIDATE(vring_avail_event(&vq->vq_ring));
 			event_idx = vring_avail_event(&vq->vq_ring);
 			return vring_need_event(event_idx, new_idx,
 						prev_idx) != 0;
@@ -644,8 +728,10 @@ static int vq_ring_must_notify(struct virtqueue *vq)
 #endif /*VIRTIO_SLAVE_ONLY*/
 #ifndef VIRTIO_MASTER_ONLY
 		if (vq->vq_dev->role == VIRTIO_DEV_SLAVE) {
+			/* CACHE: no need to invalidate used */
 			new_idx = vq->vq_ring.used->idx;
 			prev_idx = new_idx - vq->vq_queued_cnt;
+			VRING_INVALIDATE(vring_used_event(&vq->vq_ring));
 			event_idx = vring_used_event(&vq->vq_ring);
 			return vring_need_event(event_idx, new_idx,
 						prev_idx) != 0;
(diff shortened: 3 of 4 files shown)


edmooring (commented):
This seems painful, but manual cache management generally is.

lib/virtio/virtqueue.c
```diff
@@ -28,6 +29,20 @@ static int virtqueue_nused(struct virtqueue *vq);
 static int virtqueue_navail(struct virtqueue *vq);
 #endif
 
+#ifdef VIRTQUEUE_CACHE_OPS
+static inline void vq_ring_contro_block_flush(struct virtqueue *vq, unsigned
```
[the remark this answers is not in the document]
carlocaione: Because typos are part of life :) Fixing.

carlocaione:
> This seems painful, but manual cache management generally is.

It is indeed. The tricky part is adding the flush / invalidate ops only when strictly needed.

lib/virtio/virtqueue.c
```diff
@@ -637,6 +760,11 @@ static int vq_ring_must_notify(struct virtqueue *vq)
 		if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
 			new_idx = vq->vq_ring.avail->idx;
 			prev_idx = new_idx - vq->vq_queued_cnt;
+
+#ifdef VIRTQUEUE_CACHE_OPS
```
arnopo, line 764: do you need to re invalidate ? probably already done when get buffer for TX ( especially if you invalidate the whole structure)

arnopo (changes requested):
i finished a first review : few comments in the  code + following comments:

- I wonder, for cache invalidation (and maybe flush), if we invalidate the cache once on  the whole structure ( used or available) before processing a new buffer could simplify the code, without significant performance impact since the structures are not large. 

- seems missing cache management in vq_ring_free_chain

carlocaione:
>  I wonder, for cache invalidation (and maybe flush), if we invalidate the cache once on  the whole structure ( used or available) before processing a new buffer could simplify the code, without significant performance impact since the structures are not large.

That solution was indeed discussed internally before pushing this PR but the conclusion was that performance could suffer in that case.

arnopo:
> > I wonder, for cache invalidation (and maybe flush), if we invalidate the cache once on  the whole structure ( used or available) before processing a new buffer could simplify the code, without significant performance impact since the structures are not large.
> 
> That solution was indeed discussed internally before pushing this PR but the conclusion was that performance could suffer in that case.

Do you performed some tests to confirm? if you have some data about this could you share them?

For the reliability of the code, it seems to me that minimizing the places where we call metal_cache* is also important.

Concerning performance, i'm trying to understand the impact in term of performance:
Let's take an example with the vring_avail and vring_used structures for 16 buffers, 64-byte data cache ( it base on code review, don't hesitate to correct me if i have some errors in my reasonning).
* vring_avail: the size of the structure would be 36 Bytes so one 64-byte cache line.
* vring_used: the sizeof struct the size would be 4+16* 8 = 132 bytes  => three 64-byte cache lines

I tried  to count  cache invalidates and flushes for a reception and transmission for these 2 structures. It looks like there are about 3-4 cache invalidate/flush per structures.

I wonder if one cache flush/invalidate of n lines would be better in term of performance that n requests of 1 line.

For instance for a slave reception, I would invalidate the cache for the vring_avail struct, parse the vring and get the buffer, use it, release it and flush the vring_usedstruct. Then loop to check for next RX  buffer.

carlocaione:
> Do you performed some tests to confirm? if you have some data about this could you share them?

Not about this, sorry.
 
> For the reliability of the code, it seems to me that minimizing the places where we call metal_cache* is also important.

Totally agree, not the most elegant PR of my career :wink: 

> Concerning performance, i'm trying to understand the impact in term of performance:
> Let's take an example with the vring_avail and vring_used structures for 16 buffers, 64-byte data cache ( it base on code review, don't hesitate to correct me if i have some errors in my reasonning).

On my platform I have a 32-byte data cache line, not far from your assumption.

>     * vring_avail: the size of the structure would be 36 Bytes so one 64-byte cache line.

2 cache lines for a 32-bytes cache line.

>     * vring_used: the sizeof struct the size would be 4+16* 8 = 132 bytes  => three 64-byte cache lines

5 cache lines for a 32-bytes cache line.

But I think we also need to flush / invalidate the vring descriptor struct. That would be 16 bytes => 1 cache line

> I tried to count cache invalidates and flushes for a reception and transmission for these 2 structures. It looks like there are about 3-4 cache invalidate/flush per structures.
> 
> I wonder if one cache flush/invalidate of n lines would be better in term of performance that n requests of 1 line.

Well, for the cache IPs I have seen to flush / invalidate n lines you simply submit n requests to flush / invalidate 1 line so there is full proportionality here (you can usually invalidate the whole cache or 1 line, for all the rest you just use a for cycle on n lines).

> For instance for a slave reception, I would invalidate the cache for the vring_avail struct, parse the vring and get the buffer, use it, release it and flush the vring_usedstruct. Then loop to check for next RX buffer.
 
We should also invalidate the returned buffer.

So without considering the buffer invalidation we would have: 2 + 5 = 7 cache operations.

carlocaione:
> I wonder if one cache flush/invalidate of n lines would be better in term of performance that n requests of 1 line.

So, I did some tests about this and these are the results (in systick cycles and with an hot I-Cache):

Flushing:
- 1 line in the vring_used struct [1 line]: 57 cycles
- the whole vring_used struct [5 lines]: 133 cycles
- 1 line in the vring_avail struct [1 line]: 57 cycles
- the whole vring_avail struct [2 lines]: 72 cycles
- 1 line in the vring_desc struct [1 line]: 57 cycles
- the whole vring_desc struct [1 line]: 57 cycles

Invalidating:
- 1 line in the vring_used struct [1 line]: 27 cycles
- the whole vring_used struct [5 lines]: 71 cycles
- 1 line in the vring_avail struct [1 line]: 27 cycles
- the whole vring_avail struct [2 lines]: 38 cycles
- 1 line in the vring_desc struct [1 line]: 27 cycles
- the whole vring_desc struct [1 line]: 27 cycles

So, it looks like it is indeed convenient performing a flush/inv of n lines instead of n operations on 1 line.

arnopo:
> > I wonder if one cache flush/invalidate of n lines would be better in term of performance that n requests of 1 line.
> 
> So, I did some tests about this and these are the results (in systick cycles and with an hot I-Cache):
> 
> Flushing:
> 
> * 1 line in the vring_used struct [1 line]: 57 cycles
> * the whole vring_used struct [5 lines]: 133 cycles
> * 1 line in the vring_avail struct [1 line]: 57 cycles
> * the whole vring_avail struct [2 lines]: 72 cycles
> * 1 line in the vring_desc struct [1 line]: 57 cycles
> * the whole vring_desc struct [1 line]: 57 cycles
> 
> Invalidating:
> 
> * 1 line in the vring_used struct [1 line]: 27 cycles
> * the whole vring_used struct [5 lines]: 71 cycles
> * 1 line in the vring_avail struct [1 line]: 27 cycles
> * the whole vring_avail struct [2 lines]: 38 cycles
> * 1 line in the vring_desc struct [1 line]: 27 cycles
> * the whole vring_desc struct [1 line]: 27 cycles
> 
> So, it looks like it is indeed convenient performing a flush/inv of n lines instead of n operations on 1 line.

Thanks for the data! 
Yes, it seems that simplifying the code and maintaining performance are compatible.

Please notice also that the vring_desc is read only by slave, as the buffers are allocated by the Master. So for master no cache management seems needed, for slave only invalidating.

carlocaione:
> So for master no cache management seems needed

The problem I see is that the caches for master and slave could be different, so that means that master must flush after writing to be sure that the cache content is flushed to memory, otherwise it could not be seen from the slave, even after invalidating.

arnopo:
> > So for master no cache management seems needed
> 
> The problem I see is that the caches for master and slave could be different, so that means that master must flush after writing to be sure that the cache content is flushed to memory, otherwise it could not be seen from the slave, even after invalidating.

You are totally right! the flush is needed

carlocaione:
@arnopo @edmooring I pushed a new revision this time flushing and invalidating on the whole structs as suggested by @arnopo.

lib/virtio/virtqueue.c
```diff
@@ -28,6 +29,69 @@ static int virtqueue_nused(struct virtqueue *vq);
...
+	struct vring *vr;
+	unsigned int size;
+
+	vr = &vq->vq_ring;
+	size = vring_size(vr->num, vr->align);
+
+	metal_cache_flush(vr->desc, size);
+}
+
+static inline void vring_desc_cache_op(struct virtqueue *vq,
```
arnopo, line 49: following functions look to me no very efficient, what about creating Macro to save the if condition?
 
#define VRING_NUM(virtqueue)  (vq->vq_ring->num)
#define VRING_ALIGN(virtqueue)  (vq->vq_ring->align)
#define VRING_DESC(virtqueue)  (vq->vq_ring->desc)

then instead of calling vring_desc_cache_op(vq,jVRING_CACHE_FLUSH)
just call  
metal_cache_flush(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));

lib/virtio/virtqueue.c
```diff
@@ -184,6 +261,10 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 
 	atomic_thread_fence(memory_order_seq_cst);
 
+#ifdef VIRTIO_D_CACHE
```
arnopo, line 264: why doing it twice time?

arnopo (changes requested):
I'm quite surprised that the number of invalidates and flushes is not decreasing. I will need time to review this patch to determine on RX and TX if there are any duplications that could be optimized.

Waiting a deeper review, please find few comment resulting from  a first quick review

cmake/options.cmake
```diff
@@ -74,6 +74,12 @@ if (NOT WITH_VIRTIO_SLAVE)
   add_definitions(-DVIRTIO_MASTER_ONLY)
 endif (NOT WITH_VIRTIO_SLAVE)
 
+option (WITH_DCACHE_MGT "Build with virtqueue cache operations enabled" OFF)
```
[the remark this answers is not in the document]
carlocaione: ok

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -365,8 +372,11 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 
 	/* Copy data to rpmsg buffer. */
 	io = rvdev->shbuf_io;
-	status = metal_io_block_write(io, metal_io_virt_to_offset(io, hdr),
-				      &rp_hdr, sizeof(rp_hdr));
+	offset = metal_io_virt_to_offset(io, hdr);
+	status = metal_io_block_write(io, offset, &rp_hdr, sizeof(rp_hdr));
+
+	metal_cache_flush(metal_io_virt(io, offset), sizeof(rp_hdr));
```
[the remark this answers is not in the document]
carlocaione: yup, makes sense

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -426,8 +437,11 @@ static int rpmsg_virtio_send_offchannel_raw(struct rpmsg_device *rdev,
 	if (len > (int)buff_len)
 		len = buff_len;
 	io = rvdev->shbuf_io;
-	status = metal_io_block_write(io, metal_io_virt_to_offset(io, buffer),
-				      data, len);
+	offset = metal_io_virt_to_offset(io, buffer);
+	status = metal_io_block_write(io, offset, data, len);
+
+	metal_cache_flush(metal_io_virt(io, offset), len);
```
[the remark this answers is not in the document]
carlocaione: ack

lib/virtio/virtqueue.c
carlocaione: ok

lib/virtio/virtqueue.c
carlocaione: yeah, I was not 100% sure about this and now thinking about this again it looks wrong indeed. I added it twice because of the synchronization fence `atomic_thread_fence` but that's valid only on the same CPU so the second invalidation is not needed.

carlocaione:
> I'm quite surprised that the number of invalidates and flushes is not decreasing

yes, I am too and if I didn't screw this new revision up, it looks like it's still **not convenient** to invalidate / flush the whole structs vs invalidating / flushing the single cache lines involved. So @arnopo please take a look in case I did some ugly mistakes on this revision, but if everything is still going to stay as it is, I'm afraid that I have to revert back to the previous version with flush / invalidate of single cache lines.

lib/virtio/virtqueue.c
```diff
@@ -632,6 +733,13 @@ static int vq_ring_must_notify(struct virtqueue *vq)
 {
 	uint16_t new_idx, prev_idx, event_idx;
 
+#ifdef VIRTIO_D_CACHE
+	metal_cache_invalidate(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
```
arnopo, line 737: only slave should flush this

lib/virtio/virtqueue.c
```diff
@@ -632,6 +733,13 @@ static int vq_ring_must_notify(struct virtqueue *vq)
...
+	metal_cache_invalidate(VRING_AVAIL(vq),
+			       vring_avail_size(VRING_NUM(vq),
```
arnopo, line 739: only master should flush this

lib/virtio/virtqueue.c
```diff
@@ -688,6 +796,10 @@ static int virtqueue_nused(struct virtqueue *vq)
 {
 	uint16_t used_idx, nused;
 
+#ifdef VIRTIO_D_CACHE
+	metal_cache_invalidate(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
```
arnopo, line 800: here seems that you can just invalidate vq->vq_ring.used->idx

lib/virtio/virtqueue.c
```diff
@@ -707,6 +819,12 @@ static int virtqueue_navail(struct virtqueue *vq)
 {
 	uint16_t avail_idx, navail;
 
+#ifdef VIRTIO_D_CACHE
+	metal_cache_invalidate(VRING_AVAIL(vq),
```
arnopo, line 823: here seems that you can just invalidate vq->vq_ring.avail->idx

arnopo (changes requested):
the kind of review that can  give a headhache :-)

here is the callstack I observed reviewing the code:

**Master**:
```
rpmsg_init_vdev
	-> rpmsg_virtio_create_virtqueues
		-> virtio_create_virtqueues
			-> virtqueue_create
				-> vq_ring_init
					-> metal_cache_flush(VRING_DESC(vq),  vring_size(VRING_NUM(vq), VRING_ALIGN(vq)));   ,???????
	vq_ring_add_buffer
					-> metal_cache_flush(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));

```
```
rpmsg_virtio_send_offchannel_raw
	-> rpmsg_virtio_get_tx_payload_buffer
		 -> rpmsg_virtio_get_tx_buffer
			-> virtqueue_get_buffer 
				-> metal_cache_invalidate(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
					-> vq_ring_free_chain
						-> metal_cache_invalidate(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
						-> metal_cache_flush(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
	-> rpmsg_virtio_send_offchannel_nocopy
		-> metal_cache_flush(metal_io_virt(io, offset), len);
		-> rpmsg_virtio_enqueue_buffer
			-> virtqueue_add_buffer 
				-> vq_ring_add_buffer
					 ->  metal_cache_flush(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
				-> vq_ring_update_avail
					-> metal_cache_flush(VRING_AVAIL(vq),  vring_avail_size(VRING_NUM(vq), VRING_ALIGN(vq)));

```
```
rpmsg_virtio_rx_callback
	-> rpmsg_virtio_get_rx_buffer
		-> virtqueue_get_buffer
			-> metal_cache_invalidate(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
					-> vq_ring_free_chain
						-> metal_cache_invalidate(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
						-> metal_cache_flush(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
	-> metal_cache_invalidate(data, *len);
	-> rpmsg_virtio_return_buffer
		-> virtqueue_add_buffer
				-> vq_ring_add_buffer
					  -> metal_cache_flush(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
				-> vq_ring_update_avail
					-> metal_cache_flush(VRING_AVAIL(vq),  vring_avail_size(VRING_NUM(vq), VRING_ALIGN(vq)));

```

**Slave**:
```
rpmsg_init_vdev
	-> rpmsg_virtio_create_virtqueues
		-> virtio_create_virtqueues
			-> virtqueue_create
				-> vq_ring_init

```

```
rpmsg_virtio_send_offchannel_raw
	-> rpmsg_virtio_get_tx_payload_buffer
		 -> rpmsg_virtio_get_tx_buffer
			-> virtqueue_get_available_buffer
				-> metal_cache_invalidate(VRING_AVAIL(vq),  vring_avail_size(VRING_NUM(vq), VRING_ALIGN(vq)));
				-> metal_cache_invalidate(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));

	-> rpmsg_virtio_send_offchannel_nocopy
		-> metal_cache_flush(metal_io_virt(io, offset), len);
			-> rpmsg_virtio_enqueue_buffer
				-> virtqueue_add_consumed_buffer
					-> metal_cache_invalidate(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
					-> metal_cache_flush(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));

```				
```
rpmsg_virtio_rx_callback
	-> rpmsg_virtio_get_rx_buffer
		-> virtqueue_get_available_buffer
				-> metal_cache_invalidate(VRING_AVAIL(vq),  vring_avail_size(VRING_NUM(vq), VRING_ALIGN(vq)));
				-> metal_cache_invalidate(VRING_DESC(vq), vring_desc_size(VRING_NUM(vq)));
		-> metal_cache_invalidate(data, *len);
	-> rpmsg_virtio_return_buffer
		-> virtqueue_add_consumed_buffer
				-> metal_cache_invalidate(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
				->  metal_cache_flush(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));

```
based on  that, i proposed some optimisations in the code.

Then as you mentioned perpahs sometime when only one variable is updated or used, it can be more efficient to not update the whole structure.

arnopo:
One more point that was just discussed in the OpenAMp be-weekly. It would be nice to differentiate the cache management of the vrings and buffers for future needs such as support of new virtio services.

That would means to splitt VIRTIO_D_CACHE config.
could be something like:
- VIRTIO_CACHED_VRINGS
- VIRTIO_CACHED_BUFFERS

Don't hesitate if you have more inspiration for the naming

arnopo:
One more point that was just discussed in the OpenAMp be-weekly. It would be nice to differentiate the cache management of the vrings and buffers for future needs such as support of new virtio services.

That would means to splitt VIRTIO_D_CACHE config.
could be something like:
- VIRTIO_CACHED_VRINGS
- VIRTIO_CACHED_BUFFERS

Don't hesitate if you have more inspiration for the naming

lib/virtio/virtqueue.c
```diff
@@ -469,6 +544,10 @@ static void vq_ring_free_chain(struct virtqueue *vq, uint16_t desc_idx)
 	struct vring_desc *dp;
 	struct vq_desc_extra *dxp;
 
+#ifdef VIRTIO_D_CACHE
```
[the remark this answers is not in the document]
carlocaione: ACK

lib/virtio/virtqueue.c
```diff
@@ -279,6 +312,10 @@ int virtqueue_add_consumed_buffer(struct virtqueue *vq, uint16_t head_idx,
 
 	VQUEUE_BUSY(vq);
 
+#ifdef VIRTIO_D_CACHE
```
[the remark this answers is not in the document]
carlocaione: ACK

lib/virtio/virtqueue.c
```diff
@@ -88,6 +96,16 @@ int virtqueue_create(struct virtio_device *virt_dev, unsigned short id,
...
 	}
 
+#ifdef VIRTIO_D_CACHE
+#ifndef VIRTIO_SLAVE_ONLY
+	if (vq->vq_dev->role == VIRTIO_DEV_MASTER) {
+		/* Flush the vring control block */
+		metal_cache_flush(VRING_DESC(vq),
+				  vring_size(VRING_NUM(vq), VRING_ALIGN(vq)));
+	}
+#endif /* !VIRTIO_SLAVE_ONLY */
```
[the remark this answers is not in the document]
carlocaione: ACK

lib/virtio/virtqueue.c
```diff
@@ -586,6 +681,12 @@ static int vq_ring_enable_interrupt(struct virtqueue *vq, uint16_t ndesc)
 #endif /*VIRTIO_MASTER_ONLY*/
 	}
 
+#ifdef VIRTIO_D_CACHE
+	metal_cache_flush(VRING_USED(vq), vring_used_size(VRING_NUM(vq)));
```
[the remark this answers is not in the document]
carlocaione: `vq->vq_ring.used->flags` could be updated here by slave and it could be read by master in `vq_ring_must_notify()`
`vq->vq_ring.avail->flags` could be updated here by master and it could be read by slave in `vq_ring_must_notify()`

lib/virtio/virtqueue.c
carlocaione: We need to invalidate (no flush right?) also for master because used could have been modified by slave in `vq_ring_enable_interrupt()`

lib/virtio/virtqueue.c
carlocaione: assuming you mean invalidate avail could have been modified by master in `vq_ring_enable_interrupt()` so we need to invalidate this also for slave

lib/virtio/virtqueue.c
carlocaione: yeah, I'll revise everything and trying to flush / invalidate as little as possible again

lib/virtio/virtqueue.c
carlocaione: ACK

lib/virtio/virtqueue.c
arnopo: Sorry i was spoken about invalidate yes.
vq_ring_must_notify is called by virtqueue_kick at the end of a message sent or received, so caches has already been invalidated, i wonder if we could supress oinvalidate

lib/virtio/virtqueue.c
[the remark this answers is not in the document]
arnopo: Seems that I have  some comments are not at the good place...need to resetrt the review

lib/virtio/virtqueue.c
```diff
@@ -586,6 +681,12 @@ static int vq_ring_enable_interrupt(struct virtqueue *vq, uint16_t ndesc)
...
+	metal_cache_flush(VRING_AVAIL(vq),
```
arnopo, line 686: only master should fetch the VRING_AVAIL and only slave should flush the VRING_USED, else you ca create a memory corruption.
same issue for the virtqueue_disable_cb function

lib/virtio/virtqueue.c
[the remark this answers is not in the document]
carlocaione: Wait for my next revision before restarting the review, I'm doing a bit of cleanup.

arnopo (changes requested):

lib/virtio/virtqueue.c
carlocaione: ACK

carlocaione:
I liked your idea, so I created this for this new revision ;)
```
MASTER
======
-> rpmsg_init_vdev
  -> rpmsg_virtio_create_virtqueues
    -> virtio_create_virtqueues
      -> virtqueue_create
        -> vq_ring_init [W desc (.next)]
          -> vring_init
  -> virtqueue_add_buffer
    -> vq_ring_add_buffer [R desc, W desc (.addr, .len, .flags)]
  -> vq_ring_update_avail [R avail (.idx), W avail (.idx, .ring))]

-> rpmsg_virtio_send_offchannel_raw
  -> rpmsg_virtio_get_tx_payload_buffer
    -> rpmsg_virtio_get_tx_buffer
      -> virtqueue_get_buffer [R used (.idx, .ring)]
        -> vq_ring_free_chain [R desc (.flags) , W desc (.next)]
  -> rpmsg_virtio_send_offchannel_nocopy
    -> virtqueue_get_buffer_length [R desc (.len)]
    -> rpmsg_virtio_enqueue_buffer
      -> virtqueue_add_buffer
        -> vq_ring_add_buffer [R desc, W desc (.addr, .len, .flags)]
      -> vq_ring_update_avail [R avail (.idx), W avail (.idx, .ring)]

-> rpmsg_virtio_rx_callback
  -> rpmsg_virtio_get_rx_buffer
    -> virtqueue_get_buffer [R used (.idx, .ring)]
      -> vq_ring_free_chain [R desc (.flags), W desc (.next)]
  -> rpmsg_virtio_return_buffer
    -> virtqueue_add_buffer
      -> vq_ring_add_buffer [R desc, W desc (.addr, .len, .flags)]
    -> vq_ring_update_avail [R avail (.idx), W avail (.idx, .ring)]
  -> rpmsg_virtio_get_rx_buffer
    -> virtqueue_get_buffer [R used (.idx, .ring)]
      -> vq_ring_free_chain [R desc (.flags), W desc (.next)]
  -> virtqueue_kick
    -> vq_ring_must_notify [R avail (.idx), R used (.flags)]

-> virtqueue_disable_cb [W avail (.flags, .ring)]

-> virtqueue_enable_cb
  -> vq_ring_enable_interrupt [W avail (.flags, .ring), R used (.idx)]

-> virtqueue_get_desc_size [R avail (.idx, .ring), R desc (.len)]

-> virtqueue_dump [R avail (.idx, .flags), R used (.idx, .flags)]

SLAVE
=====
-> rpmsg_init_vdev
  -> rpmsg_virtio_create_virtqueues
    -> virtio_create_virtqueues
      -> virtqueue_create
        -> vq_ring_init
          -> vring_init

-> rpmsg_virtio_send_offchannel_raw
  -> rpmsg_virtio_get_tx_payload_buffer
    -> rpmsg_virtio_get_tx_buffer
      -> virtqueue_get_available_buffer [R avail (.idx, .ring), R desc (.addr, .len)]
  -> rpmsg_virtio_send_offchannel_nocopy
    -> virtqueue_get_buffer_length [R desc, (.len)]
      -> rpmsg_virtio_enqueue_buffer
        -> virtqueue_add_consumed_buffer [R used (.idx, .ring), W used (.idx)]

-> rpmsg_virtio_rx_callback
  -> rpmsg_virtio_get_rx_buffer
    -> virtqueue_get_available_buffer [R avail, (.idx, .ring), R desc (.addr, .len)]
  -> rpmsg_virtio_return_buffer
    -> virtqueue_add_consumed_buffer [R used (.idx, .ring), W used (.idx)]
  -> rpmsg_virtio_get_rx_buffer
    -> virtqueue_get_available_buffer [R avail, (.idx, .ring), R desc (.addr, .len)]
  -> virtqueue_kick
    -> vq_ring_must_notify [R used (.idx), R avail (.flags)]

-> virtqueue_disable_cb [W used (.flags, .ring)]

-> virtqueue_enable_cb
  -> vq_ring_enable_interrupt [W used (.flags, .ring), R used (.idx)]

-> virtqueue_get_desc_size [R avail (.idx, .ring), R desc (.len)]

-> virtqueue_dump [R avail (.idx, .flags), R used (.idx, .flags)]

============================================================================================

MASTER
======
R desc (.flags, .len)
W desc (.addr, .next, .len, .flags)
R avail (.idx)
W avail (.idx, .ring, .flags)
R used (.idx, .ring, .flags)
W used -

SLAVE
=====
R desc (.addr, .len)
W desc -
R avail (.idx, .ring, .flags)
W avail -
R used (.idx, .ring)
W used (.idx, .ring, .flags)
```

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -369,6 +378,11 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 				      &rp_hdr, sizeof(rp_hdr));
 	RPMSG_ASSERT(status == sizeof(rp_hdr), "failed to write header\r\n");
 
+#ifdef VIRTIO_CACHED_BUFFERS
+	offset = metal_io_virt_to_offset(io, (void *) data);
```
arnopo, line 382: coding rule: No space is necessary after a cast  (void *)data

lib/virtio/virtqueue.c
```diff
@@ -469,6 +564,10 @@ static void vq_ring_free_chain(struct virtqueue *vq, uint16_t desc_idx)
 	struct vring_desc *dp;
 	struct vq_desc_extra *dxp;
 
+	/*
+	 * CACHE: desc is never written by slave, no need to invalidate
```
arnopo, line 568: comment on one line

lib/virtio/virtqueue.c
```diff
@@ -279,6 +315,11 @@ int virtqueue_add_consumed_buffer(struct virtqueue *vq, uint16_t head_idx,
 
 	VQUEUE_BUSY(vq);
 
+	/*
+	 * CACHE: used is never written by master, so it's safe to directly
+	 * access it
+	 */
+
```
arnopo, line 322: suppress useless blank line

arnopo (changes requested):
Look to me a good compromize as i have not better solution to propose.  I added few minor comments on typo and coding rules.
Need also @edmooring  approve and give time others to review this PR.

Thanks
Arnaud

lib/virtio/virtqueue.c
```diff
@@ -440,6 +521,11 @@ static uint16_t vq_ring_add_buffer(struct virtqueue *vq,
 		VQASSERT(vq, idx != VQ_RING_DESC_CHAIN_END,
 			 "premature end of free desc chain");
 
+		/*
+		 * CACHE: No need to invalidate desc because it is only written
+		 * by master
```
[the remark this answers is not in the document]
carlocaione: ack

lib/virtio/virtqueue.c
```diff
@@ -541,14 +642,27 @@ static void vq_ring_update_avail(struct virtqueue *vq, uint16_t desc_idx)
 	 * deferring to virtqueue_notify() in the hopes that if the host is
 	 * currently running on another CPU, we can keep it processing the new
 	 * descriptor.
+	 *
+	 * CACHE: avail is never written by slave, so it is safe to not
+	 * invalidate here
```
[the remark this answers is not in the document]
carlocaione: ack

lib/virtio/virtqueue.c
```diff
@@ -184,6 +198,11 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 
 	atomic_thread_fence(memory_order_seq_cst);
 
+#ifdef VIRTIO_CACHED_VRINGS
+	/* used.ring is written by slave, invalidate it */
```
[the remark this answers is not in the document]
carlocaione: ack

lib/virtio/virtqueue.c
```diff
@@ -174,6 +183,11 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx)
 	void *cookie;
 	uint16_t used_idx, desc_idx;
 
+#ifdef VIRTIO_CACHED_VRINGS
+	/* used.idx is update by slave, so we need to invalidate */
+	VRING_INVALIDATE(vq->vq_ring.used->idx);
+#endif
```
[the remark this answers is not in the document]
carlocaione: ack

lib/rpmsg/rpmsg_virtio.c
carlocaione: ack

lib/virtio/virtqueue.c
carlocaione: ack

lib/virtio/virtqueue.c
carlocaione: ack

arnopo (approved):
LGTM

edmooring (approved):
This looks good to me. It is very clear what the intent of the code is.

A future visual optimization might be to put the '#ifdef VIRTIO_CACHED_VRINGS" inside the implementation of VRING_FLUSH and VRING_INVALIDATE to reduce visual clutter in the code. On the other hand, that might be a little harder to understand at first.

Also, we need to take the documentation of the call chains, and which vring elements are written and read by each side, and put it into the OpenAMP repo somewhere. It's too valuable to just leave in a pull request. Very nice work @arnopo and @carlocaione.

arnopo:
Hi @edmooring
very good remark concerning visual optimization!

@carlocaione: May I ask you to make one last effort to do it?

lib/virtio/virtqueue.c
```diff
@@ -28,6 +29,14 @@ static int virtqueue_nused(struct virtqueue *vq);
 static int virtqueue_navail(struct virtqueue *vq);
 #endif
 
+#ifdef VIRTIO_CACHED_VRINGS
+#define VRING_FLUSH(x)		metal_cache_flush(&x, sizeof(x))
+#define VRING_INVALIDATE(x)	metal_cache_invalidate(&x, sizeof(x))
+#else
+#define VRING_FLUSH(x)		do { } while (0)
+#define VRING_INVALIDATE(x)	do { } while (0)
```
arnopo, line 37: Thanks for the update!
I wondered if the  "do { } while (0)"  add useless assembler code, but seems not the case with GCC, and seems better to implement it in this way to avoid warnings with some old compilers.  I will cross check the point for ARMCC and EWARM compiler for the next release and fix it if needed. 
If no more remarks we will probably merge it next week

carlocaione: thanks!

carlocaione:
@arnopo found an issue. Converting this to draft in case you wanted to merge it. Fixing coming soon. Sorry about this.

carlocaione:
This is the diff with the previous version:
```
diff --git a/lib/rpmsg/rpmsg_virtio.c b/lib/rpmsg/rpmsg_virtio.c
index 17715a8..f15f45d 100644
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -354,9 +354,6 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
        uint32_t buff_len;
        uint16_t idx;
        int status;
-#ifdef VIRTIO_CACHED_BUFFERS
-       unsigned long offset;
-#endif /* VIRTIO_CACHED_BUFFERS */
 
@@ -379,8 +376,7 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
        RPMSG_ASSERT(status == sizeof(rp_hdr), "failed to write header\r\n");
 
 #ifdef VIRTIO_CACHED_BUFFERS
-       offset = metal_io_virt_to_offset(io, (void *)data);
-       metal_cache_flush(metal_io_virt(io, offset), len);
+       metal_cache_flush(data, len);
 #endif /* VIRTIO_CACHED_BUFFERS */
 
@@ -559,6 +555,10 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
        if (len != sizeof(*ns_msg))
                /* Returns as the message is corrupted */
                return RPMSG_SUCCESS;
+
+#ifdef VIRTIO_CACHED_BUFFERS
+       metal_cache_invalidate(data, len);
+#endif /* VIRTIO_CACHED_BUFFERS */
        metal_io_block_read(io,
                            metal_io_virt_to_offset(io, ns_msg->name),
                            &name, sizeof(name));
diff --git a/lib/virtio/virtqueue.c b/lib/virtio/virtqueue.c
index 05952a6..889f5b6 100644
--- a/lib/virtio/virtqueue.c
+++ b/lib/virtio/virtqueue.c
@@ -314,6 +314,9 @@ int virtqueue_add_consumed_buffer(struct virtqueue *vq, uint16_t head_idx,
        used_desc->id = head_idx;
        used_desc->len = len;
 
+       /* We still need to flush it because this is read by master */
+       VRING_FLUSH(vq->vq_ring.used->ring[used_idx]);
+
 
        vq->vq_ring.used->idx++;

```
It's especially important that I missed a `VRING_FLUSH`.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -545,6 +555,10 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	if (len != sizeof(*ns_msg))
 		/* Returns as the message is corrupted */
 		return RPMSG_SUCCESS;
+
+#ifdef VIRTIO_CACHED_BUFFERS
```
arnopo, line 559: Seem not needed, already done in rpmsg_virtio_get_rx_buffer, function is an endpoint callback

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -369,6 +375,10 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 				      &rp_hdr, sizeof(rp_hdr));
 	RPMSG_ASSERT(status == sizeof(rp_hdr), "failed to write header\r\n");
 
+#ifdef VIRTIO_CACHED_BUFFERS
+	metal_cache_flush((void *)data, len);
```
arnopo, line 379: size to flush is len ( payload) + sizeof(hdr)
But it would be more symmetrical to do this in rpmsg_virtio_enqueue_buffer.

arnopo (changes requested):

lib/rpmsg/rpmsg_virtio.c
carlocaione: ACK

lib/rpmsg/rpmsg_virtio.c
carlocaione: yeah, that seems better

carlocaione:
@arnopo FWIW this code has been now tested on real hardware with IPC behind cache and it seems to work correctly.

arnopo (approved):
As few update i let @edmooring make another pass before merging

edmooring (approved):
Looks good to me now.

PR #313, opened 2021-11-09
rproc: zero len arrays cause invalid pointer access.

thesamprice:
When trying to setup a remote proc to load an elf file.
I was getting segfaults from having non initialized / empty rproc->mems.
Checking for null here would fix this issue
Unsure if this is the best place to put this patch or if a better place would be in libmetal, and have the list_for_each not return / enter the loop if the node was null.

### DIFF
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -46,6 +46,9 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 		return NULL;
 
 	metal_list_for_each(&rproc->mems, node) {
+		if(node == 0x0){
+			return NULL;
+		}
 		mem = metal_container_of(node, struct remoteproc_mem, node);
 		if (name) {
 			if (!strncmp(name, mem->name, strlen(name)))


lib/remoteproc/remoteproc.c
```diff
@@ -46,6 +46,9 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 		return NULL;
 
 	metal_list_for_each(&rproc->mems, node) {
+		if(node == 0x0){
+			return NULL;
+		}
```
arnopo, line 51: To reproduce the issue seems that rpoc->mems should not initialized. 
I suppose that this issue occurs when rproc structure is not initialized (`remoteproc_init` not called), right?

So here is look like a coding bug, which could be solved by an assert outside the loop.
`metal_assert(!rproc->mems.prev || !rproc->mems.next)`

thesamprice: You could also consider modifying the libmetal code, so null pointers / empty lists dont run.
```c
#define metal_list_for_each(list, node)		\
	for ((node) = (list)->next;		\
	     (node) != (list);			\
	     (node) = (node)->next)
	     ```
to 
```c
#define metal_list_for_each(list, node)		\
	for ((node) = (list)->next;		\
	     ((list) != NULL) && ((node) != NULL) && (node) != (list);			\
	     (node) = (node)->next)
	     ```
Im not really familiar enough with libmetal / open-amp to judge the correct approach.
Was taking a stab at making a stand alone elf loader that used the linux devmem interface to get memory from elf file, and got a segfault from not having my remoteproc initialized.
So I most likely was not calling remoteproc_init.

thesamprice:
Closing , after calling remoteproc_init this worked.

Issue #314, opened 2021-11-11
About notifyid in "handle_vdev_rsc" function


lozdznil:
Hi,

I'm a little confused about the following code in `handle_vdev_rsc` function in `rsc_table_parser.c` (lib/remoteproc/):

```c
for (i = 0; i < num_vrings; i++) {
	struct fw_rsc_vdev_vring *vring_rsc;

	vring_rsc = &vdev_rsc->vring[i];
	notifyid = vring_rsc->notifyid;
	notifyid = remoteproc_allocate_id(rproc,
					  notifyid,
					  notifyid + 1);
	if (notifyid != RSC_NOTIFY_ID_ANY)
		vdev_rsc->notifyid = notifyid;
}
```

Here it uses `vring_rsc->notifyid` to allocate new IDs, but finally updates `vdev_rsc->notifyid` rather than `vring_rsc->notifyid`.

Is this a typo? My understanding is that it should be like this:

```diff
for (i = 0; i < num_vrings; i++) {
	struct fw_rsc_vdev_vring *vring_rsc;

	vring_rsc = &vdev_rsc->vring[i];
	notifyid = vring_rsc->notifyid;
	notifyid = remoteproc_allocate_id(rproc,
					  notifyid,
					  notifyid + 1);
	if (notifyid != RSC_NOTIFY_ID_ANY)
-		vdev_rsc->notifyid = notifyid;
+		vring_rsc->notifyid = notifyid;
}
```

Thanks!

arnopo:
Hello @lozdznil 

Thanks for poiting out this!
Yes this seems not correct. Could you send a PR to propose a fix?

PR #316, opened 2021-11-16
remoteproc: fix notifyid assignment in handle_vdev_rsc

lozdznil:
Fix the issue that it used vring_rsc->notifyid to allocate new IDs but
finally updated vdev_rsc->notifyid rather than vring_rsc->notifyid.

(See https://github.com/OpenAMP/open-amp/issues/314)

Signed-off-by: Junyan Lin <junyanlin@allwinnertech.com>

### DIFF
--- a/lib/remoteproc/rsc_table_parser.c
+++ b/lib/remoteproc/rsc_table_parser.c
@@ -148,7 +148,7 @@ int handle_vdev_rsc(struct remoteproc *rproc, void *rsc)
 						  notifyid,
 						  notifyid + 1);
 		if (notifyid != RSC_NOTIFY_ID_ANY)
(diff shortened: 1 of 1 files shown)


arnopo:
Hi @edmooring  @xiaoxiang781216,

Please, Could you test this fix in you environment ( i.e. Xilinx and NuttX)?
I wonder why the problem did not appear before...

Thanks in advance
Arnaud

edmooring:
Hi @arnopo,
It works for me in my emulated Xilinx environments. Like you, I wonder why it worked before. Perhaps our tests are not strong enough.

arnopo (approved):

edmooring (approved):
Looks good to me.

PR #319, opened 2021-11-23
rpmsg: Make RPMSG_BUFFER_SIZE a fixed define

supergaute:
Since the rpmsg_virtio.h file is exported, it was possible to have
different values for RPMSG_BUFFER_SIZE when open-amp is built and
when the library is included in an application.
By making this parameter a CMake option and hard-coding the value
in rpmsg_virtio.h, it will make sure this parameter cannot differ
between building of open-amp and the user application.

Signed-off-by: Gaute Nilsson <gaute.nilsson@siemens-energy.com>

### DIFF
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -22,9 +22,7 @@ extern "C" {
 #endif
 
 /* Configurable parameters */
-#ifndef RPMSG_BUFFER_SIZE
-#define RPMSG_BUFFER_SIZE	(512)
-#endif
+#define RPMSG_BUFFER_SIZE	@RPMSG_BUFFER_SIZE@
 
 /* The feature bitmap for virtio rpmsg */
 #define VIRTIO_RPMSG_F_NS	0 /* RP supports name service notifications */
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -101,9 +101,7 @@ endif (WITH_ZEPHYR)
 
 option (WITH_LIBMETAL_FIND "Check Libmetal library can be found" ON)
 
-if (DEFINED RPMSG_BUFFER_SIZE)
-  add_definitions( -DRPMSG_BUFFER_SIZE=${RPMSG_BUFFER_SIZE} )
-endif (DEFINED RPMSG_BUFFER_SIZE)
+set (RPMSG_BUFFER_SIZE 512 CACHE STRING "Adjust the size of the RPMsg buffers")
 
 message ("-- C_FLAGS : ${CMAKE_C_FLAGS}")
 # vim: expandtab:ts=2:sw=2:smartindent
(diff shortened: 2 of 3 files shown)


lib/include/openamp/rpmsg_virtio.h
```diff
@@ -22,9 +22,7 @@ extern "C" {
 #endif
 
 /* Configurable parameters */
-#ifndef RPMSG_BUFFER_SIZE
-#define RPMSG_BUFFER_SIZE	(512)
-#endif
+#define RPMSG_BUFFER_SIZE	@RPMSG_BUFFER_SIZE@
```
arnopo, line 25: Today the open-amp can be used without the cmakeand a library build, but with files directly build in the application project. 
This break the legacy , imposing the cmake usage (only optional version_def.h is handled by cmake).

And this would not guarantee that include files would be aligned  with a dynamic library. 

On the other side if the application projetc define the RPMSG_BUFFER_SIZE and open-amp library build as a sub project could this solve the issue?

I also wonder if a new APi to get the rpmsg _size such as the introduce in [Linux kernel ]()https://elixir.bootlin.com/linux/v5.16-rc2/source/include/linux/rpmsg.h#L189) would be more reliable.
The drawback would be for static allocation.

edmooring: I think making things more dependent on cmake is a possible problem. Also, how many people ever build with RPMSG_BUFFER_SIZE anything but 512?

supergaute: Well, I can only speak for myself, but I use OpenAMP with FreeRTOS, build with CMake and RPMSG_BUFFER_SIZE set to 4096.
What is so special about using RPMSG_BUFFER_SIZE to anything other than 512?

arnopo: From MPOV, it is also a valid use case to decrease the buffer size for MCU to MCU communications, to limit the shared memory.

I think we also have to think about evolution such as  https://github.com/OpenAMP/open-amp/pull/155 pull request is also a possible evolution that would make the size negociated.
That's why a new API could be nice to get the buffer effective size.
But as mentionned before this is not compatible with a static allocation. that said, that could allow to check the alignement between the library and the application...

edmooring: > Well, I can only speak for myself, but I use OpenAMP with FreeRTOS, build with CMake and RPMSG_BUFFER_SIZE set to 2048.
> What is so special about using RPMSG_BUFFER_SIZE to anything other than 512?

The only thing special about 512 is that Linux remoteproc is hard-coded to 512. I hadn't thought of the MCU->MCU case or other such configurations.

arnopo: @supergaute 

Could you confirm that https://github.com/OpenAMP/open-amp/pull/328 answer to your need and that we can close this one?

arnopo: Without any answer, I consider that the https://github.com/OpenAMP/open-amp/pull/328 answer to your need, please reopen this pull request if it is not the case

PR #320, opened 2021-11-25
openamp: fix scenario case

GUIDINGLI:
description:
There are two CPUs use IPC.
Each cpu has one thread to handle RX callback.
When meet scenario case, both CPU0 & CPU1 can't get tx buffer.
IPC can't do communication any more.

rootcase:
CPU0                               CPU1
TX thread0 send msg         -->
RX thread get ack           <--  RX thread get msg, send ack
                            <--  TX thread0 send msg
RX thread get msg, send ack -->  RX thread get ack

TX thread1 send msg         -->
RX thread get ack           <--  RX thread get msg, send ack
                            <--  TX thread1 send msg
RX thread get msg, send ack -->  RX thread get ack

TX thread2 send msg         -->
TX thread3 send msg         -->
....
                            <--  TX thread2 send msg
                            <--  TX thread3 send msg
                                 ...

when CPU0 & CPU1 msg send too quickly, then CPU0 RX thread can't
get tx buffer(ack), wait CPU1 RX thread return buffer. But at this time,
CPU1 RX thread also can't get tx buffer(ack), wait CPU0 RX thread
return buffer. So, you will find two cpu locked.

last call stack:

CPU0 RX thread:                     CPU1 RX thread:
wait tx buffer                      wait tx buffer
can't handle rx buffer              can't handle rx buffer

Resolve:
add recursive call rpmsg_virtio_rx_callback() when get_tx_buffer() failed

### DIFF
--- a/lib/include/openamp/remoteproc.h
+++ b/lib/include/openamp/remoteproc.h
@@ -432,6 +432,16 @@ struct remoteproc_ops {
 					  metal_phys_addr_t da,
 					  void *va, size_t size,
 					  struct remoteproc_mem *buf);
+	/**
+	 * wait_tx_buffer
+	 *
+	 * Wait tx buffer available
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 *
+	 * return none
+	 */
+	int (*wait_tx_buffer)(struct remoteproc *rproc);
 };
 
 /* Remoteproc error codes */
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -109,6 +109,12 @@ rpmsg_virtio_create_virtqueues(struct rpmsg_virtio_device *rvdev,
 					callbacks);
 }
 
+static inline int rpmsg_virtio_wait_tx_buffer(struct rpmsg_virtio_device *rvdev)
+{
+	return rvdev->vdev->func->wait_tx_buffer ?
+			rvdev->vdev->func->wait_tx_buffer(rvdev->vdev) : -EAGAIN;
+}
+
 /**
  * rpmsg_virtio_get_buffer_size - get rpmsg virtio buffer size
  *
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -304,6 +304,72 @@ static void rpmsg_virtio_release_rx_buffer(struct rpmsg_device *rdev,
 	metal_mutex_release(&rdev->lock);
 }
 
+/**
+ * rpmsg_virtio_rx_callback
+ *
+ * Rx callback function.
+ *
+ * @param vq - pointer to virtqueue on which messages is received
+ *
+ */
+static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
+{
+	struct virtio_device *vdev = vq->vq_dev;
+	struct rpmsg_virtio_device *rvdev = vdev->priv;
+	struct rpmsg_device *rdev = &rvdev->rdev;
+	struct rpmsg_endpoint *ept;
+	struct rpmsg_hdr *rp_hdr;
+	uint32_t len;
+	uint16_t idx;
+	int status;
+
+	metal_mutex_acquire(&rdev->lock);
+
+	/* Process the received data from remote node */
+	rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
+
+	metal_mutex_release(&rdev->lock);
+
+	while (rp_hdr) {
+		rp_hdr->reserved = idx;
+
+		/* Get the channel node from the remote device channels list. */
+		metal_mutex_acquire(&rdev->lock);
+		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
+		metal_mutex_release(&rdev->lock);
+
+		if (ept) {
+			if (ept->dest_addr == RPMSG_ADDR_ANY) {
+				/*
+				 * First message received from the remote side,
+				 * update channel destination address
+				 */
+				ept->dest_addr = rp_hdr->src;
+			}
+			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
+					 rp_hdr->len, rp_hdr->src, ept->priv);
+
+			RPMSG_ASSERT(status >= 0,
+				     "unexpected callback status\r\n");
+		}
+
+		metal_mutex_acquire(&rdev->lock);
+
+		/* Check whether callback wants to hold buffer */
+		if (!(rp_hdr->reserved & RPMSG_BUF_HELD)) {
+			/* No, return used buffers. */
+			rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
+		}
+
+		rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
+		if (!rp_hdr) {
+			/* tell peer we return some rx buffer */
+			virtqueue_kick(rvdev->rvq);
+		}
+		metal_mutex_release(&rdev->lock);
+	}
+}
+
 static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 						uint32_t *len, int wait)
 {
@@ -333,6 +399,8 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		metal_mutex_release(&rdev->lock);
 		if (rp_hdr || !tick_count)
 			break;
+		if (rpmsg_virtio_wait_tx_buffer(rvdev) >= 0)
+			continue;
 		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
 		tick_count--;
 	}
@@ -456,72 +524,6 @@ static int rpmsg_virtio_send_offchannel_raw(struct rpmsg_device *rdev,
 static void rpmsg_virtio_tx_callback(struct virtqueue *vq)
 {
 	(void)vq;
-}
-
-/**
- * rpmsg_virtio_rx_callback
- *
- * Rx callback function.
- *
- * @param vq - pointer to virtqueue on which messages is received
- *
- */
-static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
-{
-	struct virtio_device *vdev = vq->vq_dev;
-	struct rpmsg_virtio_device *rvdev = vdev->priv;
-	struct rpmsg_device *rdev = &rvdev->rdev;
-	struct rpmsg_endpoint *ept;
-	struct rpmsg_hdr *rp_hdr;
-	uint32_t len;
-	uint16_t idx;
-	int status;
-
-	metal_mutex_acquire(&rdev->lock);
-
-	/* Process the received data from remote node */
-	rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
-
-	metal_mutex_release(&rdev->lock);
-
-	while (rp_hdr) {
-		rp_hdr->reserved = idx;
-
-		/* Get the channel node from the remote device channels list. */
-		metal_mutex_acquire(&rdev->lock);
-		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
-		metal_mutex_release(&rdev->lock);
-
-		if (ept) {
-			if (ept->dest_addr == RPMSG_ADDR_ANY) {
-				/*
-				 * First message received from the remote side,
-				 * update channel destination address
-				 */
-				ept->dest_addr = rp_hdr->src;
-			}
-			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
-					 rp_hdr->len, rp_hdr->src, ept->priv);
-
-			RPMSG_ASSERT(status >= 0,
-				     "unexpected callback status\r\n");
-		}
-
-		metal_mutex_acquire(&rdev->lock);
-
-		/* Check whether callback wants to hold buffer */
-		if (!(rp_hdr->reserved & RPMSG_BUF_HELD)) {
-			/* No, return used buffers. */
-			rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
-		}
-
-		rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
-		if (!rp_hdr) {
-			/* tell peer we return some rx buffer */
-			virtqueue_kick(rvdev->rvq);
-		}
-		metal_mutex_release(&rdev->lock);
-	}
 }
 
 /**
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -137,6 +137,7 @@ struct virtio_dispatch {
 			     void *src, int length);
 	void (*reset_device)(struct virtio_device *dev);
 	void (*notify)(struct virtqueue *vq);
+	int (*wait_tx_buffer)(struct virtio_device *dev);
 };
 
 int virtio_create_virtqueues(struct virtio_device *vdev, unsigned int flags,
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -30,6 +30,17 @@ static void rproc_virtio_virtqueue_notify(struct virtqueue *vq)
 	rpvdev->notify(rpvdev->priv, vring_info->notifyid);
 }
 
+static int rproc_virtio_wait_tx_buffer(struct virtio_device *vdev)
+{
+	struct remoteproc_virtio *rpvdev;
+	struct remoteproc *rproc;
+
+	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
+	rproc  = rpvdev->priv;
+
+	return rproc->ops->wait_tx_buffer ? rproc->ops->wait_tx_buffer(rproc) : -EAGAIN;
+}
+
 static unsigned char rproc_virtio_get_status(struct virtio_device *vdev)
 {
 	struct remoteproc_virtio *rpvdev;
@@ -179,6 +190,7 @@ static const struct virtio_dispatch remoteproc_virtio_dispatch_funcs = {
 	.get_features = rproc_virtio_get_features,
 	.read_config = rproc_virtio_read_config,
 	.notify = rproc_virtio_virtqueue_notify,
+	.wait_tx_buffer = rproc_virtio_wait_tx_buffer,
 #ifndef VIRTIO_SLAVE_ONLY
 	/*
 	 * We suppose here that the vdev is in a shared memory so that can


arnopo:
Hello @GUIDINGLI

If i well understood your application is blocked in `rpmsg_virtio_get_tx_payload_buffer` waiting that a TX buffer is available.
In such case it does not process the incoming RPMsg and after a while the remote processor is also blocked waiting available buffer.

The patch you propose probably works for your senario, but look to me strange to use the rpmsg_send bocking interface. If you issue is that you are blocked during send without possibility to treat incoming message.

I would see 2 other alternatives that could solve your issue.
- use non blocking APi instead `rpmsg_trysend` instead
- handle new RX message notification in a seperate RX thread to manage remoteproc_get_notification mechanism ( releying on IPC signaling)  (see zephyr as example [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/ipc/rpmsg_service/rpmsg_backend.c#L150)

xiaoxiang781216:
> Hello @GUIDINGLI
> 
> If i well understood your application is blocked in `rpmsg_virtio_get_tx_payload_buffer` waiting that a TX buffer is available. In such case it does not process the incoming RPMsg and after a while the remote processor is also blocked waiting available buffer.
> 
> The patch you propose probably works for your senario, but look to me strange to use the rpmsg_send bocking interface. If you issue is that you are blocked during send without possibility to treat incoming message.
> 
> I would see 2 other alternatives that could solve your issue.
> 
> * use non blocking APi instead `rpmsg_trysend` instead

But, how to handle rpmsg_trysend fail?

> * handle new RX message notification in a seperate RX thread to manage remoteproc_get_notification mechanism

We already have a dedicated thread to handle the notification:
https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L174-L206

> ( releying on IPC signaling)  (see zephyr as example [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/ipc/rpmsg_service/rpmsg_backend.c#L150)

Zephyr implementation also suffer the same issue too: rpmsg_endpoint::cb may wait in rpmsg_send when the virtio buffer exhaust which then block the work thread to process the next incoming request.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -333,6 +399,8 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		metal_mutex_release(&rdev->lock);
 		if (rp_hdr || !tick_count)
 			break;
+		if (rproc_virtio_can_recursive(rvdev->vdev))
```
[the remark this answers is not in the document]
xiaoxiang781216: Look like we need put the callback to virtio_dispatch.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -109,6 +109,11 @@ rpmsg_virtio_create_virtqueues(struct rpmsg_virtio_device *rvdev,
 					callbacks);
 }
 
+static inline int rpmsg_virtio_can_recursive(struct rpmsg_virtio_device *rvdev)
+{
+	return rvdev->vdev->func->can_recursive(rvdev->vdev);
```
xiaoxiang781216, line 114: check can_recursive isn't null before invoke

lib/include/openamp/remoteproc.h
```diff
@@ -432,6 +432,16 @@ struct remoteproc_ops {
...
+	 * can_recursive
+	 *
+	 * Check current vdev can do recursive rx_callabck or not
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 *
+	 * return true if can do recursive rx_callabck.
+	 */
+	int (*can_recursive)(struct remoteproc *rproc);
 };
```
edmooring, line 445: I don't like this name. It doesn't immediately tell the reader what is going on. Something like rx_callback_can_recurse() is more explanatory, but still not good.

edmooring (changes requested):
The commit message needs a better title. "fix scenario case" is not going to tell people what they need to know when viewing a shortlog.

This is also the wrong place to handle such a problem. It forces all applications to handle this condition in the same way. Other applications might want to treat this kind of deadlock as a serious error, for example. 

This kind of error should be handled in a higher-level protocol within the application, rather than as a special case within the virtio code.

arnopo:
> > Hello @GUIDINGLI
> > If i well understood your application is blocked in `rpmsg_virtio_get_tx_payload_buffer` waiting that a TX buffer is available. In such case it does not process the incoming RPMsg and after a while the remote processor is also blocked waiting available buffer.
> > The patch you propose probably works for your senario, but look to me strange to use the rpmsg_send bocking interface. If you issue is that you are blocked during send without possibility to treat incoming message.
> > I would see 2 other alternatives that could solve your issue.
> > 
> > * use non blocking APi instead `rpmsg_trysend` instead
> 
> But, how to handle rpmsg_trysend fail? it isn't a solution but push the problem to user.

RPMSG_ERR_NO_BUFF error should be used for this, it should not be the role of the library to make expectations about the route cause.

> 
> > * handle new RX message notification in a seperate RX thread to manage remoteproc_get_notification mechanism
> 
> We already have a dedicated thread to handle the notification: https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L174-L206

Could you detail why this not unlock the use case?
regarding the code I supose that the thread is blocked on `nxsem_wait_uninterruptible(&priv->sem);`, right?
In this case why the mailbox does not unlock the semaphore in `rptun_wakeup`?

> 
> A dedicated thread or shared work thread can't help here. The dead lock can be fixed by:
> 
> 1. Process the incoming notification recursively in a same thread
> 2. Or creating more threads to process the notification
> 
> We select the first approach, because it save the resource which is important in embedded environment.

Right , that  why we need  to full understand the rational and explore other solutions before impacting other platform resources adding this extra code.

> 
> > ( releying on IPC signaling)  (see zephyr as example [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/ipc/rpmsg_service/rpmsg_backend.c#L150)
> 
> Zephyr implementation also suffer the same issue too: rpmsg_endpoint::cb may wait in rpmsg_send when the virtio buffer exhaust which then block the work thread to process the next incoming notification.
> 
> Please reread the description at the begin, and think about the work thread may blocking in waiting a new free buffer. This is a dead lock involve two CPU in the pressure testing.

I cannot see semaphore in openamp lib that would block the RX work thread when rpmsg_send is called. So please, could you detail for one CPU how exactly you are blocked? 
I woul like to understandf why the RX thread does not get the hand when TX thread is blocked waiting free buffer.
Does  the mailbox does not do its job to unlock the semaphore for the RX thread, or there is no task  reschedulling when metal_sleep_usec is called?

xiaoxiang781216:
> > > Hello @GUIDINGLI
> > > If i well understood your application is blocked in `rpmsg_virtio_get_tx_payload_buffer` waiting that a TX buffer is available. In such case it does not process the incoming RPMsg and after a while the remote processor is also blocked waiting available buffer.
> > > The patch you propose probably works for your senario, but look to me strange to use the rpmsg_send bocking interface. If you issue is that you are blocked during send without possibility to treat incoming message.
> > > I would see 2 other alternatives that could solve your issue.
> > > 
> > > * use non blocking APi instead `rpmsg_trysend` instead
> > 
> > 
> > But, how to handle rpmsg_trysend fail? it isn't a solution but push the problem to user.
> 
> RPMSG_ERR_NO_BUFF error should be used for this, it should not be the role of the library to make expectations about the route cause.
> 
> > > * handle new RX message notification in a seperate RX thread to manage remoteproc_get_notification mechanism
> > 
> > 
> > We already have a dedicated thread to handle the notification: https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L174-L206
> 
> Could you detail why this not unlock the use case? regarding the code I supose that the thread is blocked on `nxsem_wait_uninterruptible(&priv->sem);`, right? In this case why the mailbox does not unlock the semaphore in `rptun_wakeup`?
> 
> > A dedicated thread or shared work thread can't help here. The dead lock can be fixed by:
> > 
> > 1. Process the incoming notification recursively in a same thread
> > 2. Or creating more threads to process the notification
> > 
> > We select the first approach, because it save the resource which is important in embedded environment.
> 
> Right , that why we need to full understand the rational and explore other solutions before impacting other platform resources adding this extra code.
> 
> > > ( releying on IPC signaling)  (see zephyr as example [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/ipc/rpmsg_service/rpmsg_backend.c#L150)
> > 
> > 
> > Zephyr implementation also suffer the same issue too: rpmsg_endpoint::cb may wait in rpmsg_send when the virtio buffer exhaust which then block the work thread to process the next incoming notification.
> > Please reread the description at the begin, and think about the work thread may blocking in waiting a new free buffer. This is a dead lock involve two CPU in the pressure testing.
> 
> I cannot see semaphore in openamp lib that would block the RX work thread when rpmsg_send is called. So please, could you detail for one CPU how exactly you are blocked? I woul like to understandf why the RX thread does not get the hand when TX thread is blocked waiting free buffer. Does the mailbox does not do its job to unlock the semaphore for the RX thread, or there is no task reschedulling when metal_sleep_usec is called?

To simplify the case, let's assume we just have two virtio buffers in each direction:

1. Two threads on CPU1 call rpmsg_send and then no more free buffer from CPU1 to CPU2
2. Two threads on CPU2 call rpmsg_send at the same time and then no more free buffer from CPU2 to CPU1
3. Next, CPU2 RX dispatch call some EPT callback which may call rpmsg_send but has to wait due to step 2
4. Finally, CPU1 RX dispatch call some EPT callback which may call rpmsg_send too but has to wait due to step 1

With this patch, rpmsg_send(e.g. CPU2) will try to dispatch the next incoming message. If the new handler doesn't send something back, the dispatch handler return the buffer back to the pool immediately and then the blocking caller(CPU2) can get the free buffer now.

xiaoxiang781216:
> > > I cannot see semaphore in openamp lib that would block the RX work thread when rpmsg_send is called. So please, could you detail for one CPU how exactly you are blocked? I woul like to understandf why the RX thread does not get the hand when TX thread is blocked waiting free buffer. Does the mailbox does not do its job to unlock the semaphore for the RX thread, or there is no task reschedulling when metal_sleep_usec is called?
> > 
> > 
> > To simplify the case, let's assume we just have two virtio buffers in each direction:
> > 
> > 1. Two threads on CPU1 call rpmsg_send and then no more free buffer from CPU1 to CPU2
> > 2. Two threads on CPU2 call rpmsg_send at the same time and then no more free buffer from CPU2 to CPU1
> > 3. Next, CPU2 RX dispatch call some EPT callback which may call rpmsg_send but has to wait due to step 2
> > 4. Finally, CPU1 RX dispatch call some EPT callback which may call rpmsg_send too but has to wait due to step 1
> > 
> > With this patch, rpmsg_send(e.g. CPU2) will try to dispatch the next incoming message. If the new handler doesn't send something back, the dispatch handler return the buffer back to the pool immediately and then the blocking caller(CPU2) can get the free buffer now.
> 
> Thanks for the clarification. In your use case, an rpmsg message is sent in the rpmsg RX callback, and the RX buffer is released only if the TX rpmsg is sent, right? Indeed, this will lead to a deadlock. And as you mentioned the patch is working only in the case where the RX thread free the RX buffer without sending something back.
> 

Yes, in case of two virtio buffers, the next one can't call rpmsg_send in ept callback, but if there are more virtio buffers, the recursive could continue to next level.

> It seems to me that a better way to solve this problem would be to free the RX buffer before sending the rpmsg TX.
> 

We have try this approach too, but it can't avoid all possible case, since it can just return one buffer to pool, and this buffer may grab by another thread(imaging we extend the above case to three threads) not the RX dispatch thread. And it require more change:

1. Add new API to return the buffer before calling rpmsg_send
2. All ept callback which call rpmsg_send must call the new API

Actually, we plan to enable this approach too.

> For NuttX, why not use rpmsg_trysend and, if it fails, create a queue?

The message is critical in many case, so even the user call rpmsg_trysend, it still has to handle the buffer exhaust case. So, I don't see any benefit to call rpmsg_trysend(why not always use the queue directly?).

> The zero copy API could also be used in RX to treat the message out of the RX callback

rpmsg_send we mentioned here just to simplify the case,  we call rpmsg_get_tx_payload_buffer in RX callback too, which also suffer the potential deadlock.

arnopo:
> > Thanks for the clarification. In your use case, an rpmsg message is sent in the rpmsg RX callback, and the RX buffer is released only if the TX rpmsg is sent, right? Indeed, this will lead to a deadlock. And as you mentioned the patch is working only in the case where the RX thread free the RX buffer without sending something back.
> 
> Yes, in case of two virtio buffers, the next one can't call rpmsg_send in ept callback, but if there are more virtio buffers, the recursive could continue to next level.
> 
> > It seems to me that a better way to solve this problem would be to free the RX buffer before sending the rpmsg TX.
> 
> We have try this approach too, but it can't avoid all possible case, since it can just return one buffer to pool, and this buffer may grab by another thread(imaging we extend the above case to three threads) not the RX dispatch thread. 

calling rpmsg_virtio_rx_callback should it will not guarantee it either, and could lead to a call stack overflow.
 
> And it require more change:
> 
> 1. Add new API to return the buffer before calling rpmsg_send

The APi already exists: `rpmsg_release_rx_buffer`. But doing this in the callback would means additional code to detect in rpmsg_virtio_rx_callback that the buffer as already been released.  Not in favour of such workaround.  

> 2. All ept callback which call rpmsg_send must call the new API
> 
> Actually, we plan to enable this approach too.
> 
> > For NuttX, why not use rpmsg_trysend and, if it fails, create a queue?
> 
> The message is critical in many case, so even the user call rpmsg_trysend, it still has to handle the buffer exhaust case. So, I don't see any benefit to call rpmsg_trysend(why not always use the queue directly?).
> 
> > The zero copy API could also be used in RX to treat the message out of the RX callback
> 
> rpmsg_send we mentioned here just to simplify the case, we call rpmsg_get_tx_payload_buffer in RX callback too, which also suffer the potential deadlock.

Here are 2 approaches I have in mind (dirty code just to show the principle):

1) create a workqueue systematically

```
struct rpmsg_rcv_msg {
	struct _payload *payload;
	size_t len;
	struct rpmsg_endpoint *ept;
};
static workqueue_rx_cb ( void * data) {
        struct rpmsg_rcv_msg *rcv_msg = (struct rpmsg_rcv_msg *)data;

	rpmsg_release_rx_buffer(rcv_msg->ept, r_payload);
        /* treat your rx message */
       rpmsg_send(ept, tx_data, len);
}

static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
			     uint32_t src, void *priv)
{
        struct rpmsg_rcv_msg *rcv_msg;

	rcv_msg.ept = ept;
	rcv_msg.payload =  (struct _payload *)data;
	rcv_msg.len = len;

	rpmsg_hold_rx_buffer(ept, data);
        work_queue(RX_CB_WQ_ID, work, workqueue_rx_cb, rcv_msg, 0);

	return RPMSG_SUCCESS;
}

```
1) using try_send to be more proactive, create a workqueue as backup

```
struct rpmsg_tx_msg {
	struct _payload *payload;
	size_t len;
	struct rpmsg_endpoint *ept;
};
static workqueue_tx_cb ( void * data) {
       struct rpmsg_tx_msg *tx_msg = (struct rpmsg_tx_msg *)data;

       rpmsg_send(rpmsg_tx_msg.ept , rpmsg_tx_msg.payload, rpmsg_tx_msg.len );
}

static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
			     uint32_t src, void *priv)
{
	(void)priv;
        struct rpmsg_tx_msg *tx_msg;

        /* treat your rx message */

       if (rpmsg_trysend (ept, tx_data, len)) {
	    rpmsg_tx_msg.ept = ept;
	    rpmsg_tx_msg.payload =  (struct _payload *)tx_data;
	    rpmsg_tx_msg.len = len;
            work_queue(RX_CB_WQ_ID, work, workqueue_tx_cb, rpmsg_tx_msg, 0);
       }
	return RPMSG_SUCCESS;
}

```

xiaoxiang781216:
> > > Thanks for the clarification. In your use case, an rpmsg message is sent in the rpmsg RX callback, and the RX buffer is released only if the TX rpmsg is sent, right? Indeed, this will lead to a deadlock. And as you mentioned the patch is working only in the case where the RX thread free the RX buffer without sending something back.
> > 
> > 
> > Yes, in case of two virtio buffers, the next one can't call rpmsg_send in ept callback, but if there are more virtio buffers, the recursive could continue to next level.
> > > It seems to me that a better way to solve this problem would be to free the RX buffer before sending the rpmsg TX.
> > 
> > 
> > We have try this approach too, but it can't avoid all possible case, since it can just return one buffer to pool, and this buffer may grab by another thread(imaging we extend the above case to three threads) not the RX dispatch thread.
> 
> calling rpmsg_virtio_rx_callback should it will not guarantee it either, and could lead to a call stack overflow.
> 
> > And it require more change:
> > 
> > 1. Add new API to return the buffer before calling rpmsg_send
> 
> The APi already exists: `rpmsg_release_rx_buffer`. But doing this in the callback would means additional code to detect in rpmsg_virtio_rx_callback that the buffer as already been released. Not in favour of such workaround.
> 

Yes, @GUIDINGLI and me discuss this before, the best approach we can come up is returning a special value from ept callback, so rpmsg_virtio_rx_callback can check this value and then skip the release.

> > 2. All ept callback which call rpmsg_send must call the new API
> > 
> > Actually, we plan to enable this approach too.
> > > For NuttX, why not use rpmsg_trysend and, if it fails, create a queue?
> > 
> > 
> > The message is critical in many case, so even the user call rpmsg_trysend, it still has to handle the buffer exhaust case. So, I don't see any benefit to call rpmsg_trysend(why not always use the queue directly?).
> > > The zero copy API could also be used in RX to treat the message out of the RX callback
> > 
> > 
> > rpmsg_send we mentioned here just to simplify the case, we call rpmsg_get_tx_payload_buffer in RX callback too, which also suffer the potential deadlock.
> 
> Here are 2 approaches I have in mind (dirty code just to show the principle):
> 
> 1. create a workqueue systematically
> 
> ```
> struct rpmsg_rcv_msg {
> 	struct _payload *payload;
> 	size_t len;
> 	struct rpmsg_endpoint *ept;
> };
> static workqueue_rx_cb ( void * data) {
>         struct rpmsg_rcv_msg *rcv_msg = (struct rpmsg_rcv_msg *)data;
> 
> 	rpmsg_release_rx_buffer(rcv_msg->ept, r_payload);
>         /* treat your rx message */
>        rpmsg_send(ept, tx_data, len);
> }
> 
> static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
> 			     uint32_t src, void *priv)
> {
>         struct rpmsg_rcv_msg *rcv_msg;
> 
> 	rcv_msg.ept = ept;
> 	rcv_msg.payload =  (struct _payload *)data;
> 	rcv_msg.len = len;
> 
> 	rpmsg_hold_rx_buffer(ept, data);
>         work_queue(RX_CB_WQ_ID, work, workqueue_rx_cb, rcv_msg, 0);
> 
> 	return RPMSG_SUCCESS;
> }
> ```
> 
> 1. using try_send to be more proactive, create a workqueue as backup
> 
> ```
> struct rpmsg_tx_msg {
> 	struct _payload *payload;
> 	size_t len;
> 	struct rpmsg_endpoint *ept;
> };
> static workqueue_tx_cb ( void * data) {
>        struct rpmsg_tx_msg *tx_msg = (struct rpmsg_tx_msg *)data;
> 
>        rpmsg_send(rpmsg_tx_msg.ept , rpmsg_tx_msg.payload, rpmsg_tx_msg.len );
> }
> 
> static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
> 			     uint32_t src, void *priv)
> {
> 	(void)priv;
>         struct rpmsg_tx_msg *tx_msg;
> 
>         /* treat your rx message */
> 
>        if (rpmsg_trysend (ept, tx_data, len)) {
> 	    rpmsg_tx_msg.ept = ept;
> 	    rpmsg_tx_msg.payload =  (struct _payload *)tx_data;
> 	    rpmsg_tx_msg.len = len;
>             work_queue(RX_CB_WQ_ID, work, workqueue_tx_cb, rpmsg_tx_msg, 0);
>        }
> 	return RPMSG_SUCCESS;
> }
> ```

Actually, we discuss and even try all methods you mention above:

1. Call rpmsg_release_rx_buffer before rpmsg_send
2. Schedule to the work queue
3. Call rpmsg_virtio_rx_callback recursively

Method 1 can just return one buffer to pool, method 2 require more work threads to return more buffers(1:1), method 3 require additional recursive to return additional buffer(1:1).

Thread consumption is much bigger than recursive function, so I think method 3 is better than method 2. As I said before, we will provide method 1 in the next step, but only method 1 may still hit the deadlock in the high pressure IPC loading. The best fix is combine both method 1 and method 3.

edmooring:
@GUIDINGLI @xiaoxiang781216 
Can you test your scenario with the attached patch?

I noticed that the rx callback will only process one buffer at a time. If the buffer is held, or more than one buffer has been used since the last notification, no buffers will be freed. This patch changes the callback to process as many buffers as are available
[0001-rpmsg_virtio-Process-multiple-buffers-in-a-single-re.patch.txt](https://github.com/OpenAMP/open-amp/files/7677511/0001-rpmsg_virtio-Process-multiple-buffers-in-a-single-re.patch.txt)
.

PR #321, opened 2021-12-02
CI: Update branch from "master" to "main".

edmooring:
Rename branch for CI and compliance checks to "main".


arnopo (approved):
I let you finalize the migration on main branch, as you already done it in libmetal
Thanks

PR #323, opened 2021-12-06
Replace potentially objectionable terms in OpenAMP variable names.

edmooring:
The OpenAMP project Technical Steering Committee has requested that potentially objectionable terms be removed from the OpenAMP code base. This is the first step in removing them. It replaces these terms in variable/identifier names within the code base. 

Further pull requests will fix the documentation and comments in the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,10 +82,10 @@ library for it project:
 * **WITH_PROXY** (default OFF): Include proxy support in the library.
 * **WITH APPS** (default OFF): Build with sample applications.
 * **WITH_PROXY_APPS** (default OFF):Build with proxy sample applications.
-* **WITH_VIRTIO_MASTER** (default ON): Build with virtio master enabled.
+* **WITH_VIRTIO_DRIVER** (default ON): Build with virtio driver enabled.
   This option can be set to OFF if the only the remote mode is implemented.
-* **WITH_VIRTIO_SLAVE** (default ON): Build with virtio slave enabled.
-  This option can be set to OFF if the only the master mode is implemented.
+* **WITH_VIRTIO_DEVICE** (default ON): Build with virtio device enabled.
+  This option can be set to OFF if the only the driver mode is implemented.
 * **WITH_STATIC_LIB** (default ON): Build with a static library.
 * **WITH_SHARED_LIB** (default ON): Build with a shared library.
 * **WITH_ZEPHYR** (default OFF): Build open-amp as a zephyr library. This option
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -63,16 +63,25 @@ if (NOT ${MACHINE} STREQUAL "zynqmp_r5")
  set (WITH_LOAD_FW OFF)
 endif(NOT ${MACHINE} STREQUAL "zynqmp_r5")
 
-option (WITH_VIRTIO_MASTER "Build with virtio master enabled" ON)
-option (WITH_VIRTIO_SLAVE "Build with virtio slave enabled" ON)
+option (WITH_VIRTIO_DRIVER "Build with virtio driver (front end)  enabled" ON)
+option (WITH_VIRTIO_DEVICE "Build with virtio device (back end)  enabled" ON)
+option (WITH_VIRTIO_MASTER "Build with virtio driver (front end)  enabled" OFF)
+option (WITH_VIRTIO_SLAVE "Build with virtio device (back end)  enabled" OFF)
 
-if (NOT WITH_VIRTIO_MASTER)
-  add_definitions(-DVIRTIO_SLAVE_ONLY)
-endif (NOT WITH_VIRTIO_MASTER)
+if (WITH_VIRTIO_MASTER)
+  message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DRIVER" ...)
+endif (WITH_VIRTIO_MASTER)
+if (WITH_VIRTIO_SLAVE)
+  message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DEVICE" ...)
+endif (WITH_VIRTIO_SLAVE)
 
-if (NOT WITH_VIRTIO_SLAVE)
-  add_definitions(-DVIRTIO_MASTER_ONLY)
-endif (NOT WITH_VIRTIO_SLAVE)
+if (NOT WITH_VIRTIO_DRIVER AND NOT WITH_VIRTIO_MASTER)
+	add_definitions(-DVIRTIO_DEVICE_ONLY)
+endif (NOT WITH_VIRTIO_DRIVER AND NOT WITH_VIRTIO_MASTER)
+
+if (NOT WITH_VIRTIO_DEVICE AND NOT WITH_VIRTIO_SLAVE)
+	add_definitions(-DVIRTIO_DRIVER_ONLY)
+endif (NOT WITH_VIRTIO_DEVICE AND NOT WITH_VIRTIO_SLAVE)
 
 option (WITH_DCACHE_VRINGS "Build with vrings cache operations enabled" OFF)
 
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -75,8 +75,23 @@ struct rpmsg_virtio_device {
 	struct rpmsg_virtio_shm_pool *shpool;
 };
 
-#define RPMSG_REMOTE	VIRTIO_DEV_SLAVE
-#define RPMSG_MASTER	VIRTIO_DEV_MASTER
+#define RPMSG_REMOTE	VIRTIO_DEV_DEVICE
+#define RPMSG_HOST	VIRTIO_DEV_DRIVER
+
+#define RPMSG_SLAVE        deprecated_rpmsg_slave()
+#define RPMSG_MASTER       deprecated_rpmsg_master()
+
+__deprecated static inline int deprecated_rpmsg_master(void)
+{
+	/* "RPMSG_MASTER is deprecated, please use RPMSG_HOST" */
+	return RPMSG_HOST;
+}
+
+__deprecated static inline int deprecated_rpmsg_slave(void)
+{
+	/* "RPMSG_SLAVE is deprecated, please use RPMSG_REMOTE" */
+	return RPMSG_REMOTE;
+}
 
 static inline unsigned int
 rpmsg_virtio_get_role(struct rpmsg_virtio_device *rvdev)
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -34,8 +34,33 @@ extern "C" {
 #define VIRTIO_CONFIG_STATUS_FAILED    0x80
 
 /* Virtio device role */
-#define VIRTIO_DEV_MASTER	0UL
-#define VIRTIO_DEV_SLAVE	1UL
+#define VIRTIO_DEV_DRIVER	0UL
+#define VIRTIO_DEV_DEVICE	1UL
+
+#define VIRTIO_DEV_MASTER	deprecated_virtio_dev_master()
+#define VIRTIO_DEV_SLAVE	deprecated_virtio_dev_slave()
+
+__deprecated static inline int deprecated_virtio_dev_master(void)
+{
+	/* "VIRTIO_DEV_MASTER is deprecated, please use VIRTIO_DEV_DRIVER" */
+	return VIRTIO_DEV_DRIVER;
+}
+
+__deprecated static inline int deprecated_virtio_dev_slave(void)
+{
+	/* "VIRTIO_DEV_SLAVE is deprecated, please use VIRTIO_DEV_DEVICE" */
+	return VIRTIO_DEV_DEVICE;
+}
+
+#ifdef VIRTIO_MASTER_ONLY
+#define VIRTIO_DRIVER_ONLY
+#warning "VIRTIO_MASTER_ONLY is deprecated, please use VIRTIO_DRIVER_ONLY"
+#endif
+
+#ifdef VIRTIO_SLAVE_ONLY
+#define VIRTIO_DEVICE_ONLY
+#warning "VIRTIO_SLAVE_ONLY is deprecated, please use VIRTIO_DEVICE_ONLY"
+#endif
 
 struct virtio_device_id {
 	uint32_t device;
--- a/apps/examples/echo/rpmsg-echo.c
+++ b/apps/examples/echo/rpmsg-echo.c
@@ -113,7 +113,7 @@ int main(int argc, char *argv[])
 		ret = -1;
 	} else {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
-						   VIRTIO_DEV_SLAVE,
+						   VIRTIO_DEV_DEVICE,
 						   NULL, NULL);
 		if (!rpdev) {
 			LPERROR("Failed to create rpmsg virtio device.\r\n");
--- a/apps/examples/echo/rpmsg-ping.c
+++ b/apps/examples/echo/rpmsg-ping.c
@@ -185,7 +185,7 @@ int main(int argc, char *argv[])
 		ret = -1;
 	} else {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
-						  VIRTIO_DEV_MASTER,
+						  VIRTIO_DEV_DRIVER,
 						  NULL,
 						  rpmsg_name_service_bind_cb);
 		if (!rpdev) {
--- a/apps/examples/linux_rpc_demo/linux_rpc_demo.c
+++ b/apps/examples/linux_rpc_demo/linux_rpc_demo.c
@@ -580,7 +580,7 @@ int main(int argc, char *argv[])
 		ret = -1;
 	} else {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
-						   VIRTIO_DEV_SLAVE,
+						   VIRTIO_DEV_DEVICE,
 						   NULL, NULL);
 		if (!rpdev) {
 			LPERROR("Failed to create rpmsg virtio device.\r\n");
--- a/apps/examples/linux_rpc_demo/linux_rpc_demod.c
+++ b/apps/examples/linux_rpc_demo/linux_rpc_demod.c
@@ -297,7 +297,7 @@ int main(int argc, char *argv[])
 		ret = -1;
 	} else {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
-						   VIRTIO_DEV_MASTER,
+						   VIRTIO_DEV_DRIVER,
 						   NULL, NULL);
 		if (!rpdev) {
 			LPERROR("Failed to create rpmsg virtio device.\r\n");
--- a/apps/examples/matrix_multiply/matrix_multiply.c
+++ b/apps/examples/matrix_multiply/matrix_multiply.c
@@ -240,7 +240,7 @@ int main(int argc, char *argv[])
 		ret = -1;
 	} else {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
-						  VIRTIO_DEV_MASTER,
+						  VIRTIO_DEV_DRIVER,
 						  NULL,
 						  rpmsg_name_service_bind_cb);
 		if (!rpdev) {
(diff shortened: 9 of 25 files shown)


edmooring:
This breaks the CI build on Zephyr because the corresponding changes have not been made in the Zephyr code base. This points up the need to coordinate with other consumers.

arnopo:
> This breaks the CI build on Zephyr because the corresponding changes have not been made in the Zephyr code base. This points up the need to coordinate with other consumers.

What about adding this in code to manage the deprecation? we could keep both for release 2022.04 and clean up for next one.

```
#ifdef VIRTIO_MASTER_ONLY
#define VIRTIO_DRIVER_ONLY
#if defined(__GNUC__)
_Pragma("GCC warning\"Please use VIRTIO_DRIVER_ONLY instead VIRTIO_MASTER_ONLY which is deprecated\"")
#elif defined(__ICCARM__)
#define DO_PRAGMA(x) _Pragma(#x)
DO_PRAGMA(message("Warning: Please use VIRTIO_DRIVER_ONLY instead VIRTIO_MASTER_ONLY which is deprecated"))
#endif
```
same for VIRTIO_SLAVE_ONLY and  RPMSG_MASTER

edmooring:
> > This breaks the CI build on Zephyr because the corresponding changes have not been made in the Zephyr code base. This points up the need to coordinate with other consumers.
> 
> What about adding this in code to manage the deprecation? we could keep both for release 2022.04 and clean up for next one.
> 
> ```
> #ifdef VIRTIO_MASTER_ONLY
> #define VIRTIO_DRIVER_ONLY
> #if defined(__GNUC__)
> _Pragma("GCC warning\"Please use VIRTIO_DRIVER_ONLY instead VIRTIO_MASTER_ONLY which is deprecated\"")
> #elif defined(__ICCARM__)
> #define DO_PRAGMA(x) _Pragma(#x)
> DO_PRAGMA(message("Warning: Please use VIRTIO_DRIVER_ONLY instead VIRTIO_MASTER_ONLY which is deprecated"))
> #endif
> ```
> 
> same for VIRTIO_SLAVE_ONLY and RPMSG_MASTER

This seems like a good plan. 

Since this is technically a change in the API, we might need to hold off on cleaning up until 2022.10. Also, do you think I should bump the API minor or patch version for this?

arnopo:
> > > This breaks the CI build on Zephyr because the corresponding changes have not been made in the Zephyr code base. This points up the need to coordinate with other consumers.
> > 
> > 
> > What about adding this in code to manage the deprecation? we could keep both for release 2022.04 and clean up for next one.
> > ```
> > #ifdef VIRTIO_MASTER_ONLY
> > #define VIRTIO_DRIVER_ONLY
> > #if defined(__GNUC__)
> > _Pragma("GCC warning\"Please use VIRTIO_DRIVER_ONLY instead VIRTIO_MASTER_ONLY which is deprecated\"")
> > #elif defined(__ICCARM__)
> > #define DO_PRAGMA(x) _Pragma(#x)
> > DO_PRAGMA(message("Warning: Please use VIRTIO_DRIVER_ONLY instead VIRTIO_MASTER_ONLY which is deprecated"))
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > same for VIRTIO_SLAVE_ONLY and RPMSG_MASTER
> 
> This seems like a good plan. The only problem I can see is which file to put it in. These constants are generated from options.cmake, and wind up on the compile command line as -DVIRTIOxx.

The point here is for application. Adding VIRTIO_SLAVE_ONLY and VIRTIO_MASTER_ONLY in virtio.h should work.
Same for RPMSG_HOST in rpmsg_virtio.h.

For option.cmake: I would suggest to duplicate instead of replace, as a first step, adding a deprecated warning message for legacy definitions.
```
option (WITH_VIRTIO_MASTER "Build with virtio driver (front end)  enabled" ON)
option (WITH_VIRTIO_SLAVE "Build with virtio device (back end)  enabled" ON)
option (WITH_VIRTIO_DRIVER "Build with virtio driver (front end)  enabled" ON)
option (WITH_VIRTIO_DEVICE "Build with virtio device (back end)  enabled" ON)

if (WITH_VIRTIO_MASTER)
  message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DRIVER" ...)
endif (WITH_VIRTIO_MASTER)
if (WITH_VIRTIO_SLAVE)
  message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DEVICE" ...)
endif (WITH_VIRTIO_SLAVE)

if (NOT WITH_VIRTIO_DRIVER AND NOT WITH_VIRTIO_MASTER)
	add_definitions(-DVIRTIO_DEVICE_ONLY)
endif (NOT WITH_VIRTIO_DRIVER AND NOT WITH_VIRTIO_MASTER)

if (NOT WITH_VIRTIO_DEVICE AND NOT WITH_VIRTIO_SLAVE)
	add_definitions(-DVIRTIO_DRIVER_ONLY)
endif (NOT WITH_VIRTIO_DEVICE AND NOT WITH_VIRTIO_SLAVE)

```



> 
> Since this is technically a change in the API, we might need to hold off on cleaning up until 2022.10. 

Yes

> Also, do you think I should bump the API minor or patch version for this?

I propose to discuss this when we will release the library, some other PRs could change the picture, in the meantime.

arnopo:
Hi @edmooring 
Do you plan to rework this PR before the release?

edmooring:
> Hi @edmooring
> Do you plan to rework this PR before the release?

I will be working on it. I should have this one ready for the release, but the corresponding documentation update might not be ready.

edmooring:
The checkpatch failures are all lines >80 characters. Shall we just leave them alone, or shall I try to fix them?

The Zephyr failure appears to be that Zephyr is treating CMake deprecation warnings as errors. I'm not sure what the right thing to do with them is.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -75,8 +75,26 @@ struct rpmsg_virtio_device {
...
+#endif
+
+#ifdef VIRTIO_DEV_SLAVE
+#if defined(__GNUC__)
+_Pragma("GCC warning\"Please use VIRTIO_DEV_DEVICE instead of VIRTIO_DEV_SLAVE which is deprecated\"")
+#elif defined(__ICCARM__)
+#define DO_PRAGMA(x) _Pragma(#x)
+DO_PRAGMA(message("Warning: Please use VIRTIO_DEV_DEVICE instead of VIRTIO_DEV_SLAVE which is deprecated"))
+#endif
+#endif
```
arnopo, line 97: To also limit message size here in same way
In parallel I will send a PR to increate the line length limit to 100 ( as done in Linux kernel)

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -75,8 +75,26 @@ struct rpmsg_virtio_device {
...
+#define RPMSG_REMOTE	VIRTIO_DEV_DEVICE
+#define RPMSG_HOST	VIRTIO_DEV_DRIVER
+
+#ifdef VIRTIO_DEV_MASTER
+#if defined(__GNUC__)
+_Pragma("GCC warning\"Please use VIRTIO_DEV_DRIVER instead of VIRTIO_DEV_MASTER which is deprecated\"")
+#elif defined(__ICCARM__)
+#define DO_PRAGMA(x) _Pragma(#x)
+DO_PRAGMA(message("Warning: Please use VIRTIO_DEV_DRIVER instead of VIRTIO_DEV_MASTER which is deprecated"))
+#endif
```
arnopo, line 87:
```
#if defined(__GNUC__)
_Pragma("GCC warning\" deprecated VIRTIO_MASTER_ONLY is replaced by VIRTIO_DRIVER_ONLY\"")
#elif defined(__ICCARM__)
#define DO_PRAGMA(x) _Pragma(#x)
DO_PRAGMA(message("Warning: deprecated VIRTIO_MASTER_ONLY is replaced by VIRTIO_DRIVER_ONLY"))
#endif

```
should avoid too long line ans seems aligned with the cmake comments

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -75,8 +75,26 @@ struct rpmsg_virtio_device {
 	struct rpmsg_virtio_shm_pool *shpool;
 };
 
-#define RPMSG_REMOTE	VIRTIO_DEV_SLAVE
-#define RPMSG_MASTER	VIRTIO_DEV_MASTER
+#define RPMSG_REMOTE	VIRTIO_DEV_DEVICE
+#define RPMSG_HOST	VIRTIO_DEV_DRIVER
```
arnopo, line 79: hare you sure that `VIRTIO_DEV_DRIVER` and 'VIRTIO_DEV_DEVICE'  are defined at this step?
else should be put after the pragma messgaes
```
#ifdef VIRTIO_DEV_MASTER 
[...]
+ #ifndef VIRTIO_DEV_DRIVER
+ #define VIRTIO_DEV_DRIVER
+ #endif
#endif

#ifdef VIRTIO_DEV_SLAVE
[...]
+ #ifndef VIRTIO_DEV_DEVICE
+ #define VIRTIO_DEV_DEVICE
+ #endif
#endif

#define RPMSG_REMOTE	VIRTIO_DEV_DEVICE
#define RPMSG_HOST	VIRTIO_DEV_DRIVER
```

lib/include/openamp/virtio.h
```diff
@@ -34,8 +34,28 @@ extern "C" {
...
+
+#ifdef VIRTIO_SLAVE_ONLY
+#define VIRTIO_DEVICE_ONLY
+#if defined(__GNUC__)
+_Pragma("GCC warning\"Please use VIRTIO_DEVICE_ONLY instead of VIRTIO_SLAVE_ONLY which is deprecated\"")
+#elif defined(__ICCARM__)
+#define DO_PRAGMA(x) _Pragma(#x)
+DO_PRAGMA(message("Warning: Please use VIRTIO_DEVICE_ONLY instead of VIRTIO_SLAVE_ONLY which is deprecated"))
+#endif
+#endif
```
arnopo, line 58: Same comment than in rpmsg_vitio.h

README.md
```diff
@@ -82,10 +82,10 @@ library for it project:
 * **WITH_PROXY** (default OFF): Include proxy support in the library.
 * **WITH APPS** (default OFF): Build with sample applications.
 * **WITH_PROXY_APPS** (default OFF):Build with proxy sample applications.
-* **WITH_VIRTIO_MASTER** (default ON): Build with virtio master enabled.
+* **WITH_VIRTIO_DRIVER** (default ON): Build with virtio driver (front-end)  enabled.
   This option can be set to OFF if the only the remote mode is implemented.
-* **WITH_VIRTIO_SLAVE** (default ON): Build with virtio slave enabled.
-  This option can be set to OFF if the only the master mode is implemented.
+* **WITH_VIRTIO_DEVICE** (default ON): Build with virtio device (back-end)  enabled.
+  This option can be set to OFF if the only the driver mode is implemented.
```
arnopo, line 88: `Front-end` and `back-end` terms can be ambigious.
suggestion:
 Build with virtio driver (virtio host role) enabled. 
 Build with virtio device (virtio guest role) enabled.

cmake/options.cmake
```diff
@@ -68,13 +68,15 @@ option (WITH_VIRTIO_DEVICE "Build with virtio device (back end)  enabled" ON)
...
 
+set (CMAKE_ERROR_DEPRECATED OFF)
 if (WITH_VIRTIO_MASTER)
   message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DRIVER" ...)
 endif (WITH_VIRTIO_MASTER)
 if (WITH_VIRTIO_SLAVE)
   message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DEVICE" ...)
 endif (WITH_VIRTIO_SLAVE)
 
+
```
arnopo, line 79: useless line

cmake/options.cmake
```diff
@@ -68,13 +68,15 @@ option (WITH_VIRTIO_DEVICE "Build with virtio device (back end)  enabled" ON)
 option (WITH_VIRTIO_MASTER "Build with virtio driver (front end)  enabled" ON)
 option (WITH_VIRTIO_SLAVE "Build with virtio device (back end)  enabled" ON)
 
+set (CMAKE_ERROR_DEPRECATED OFF)
```
arnopo, line 71: I don't think it's a good idea, if OS's like Zephyr decide to treat depreciation as an error, we shouldn't silence the error. 
I guess they want to be up to date by using this mechanism....

arnopo (changes requested):
please rework also you commits messages to suppress` signed-off `duplication

lib/include/openamp/rpmsg_virtio.h
arnopo: I made more tests with Gcc Armcc and arm 

> ```
> #if defined(__GNUC__)
> _Pragma("GCC warning\" deprecated VIRTIO_MASTER_ONLY is replaced by VIRTIO_DRIVER_ONLY\"")
> #elif defined(__ICCARM__)
> #define DO_PRAGMA(x) _Pragma(#x)
> DO_PRAGMA(message("Warning: deprecated VIRTIO_MASTER_ONLY is replaced by VIRTIO_DRIVER_ONLY"))
> #endif
> ```
> 
> should avoid too long line ans seems aligned with the cmake comments

In fact, "_Pragma" doesn't work for armcc. Why not make it simpler by using #warning, which should work for all compilers?
```
#ifdef VIRTIO_MASTER_ONLY
#define VIRTIO_DRIVER_ONLY
#warning("Please use VIRTIO_DRIVER_ONLY instead of VIRTIO_MASTER_ONLY which is deprecated\")
#endif

#ifdef VIRTIO_SLAVE_ONLY
#define VIRTIO_DEVICE_ONLY
#warning("Please use VIRTIO_DEVICE_ONLY instead of VIRTIO_SLAVE_ONLY which is deprecated\")
#endif
```

arnopo (changes requested):

lib/include/openamp/rpmsg_virtio.h
edmooring: They are defined in openamp/virtio.h, which is included at the top of the file, so this should be fine. I'm not sure we need the ifdef checks and warnings for the old names, since the only place these constants are ever defined is in the header file.

cmake/options.cmake
edmooring: Removed this line.

README.md
edmooring: > `Front-end` and `back-end` terms can be ambigious.
> suggestion:
> Build with virtio driver (virtio host role) enabled.
> Build with virtio device (virtio guest role) enabled.

This is the same wording as in the original. I'd like to keep this commit focused on the variable name changes.

lib/include/openamp/rpmsg_virtio.h
edmooring: That looks much better. I will switch to \#warning.

README.md
arnopo: > > `Front-end` and `back-end` terms can be ambigious.
> > suggestion:
> > Build with virtio driver (virtio host role) enabled.
> > Build with virtio device (virtio guest role) enabled.
> 
> This is the same wording as in the original. I'd like to keep this commit focused on the variable name changes.
In this case could you just suppress the Front-end` and `back-end` terms?

cmake/options.cmake
```diff
@@ -63,16 +63,25 @@ if (NOT ${MACHINE} STREQUAL "zynqmp_r5")
...
+  message(DEPRECATION "deprecated cmake option replaced by WITH_VIRTIO_DEVICE" ...)
+endif (WITH_VIRTIO_SLAVE)
+
+if (NOT WITH_VIRTIO_DRIVER AND NOT WITH_VIRTIO_MASTER)
+	add_definitions(-DVIRTIO_DEVICE_ONLY)
+endif (NOT WITH_VIRTIO_DRIVER AND NOT WITH_VIRTIO_MASTER)
+
+if (NOT WITH_VIRTIO_DEVICE AND NOT WITH_VIRTIO_SLAVE)
+	add_definitions(-DVIRTIO_DRIVER_ONLY)
+endif (NOT WITH_VIRTIO_DEVICE AND NOT WITH_VIRTIO_SLAVE)
```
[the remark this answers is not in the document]
arnopo: > Building generate dprecation warnings by default as `WITH_VIRTIO_MASTER` and `WITH_VIRTIO_SLAVE` are enabled by default. following code seems working better:
> 
> ```
> if (NOT WITH_VIRTIO_MASTER)
>   message(DEPRECATION "deprecated WITH_VIRTIO_MASTER option replaced by WITH_VIRTIO_DRIVER" ...)
> endif (NOT WITH_VIRTIO_MASTER)
> if (NOT WITH_VIRTIO_SLAVE)
>   message(DEPRECATION "deprecated WITH_VIRTIO_SLAVE option replaced by WITH_VIRTIO_DEVICE" ...)
> endif (NOT WITH_VIRTIO_SLAVE)
> 
> if (NOT WITH_VIRTIO_DRIVER OR NOT WITH_VIRTIO_MASTER)
>  	add_definitions(-DVIRTIO_DEVICE_ONLY)
> endif (NOT WITH_VIRTIO_DRIVER OR NOT WITH_VIRTIO_MASTER)
>  
> if (NOT WITH_VIRTIO_DEVICE OR NOT WITH_VIRTIO_SLAVE)
>  	add_definitions(-DVIRTIO_DRIVER_ONLY)
> endif (NOT WITH_VIRTIO_DEVICE OR NOT WITH_VIRTIO_SLAVE)
> ```
This comment as not taken into account

lib/include/openamp/virtio.h
```diff
@@ -34,8 +34,19 @@ extern "C" {
...
+#define VIRTIO_DEV_DEVICE	1UL
+
+#ifdef VIRTIO_MASTER_ONLY
+#define VIRTIO_DRIVER_ONLY
+#warning "VIRTIO_MASTER_ONLY is deprecated, please use VIRTIO_DRIVER_ONLY"
+#endif
+
+#ifdef VIRTIO_SLAVE_ONLY
+#define VIRTIO_DEVICE_ONLY
+#define DO_PRAGMA(x) _Pragma(#x)
```
arnopo, line 47: this line seems useless

arnopo (changes requested):
There are still some comments to address
Could you also put your signed-off at the end of the commit message?

cmake/options.cmake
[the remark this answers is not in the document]
edmooring: I just defaulted the deprecated variables to "OFF" instead of "ON". That gets rid of the warnings on a normal build.
It seems like the top half just warns if the offending variables are turned off, which seems backwards to me.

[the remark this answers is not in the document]
arnopo: You are right, I have re-tested in different configurations (mainly on zephyr), it works. 
My test shows a potential side effect: VIRTIO_DEVICE_ONLY or DVIRTIO_DRIVER_ONLY may not be configured 
an increase of the library.

lib/include/openamp/virtio.h
arnopo: I have taken a look at Zephyr CI fails and would like to suggest an update to your RP.
I have added a fix on top of your patch that solves the `'RPMSG_MASTER' undeclared` [issue](https://github.com/OpenAMP/open-amp/runs/6205417148?check_suite_focus=true#step:5:25772).
Please find my update on my github https://github.com/arnopo/open-amp/pull/17 with associated CI that is successful. 

The error become a warning:
```
/github/workspace/zephyrproject/zephyr/samples/subsys/ipc/openamp/src/main.c:229:2: warning: 'deprecated_rpmsg_master' is deprecated [-Wdeprecated-declarations]
  229 |  vdev.role = RPMSG_MASTER;
      |  ^~~~
In file included from /github/workspace/zephyrproject/modules/lib/open-amp/open-amp/lib/include/openamp/open_amp.h:12,
                 from /github/workspace/zephyrproject/zephyr/samples/subsys/ipc/openamp/src/main.c:18:
/github/workspace/zephyrproject/modules/lib/open-amp/open-amp/lib/include/openamp/rpmsg_virtio.h:84:32: note: declared here
   84 | __deprecated static inline int deprecated_rpmsg_master(void)
      |                                ^~~~~~~~~~~~~~~~~~~~~~~
``` 
This allows backwards compatibility with the deprecated mechanism. and it seems simple enough to understand the problem by looking in the deprecated_rpmsg_master function.

If you are agree with this approach don't hesitate to squash the update in your patch.

edmooring: I like this approach. I have added it, and squashed it into my patch, with your Signed-off.

arnopo (approved):
LGTM, 
compliance ckeck  CI is broken, i will have a look

arnopo:
Compliance check fails because the build trace added in the commit message exceeds the maximum length. 
Re-splitting the lines would be at the expense of readability.

PR #328, opened 2021-12-13
rpmsg_virtio: configuration option to set buffer sizes per instance

### LINKED ISSUE #322: Make rpmsg buffer size customizable per rpmsg instance
I'm looking for a way to efficiently transfer data between cores in an embedded system. I need to transfer small messages between cores A and B, and in parallel transfer longer messages between cores A and C. OpenAMP meets my needs. I can create one instance between cores A and B, and another one between A and C.

The problem I'm facing is inefficient usage of shared memory between cores A and B. It would be enough for me to allocate 16 64-bytes buffers, but with constant `RPMSG_BUFFER_SIZE` value I need to allocate 16 512-bytes buffers.

I'm aware that `RPMSG_BUFFER_SIZE` can be modified compile-time, but it does not solve my problem, because for A-C instance I need `RPMSG_BUFFER_SIZE=512` while for A-B instance I would like to use `RPMSG_BUFFER_SIZE=64`.

I would like to propose similar solution to the one implemented in Linux https://elixir.bootlin.com/linux/v5.16-rc3/source/drivers/rpmsg/virtio_rpmsg_bus.c#L910 https://elixir.bootlin.com/linux/v5.16-rc3/source/drivers/rpmsg/virtio_rpmsg_bus.c#L61 and introduce `buf_size` field in `struct rpmsg_virtio_device` so that each instance of rpmsg could use different value of buffer sizes.

hubertmis:
Fixes #322 

Enable user of rpmsg_virtio to set sizes of TX and RX buffers per
created rpmsg_virtio instance. Each instance can use other buffer sizes.

Signed-off-by: Hubert Miś <hubert.mis@nordicsemi.no>

### DIFF
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -42,8 +42,30 @@ struct rpmsg_virtio_shm_pool {
 };
 
 /**
+ * struct rpmsg_virtio_config - configuration of rpmsg device based on virtio
+ *
+ * This structure is used by the rpmsg virtio host to configure the virtiio
+ * layer.
+ *
+ * @h2r_buf_size: the size of the buffer used to send data from host to remote
+ * @r2h_buf_size: the size of the buffer used to send data from remote to host
+ */
+struct rpmsg_virtio_config {
+	uint32_t h2r_buf_size;
+	uint32_t r2h_buf_size;
+};
+
+/* Default configuration */
+#define RPMSG_VIRTIO_DEFAULT_CONFIG                \
+	((const struct rpmsg_virtio_config) {       \
+		.h2r_buf_size = RPMSG_BUFFER_SIZE,  \
+		.r2h_buf_size = RPMSG_BUFFER_SIZE,  \
+	})
+
+/**
  * struct rpmsg_virtio_device - representation of a rpmsg device based on virtio
  * @rdev: rpmsg device, first property in the struct
+ * @config: structure containing virtio configuration
  * @vdev: pointer to the virtio device
  * @rvq: pointer to receive virtqueue
  * @svq: pointer to send virtqueue
@@ -52,6 +74,7 @@ struct rpmsg_virtio_shm_pool {
  */
 struct rpmsg_virtio_device {
 	struct rpmsg_device rdev;
+	struct rpmsg_virtio_config config;
 	struct virtio_device *vdev;
 	struct virtqueue *rvq;
 	struct virtqueue *svq;
@@ -144,6 +167,39 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 		    rpmsg_ns_bind_cb ns_bind_cb,
 		    struct metal_io_region *shm_io,
 		    struct rpmsg_virtio_shm_pool *shpool);
+
+/**
+ * rpmsg_init_vdev_with_config - initialize rpmsg virtio device with config
+ * Host side:
+ * Initialize RPMsg virtio queues and shared buffers, the address of shm can be
+ * ANY. In this case, function will get shared memory from system shared memory
+ * pools. If the vdev has the RPMsg name service feature, this API will create a
+ * name service endpoint.
+ * Sizes of virtio data buffers used by the initialized RPMsg instance are set
+ * to values read from the passed configuration structure.
+ *
+ * Remote side:
+ * This API will not return until the driver ready is set by the host side.
+ * Sizes of virtio data buffers are set by the host side. Values passed in the
+ * configuration structure have no effect.
+ *
+ * @param rvdev  - pointer to the rpmsg virtio device
+ * @param vdev   - pointer to the virtio device
+ * @param ns_bind_cb  - callback handler for name service announcement without
+ *                      local endpoints waiting to bind.
+ * @param shm_io - pointer to the share memory I/O region.
+ * @param shpool - pointer to shared memory pool. rpmsg_virtio_init_shm_pool has
+ *                 to be called first to fill this structure.
+ * @param config - pointer to configuration structure
+ *
+ * @return - status of function execution
+ */
+int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
+				struct virtio_device *vdev,
+				rpmsg_ns_bind_cb ns_bind_cb,
+				struct metal_io_region *shm_io,
+				struct rpmsg_virtio_shm_pool *shpool,
+				const struct rpmsg_virtio_config *config);
 
 /**
  * rpmsg_deinit_vdev - deinitialize rpmsg virtio device
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -143,14 +143,14 @@ int rpmsg_virtio_get_buffer_size(struct rpmsg_device *rdev);
 
 /**
  * rpmsg_init_vdev - initialize rpmsg virtio device
- * Master side:
+ * Host side:
  * Initialize RPMsg virtio queues and shared buffers, the address of shm can be
  * ANY. In this case, function will get shared memory from system shared memory
- * pools. If the vdev has RPMsg name service feature, this API will create an
+ * pools. If the vdev has the RPMsg name service feature, this API will create a
  * name service endpoint.
  *
- * Slave side:
- * This API will not return until the driver ready is set by the master side.
+ * Remote side:
+ * This API will not return until the driver ready is set by the host side.
  *
  * @param rvdev  - pointer to the rpmsg virtio device
  * @param vdev   - pointer to the virtio device
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -3,6 +3,7 @@
  * All rights reserved.
  * Copyright (c) 2016 Freescale Semiconductor, Inc. All rights reserved.
  * Copyright (c) 2018 Linaro, Inc. All rights reserved.
+ * Copyright (c) 2021 Nordic Semiconductor ASA
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -151,8 +152,8 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
 		data = virtqueue_get_buffer(rvdev->svq, len, idx);
 		if (!data && rvdev->svq->vq_free_cnt) {
 			data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
-							RPMSG_BUFFER_SIZE);
-			*len = RPMSG_BUFFER_SIZE;
+					rvdev->config.h2r_buf_size);
+			*len = rvdev->config.h2r_buf_size;
 			*idx = 0;
 		}
 	}
@@ -250,7 +251,7 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
 		 * If device role is Master then buffers are provided by us,
 		 * so just provide the macro.
 		 */
-		length = RPMSG_BUFFER_SIZE - sizeof(struct rpmsg_hdr);
+		length = rvdev->config.h2r_buf_size - sizeof(struct rpmsg_hdr);
 	}
 #endif /*!VIRTIO_SLAVE_ONLY*/
 
@@ -384,7 +385,7 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 
 #ifndef VIRTIO_SLAVE_ONLY
 	if (rpmsg_virtio_get_role(rvdev) == RPMSG_MASTER)
-		buff_len = RPMSG_BUFFER_SIZE;
+		buff_len = rvdev->config.h2r_buf_size;
 	else
 #endif /*!VIRTIO_SLAVE_ONLY*/
 		buff_len = virtqueue_get_buffer_length(rvdev->svq, idx);
@@ -611,6 +612,17 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 		    struct metal_io_region *shm_io,
 		    struct rpmsg_virtio_shm_pool *shpool)
 {
+	return rpmsg_init_vdev_with_config(rvdev, vdev, ns_bind_cb, shm_io,
+			   shpool, &RPMSG_VIRTIO_DEFAULT_CONFIG);
+}
+
+int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
+				struct virtio_device *vdev,
+				rpmsg_ns_bind_cb ns_bind_cb,
+				struct metal_io_region *shm_io,
+				struct rpmsg_virtio_shm_pool *shpool,
+				const struct rpmsg_virtio_config *config)
+{
 	struct rpmsg_device *rdev;
 	const char *vq_names[RPMSG_NUM_VRINGS];
 	vq_callback callback[RPMSG_NUM_VRINGS];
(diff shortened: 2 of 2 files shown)


hubertmis:
I've tested this implementation with a Zephyr sample. Looks to work as expected

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -41,10 +41,21 @@ struct rpmsg_virtio_shm_pool {
...
 };
 
+/**
+ * struct rpmsg_virtio_config - configuration of rpmsg device based on virtio
+ * @txbuf_size: the tx buffer size, used to send data from host to remote
+ * @rxbuf_size: the rx buffer size, used to send data from remote to host
+ */
+struct rpmsg_virtio_config {
+	uint32_t txbuf_size;
+	uint32_t rxbuf_size;
```
carlocaione, line 51: can we use `size_t` for sizes instead?

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +157,36 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 		    struct metal_io_region *shm_io,
 		    struct rpmsg_virtio_shm_pool *shpool);
 
+/**
+ * rpmsg_init_vdev_with_config - initialize rpmsg virtio device
```
carlocaione, line 161: `initialize rpmsg virtio device with config` ?

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +157,36 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+ * Master side:
+ * Initialize RPMsg virtio queues and shared buffers, the address of shm can be
+ * ANY. In this case, function will get shared memory from system shared memory
+ * pools. If the vdev has RPMsg name service feature, this API will create an
+ * name service endpoint.
+ *
+ * Slave side:
+ * This API will not return until the driver ready is set by the master side.
```
carlocaione, line 169: Can we add here what `config` is?

lib/include/openamp/rpmsg_virtio.h
xiaoxiang781216: It derived from PR https://github.com/OpenAMP/open-amp/pull/155. How size_t map to the real built-in type(32bit v.s. 64bit) depend on the platform. So if you consider in a complex SoC: one side is 32bit MCU/DSP, another side is AArch/RISV64/X86, size_t isn't a good choice here.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -630,6 +642,13 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 	rdev->ops.send_offchannel_nocopy = rpmsg_virtio_send_offchannel_nocopy;
 	role = rpmsg_virtio_get_role(rvdev);
 
+	if (config) {
+		rvdev->config = *config;
+	} else {
+		rvdev->config.txbuf_size = RPMSG_BUFFER_SIZE;
+		rvdev->config.rxbuf_size = RPMSG_BUFFER_SIZE;
```
carlocaione, line 649: Probably not but should we think about adding `RPMSG_BUFFER_RX_SIZE` and `RPMSG_BUFFER_TX_SIZE` as compile-time defines since we are now decoupling TX and RX?

hubertmis: I think if anyone needs it, it could be implemented in other patch

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -610,6 +611,17 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+	return rpmsg_init_vdev_with_config(rvdev, vdev, ns_bind_cb, shm_io,
+					   shpool, NULL);
+}
+
+int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
+				 struct virtio_device *vdev,
+				 rpmsg_ns_bind_cb ns_bind_cb,
+				 struct metal_io_region *shm_io,
+				 struct rpmsg_virtio_shm_pool *shpool,
+				 const struct rpmsg_virtio_config *config)
```
xiaoxiang781216, line 624: can we pass the rx/tx size directly to simplify the caller to define a struct rpmsg_virtio_config? and then change the function name to rpmsg_init_vdev_with_size.

lib/include/openamp/rpmsg_virtio.h
hubertmis: I agree with @xiaoxiang781216 it's not simple decision. The type should be selected based on the integration with users of this API. I would prefer to use `uint32_t` now, because it seems flexible and well-defined enough for most users. We can change the type later if we find other are more suitable.

lib/rpmsg/rpmsg_virtio.c
hubertmis: The definition I proposed is easily extensible if we find more configuration options to add in the future. We would just need to add more fields to the structure, instead of creating a new function for each potential configuration options permutations.

But if specialized init function for overriding sizes only is preferable, I can modify this implementation.

arnopo: Agree with @hubertmis, it would be better to keep a structure for future extension..
I just wonder about naming ( but no strong opinion) could we use "config" or "attributes"

arnopo: `const struct rpmsg_virtio_config *config `:
If the config is retrieved from the resource table could not be const. For instance regarding https://github.com/OpenAMP/open-amp/pull/155/

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +157,36 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+ *
+ * @param rvdev  - pointer to the rpmsg virtio device
+ * @param vdev   - pointer to the virtio device
+ * @param ns_bind_cb  - callback handler for name service announcement without
+ *                      local endpoints waiting to bind.
+ * @param shm_io - pointer to the share memory I/O region.
+ * @param shpool - pointer to shared memory pool. rpmsg_virtio_init_shm_pool has
+ *                 to be called first to fill this structure.
+ * @param config - pointer to configuration structure, if NULL default
+ *                 configuration is used like in @ref rpmsg_init_vdev
```
arnopo, line 179: confusing, proposal:
pointer to configuration structure, if NULL (like in @ref rpmsg_init_vdev) default configuration is used

arnopo (commented):
few minor remarks but LGTM

lib/include/openamp/rpmsg_virtio.h
carlocaione: agreed

lib/rpmsg/rpmsg_virtio.c
arnopo: > > If the config is retrieved from the resource table could not be const
> 
> `rpmsg_init_vdev_with_config()` does not modify the content of the structure. C allows assigning non-const pointers to const pointer types like:
> 
> ```
> struct rpmsg_virtio_config modifiable_config;
> modifiable_config.txbuf_size = condition ? 100 : 256;
> modifiable_config.rxbuf_size = 512;
> rpmsg_init_vdev_with_config(..., &modifiable_config);
> ```

Could you confirm this by a test?  Because a simple test on my side, adding const to a function parameter, generates a " discards ‘const’ qualifier" warning on build.


> 
> If `VIRTIO_RPMSG_F_BUFSZ` was implemented like in #155, I think it would modify the buffer `config` stored inside `struct rpmsg_virtio_device`, not the one passed as an argument to the initialization function (which is copied to `struct rpmsg_virtio_device` during initialization).
> 
> I don't see the reason to remove `const` from the new parameter type. Am I missing something?

Only Build warning and MISRA concerns.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -630,6 +642,12 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 	rdev->ops.send_offchannel_nocopy = rpmsg_virtio_send_offchannel_nocopy;
 	role = rpmsg_virtio_get_role(rvdev);
 
+	if (config) {
+		rvdev->config = *config;
+	} else {
+		rvdev->config = RPMSG_VIRTIO_DEFAULT_CONFIG;
```
arnopo, line 648: What about  just past RPMSG_VIRTIO_DEFAULT_CONFIG as argument in rpmsg_init_vdev function and suppress this condition?

hubertmis: We would need some code to handle `config == NULL`. What do you suggest? Return `RPMSG_ERR_PARAM`?

lib/rpmsg/rpmsg_virtio.c
hubertmis: I've tested it with a sample Zephyr application:
```
        struct rpmsg_virtio_config config = {
            .txbuf_size = 60,
            .rxbuf_size = 160,
        };
        err = rpmsg_init_vdev_with_config(&ctx->rvdev, &ctx->vdev, ns_bind_cb,
                      ctx->shm_io, &ctx->shpool, &config);
```
C99 6.3.2.3:
> For any qualifier q, a pointer to a non-q-qualified type may be converted to a pointer to the q-qualified version of the type; the values stored in the original and converted pointers shall compare equal.

MISRA C 8.13

What code are you using that generates a warning? I'm confused, because C standard allows this conversion.

arnopo: My apologies, my quick and dirty test adding a const was not good.
By applying your patch and updating system/linux/machine/generic/platform_info.c for testing: no warning.

carlocaione (approved):
LGTM

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -53,6 +71,7 @@ struct rpmsg_virtio_shm_pool {
 struct rpmsg_virtio_device {
 	struct rpmsg_device rdev;
 	struct virtio_device *vdev;
+	struct rpmsg_virtio_config config;
```
edmooring, line 74: Nitpick: Shouldn't this go just below rdev, putting the two structs before the various pointers? This generally leads to slightly better alignment.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +164,39 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 		    struct metal_io_region *shm_io,
 		    struct rpmsg_virtio_shm_pool *shpool);
 
+/**
+ * rpmsg_init_vdev_with_config - initialize rpmsg virtio device with config
+ * Master side:
```
edmooring, line 169: In keeping with the effort to remove potentially objectionable terminology, I suggest replacing "Master" and "Slave" with "Driver" and "Device" as in PR #323.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +164,39 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+ * Initialize RPMsg virtio queues and shared buffers, the address of shm can be
+ * ANY. In this case, function will get shared memory from system shared memory
+ * pools. If the vdev has RPMsg name service feature, this API will create an
+ * name service endpoint.
+ * Sizes of virtio data buffers used by initialized RPMsg instance are set to
+ * values read from the passed configuration structure.
+ *
+ * Slave side:
```
edmooring, line 177: As above, replace "Slave" with "Device". Check throughout the PR.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +164,39 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
 		    struct metal_io_region *shm_io,
 		    struct rpmsg_virtio_shm_pool *shpool);
 
+/**
+ * rpmsg_init_vdev_with_config - initialize rpmsg virtio device with config
+ * Master side:
+ * Initialize RPMsg virtio queues and shared buffers, the address of shm can be
+ * ANY. In this case, function will get shared memory from system shared memory
+ * pools. If the vdev has RPMsg name service feature, this API will create an
```
edmooring, line 172: Minor English fixups:
s/has RPMsg name service/has the RPMsg name service/
s/an$/a/

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -145,6 +164,39 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+ * name service endpoint.
+ * Sizes of virtio data buffers used by initialized RPMsg instance are set to
```
edmooring, line 174: s/initialized/the initialized/

edmooring (changes requested):
The substance of this looks good to go. Regarding the terminology changes, if they don't go in, I can pick them up in a subsequent push to #323, since I think this will probably go in first.

This pull request could serve as a model of how to do them right.

lib/include/openamp/rpmsg_virtio.h
hubertmis: I'm changing Master and Slave here to Host and Remote. It seems this header should use RPMsg terminology (host/remote), not virtio (driver/device), because it operates on rpmsg instances.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -610,17 +611,33 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
 	struct rpmsg_device *rdev;
 	const char *vq_names[RPMSG_NUM_VRINGS];
 	vq_callback callback[RPMSG_NUM_VRINGS];
 	int status;
 	unsigned int i, role;
 
+	if (config == NULL) {
+		return RPMSG_ERR_PARAM;
+	}
+
```
arnopo, line 635: here could be 

```
#ifndef VIRTIO_SLAVE_ONLY
	role = rpmsg_virtio_get_role(rvdev);
	if (role == RPMSG_MASTER) {
		if (config == NULL) {
			return RPMSG_ERR_PARAM;
		}
		rvdev->config = *config;
	}
#endif /*!VIRTIO_SLAVE_ONLY*/

```

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -41,9 +41,27 @@ struct rpmsg_virtio_shm_pool {
 	size_t size;
 };
 
+/**
+ * struct rpmsg_virtio_config - configuration of rpmsg device based on virtio
+ * @txbuf_size: the tx buffer size, used to send data from host to remote
+ * @rxbuf_size: the rx buffer size, used to send data from remote to host
```
arnopo, line 47: Deeply reviewing your patch, look to me that, for an unfamiliar user, it is not easy to understand this structure. 
My apologize for that because I should see it before... only the host use the config structure.
Another point is the notion of RX TX which is ambiguous. If only the host defines it, the remote gets the corresponding  values from the vrings.

Here is a proposal to rework the structure

```
 * struct rpmsg_virtio_config - configuration of rpmsg device based on virtio
 *
 * This structure is use by the rpmsg virtio host to configure the virtiio layer.
 * 
 * @h2r_buf_size: the buffer size, used to send data from host to remote
 * @r2h_buf_size: the uffer size, used to send data from remote to host
 struct rpmsg_virtio_config {
	uint32_t h2r_buf_size;
	uint32_t r2h_buf_size;
}
```

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -41,9 +41,27 @@ struct rpmsg_virtio_shm_pool {
...
+#define RPMSG_VIRTIO_DEFAULT_CONFIG                \
+	((const struct rpmsg_virtio_config) {       \
+		.txbuf_size = RPMSG_BUFFER_SIZE,    \
+		.rxbuf_size = RPMSG_BUFFER_SIZE,    \
+	})
+
 /**
  * struct rpmsg_virtio_device - representation of a rpmsg device based on virtio
  * @rdev: rpmsg device, first property in the struct
+ * @config: structure containing device configuration
```
arnopo, line 64: @config: structure containing virtio configuration

arnopo (changes requested):

lib/rpmsg/rpmsg_virtio.c
hubertmis: It's fine for me to update this way, because currently configuration applies only to the MASTER role.

I implemented updating config for both roles in case we extend it in the future with options applicable to both. But the implementation can be generalized later, when such option is created.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -610,13 +611,39 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+				struct rpmsg_virtio_shm_pool *shpool,
+				const struct rpmsg_virtio_config *config)
 {
 	struct rpmsg_device *rdev;
 	const char *vq_names[RPMSG_NUM_VRINGS];
 	vq_callback callback[RPMSG_NUM_VRINGS];
 	int status;
 	unsigned int i, role;
 
+	role = rpmsg_virtio_get_role(rvdev);
```
arnopo, line 632: Segmentation fault during my non-reg tests because the `rvdev->vdev->role` is not yet set. 
The block should be move after `rdev = &rvdev->rdev;`

arnopo (changes requested):

arnopo (approved):

hubertmis:
One more update: I've silenced "unused config parameter" warning when building SLAVE_ONLY configuration.

edmooring (approved):
Looks good to go.

PR #330, opened 2022-01-08
rpmsg_virtio: move RPMSG_VIRTIO_DEFAULT_CONFIG to private location

xiaoxiang781216:
and define it to NULL when VIRTIO_SLAVE_ONLY is defined to save space


arnopo (approved):
Much better in term of code optimization
Thanks!

edmooring (approved):
Looks good to go.

edmooring (approved):
Looks good to go.

PR #332, opened 2022-01-13
rpmsg: remove rpmsg_init_ept

xiaoxiang781216:
since it is deprecated for while


arnopo (approved):
LGTM, will be integrated only in release 2022.10

edmooring (approved):
Looks good to go for 2022.10.

Issue #335, opened 2022-01-25
remoteproc get memory function issue


joshualin-petaio:
Hi,

There has one question about get memory from remoteproc memory list.
Assume there has two memory region in the remoteproc memory list and continue.
like following description:
memory list 0 record start address (pa) : 0x00000000, size : 0x1000
memory list 1 record start address (pa) : 0x00001000, size : 0x1000
If I use remoteproc_get_io_with_pa() with pa : 0x00001000 to search memory list.
I will get memory list 0 because the get condition is match (pa + size) <= pa_end).

Does there have something wrong or need some precondition before create memory list ?

arnopo:
Hi,
Thank you for pointing this out. This need to be fixed.
could you send a PR to fix remoteproc_get_mem? to test pa < pa_end ( and same for da)?

joshualin-petaio:
Hi,

Thank you for your reply.
I had send a PR to fix remoteproc_get_mem().
(pa + size) < pa_end_end and also (da + size) < da_end.

BR,
Joshua

PR #337, opened 2022-01-25
fix remoteproc get wrong memory when memory region are continuous.

joshualin-petaio:


lib/remoteproc/remoteproc.c
```diff
@@ -55,14 +55,14 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 
 			pa_start = mem->pa;
 			pa_end = pa_start + mem->size;
-			if (pa >= pa_start && (pa + size) <= pa_end)
+			if (pa >= pa_start && (pa + size) < pa_end)
```
arnopo, line 58: (pa + size) = pa_end is a valid use case if size is not equal to 0.
That why i suggested you in https://github.com/OpenAMP/open-amp/issues/335 to return an error if pa >= pa_end

arnopo (changes requested):
Please rewrite the commit message (cf compliance CI fail).
A valid commit message contains:
- a text that give the rational of the patch
- a signed_of

lib/remoteproc/remoteproc.c
joshualin-petaio: Hi Arnopo,

Thank you for your explain.
The pa >= pa_end condition is really a best method.
I will modify it and send PR again.

Thanks,
Joshua

joshualin-petaio:
Sorry about that I close this PR.
There has some issue about rewrite commit.
I will re-send the PR include issue #335 and #338.
Thank you.

PR #339, opened 2022-01-28
fix remoteproc get wrong memory when memory region are continuous and Invalidate the ring region but use non initialize ring index.

joshualin-petaio:
1. fix remoteproc get wrong memory when memory region are continuous.
2. fix Invalidate the ring region but use non initialize ring index.

Signed-off-by: Joshua Lin <jlin@petaio.com>

### DIFF
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -55,14 +55,14 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 
 			pa_start = mem->pa;
 			pa_end = pa_start + mem->size;
-			if (pa >= pa_start && (pa + size) <= pa_end)
+			if (pa >= pa_start && (pa + size) <= pa_end && pa < pa_end)
 				return mem;
 		} else if (da != METAL_BAD_PHYS) {
 			metal_phys_addr_t da_start, da_end;
 
 			da_start = mem->da;
 			da_end = da_start + mem->size;
-			if (da >= da_start && (da + size) <= da_end)
+			if (da >= da_start && (da + size) <= da_end && da < da_end)
 				return mem;
 		} else if (va) {
 			if (metal_io_virt_to_offset(mem->io, va) !=
--- a/lib/virtio/virtqueue.c
+++ b/lib/virtio/virtqueue.c
@@ -271,10 +271,10 @@ void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
 
 	VQUEUE_BUSY(vq);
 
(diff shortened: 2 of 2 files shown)


arnopo:
Please re-split your commits.
1. you should have one commit per topic
2. your commit message should follow this form to pass compliance test:
```
 <framework>: <title>
 < commit message providing details on issue and solution>
 <signed-of>

``` 
For instance the first patch shoul contains something like this
```
remoteproc: fix remoteproc get wrong memory when memory region are continuous.

when pa is equal to the end address of the memory region,... (to complet)
Signed-off-by: Joshua Lin <jlin@petaio.com>
   
```
3. Don't add commit to fix your previous one(https://github.com/OpenAMP/open-amp/pull/339/commits/138a6d2c347dc2a410b591c3d28c22baef0d8f49). use `git push -f` to update your pull request

joshualin-petaio:
Hi Arnopo,

Thank you for your explain and recommend.
I had update the pull request.
Please check it.
And if still has something need to modify.
Please feel free to let me know.

Thank you.
Joshua

arnopo:
Hi @carlocaione 
Please, could you  have look to the cache management update
Thanks,
Arnaud

carlocaione (approved):
For the invalidation part

arnopo (changes requested):
Much better!
nitpicking over details else look good to me:
Some typos to correct in the commit message (I'm not a native English speaker, so I hope the proposals are valid).

- commit 1:

> remoteproc: fix remoteproc get wrong memory
> When memory region are continuous, it maybe get wrong memory region.

When memory regions are continuous, it may get wrong memory region.it may get

> If pa is equal to the end address of the memory and size equal zero.
> It will get wrong memory region.
> So needto add check pa is smaller then end address.

need to

> Signed-off-by: Joshua Lin <jlin@petaio.com>
 
- Commit 2:

> virtqueue: fix invalidate wrong cache region
> It is always invalidate the avail ring index 0.

It always invalidates

> It need get the right vring index before incalidate.

It must get the right vring index before invalidating it.

> Signed-off-by: Joshua Lin <jlin@petaio.com>

edmooring (changes requested):
The code looks good to me. 

I would like to see the commit message changed to something like:

The current code fetches the vring index before invalidating the cache. This can lead to using an invalid index. Fetch the index before invalidation to ensure a correct index.

Feel free to change this so that it matches reality.

joshualin-petaio:
Hi Edmooring:

Thank you for your recommend.
The message had modify done.
Please check it.
If has any recommend, please feel free to let me know.
Thanks, 
Joshua

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #347, opened 2022-02-07
openamp: add new ops wait_tx_buffer() support

GUIDINGLI:
openamp: add new ops wait_tx_buffer() support

### DIFF
--- a/lib/include/openamp/remoteproc.h
+++ b/lib/include/openamp/remoteproc.h
@@ -432,6 +432,16 @@ struct remoteproc_ops {
 					  metal_phys_addr_t da,
 					  void *va, size_t size,
 					  struct remoteproc_mem *buf);
+	/**
+	 * wait_tx_buffer
+	 *
+	 * Wait tx buffer available
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 *
+	 * return none
+	 */
+	int (*wait_tx_buffer)(struct remoteproc *rproc);
 };
 
 /* Remoteproc error codes */
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -109,6 +109,12 @@ rpmsg_virtio_create_virtqueues(struct rpmsg_virtio_device *rvdev,
 					callbacks);
 }
 
+static inline int rpmsg_virtio_wait_tx_buffer(struct rpmsg_virtio_device *rvdev)
+{
+	return rvdev->vdev->func->wait_tx_buffer ?
+			rvdev->vdev->func->wait_tx_buffer(rvdev->vdev) : -EAGAIN;
+}
+
 /**
  * rpmsg_virtio_get_buffer_size - get rpmsg virtio buffer size
  *
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -137,6 +137,7 @@ struct virtio_dispatch {
 			     void *src, int length);
 	void (*reset_device)(struct virtio_device *dev);
 	void (*notify)(struct virtqueue *vq);
+	int (*wait_tx_buffer)(struct virtio_device *dev);
 };
 
 int virtio_create_virtqueues(struct virtio_device *vdev, unsigned int flags,
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -30,6 +30,17 @@ static void rproc_virtio_virtqueue_notify(struct virtqueue *vq)
 	rpvdev->notify(rpvdev->priv, vring_info->notifyid);
 }
 
+static int rproc_virtio_wait_tx_buffer(struct virtio_device *vdev)
+{
+	struct remoteproc_virtio *rpvdev;
+	struct remoteproc *rproc;
+
+	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
+	rproc  = rpvdev->priv;
+
+	return rproc->ops->wait_tx_buffer ? rproc->ops->wait_tx_buffer(rproc) : -EAGAIN;
+}
+
 static unsigned char rproc_virtio_get_status(struct virtio_device *vdev)
 {
 	struct remoteproc_virtio *rpvdev;
@@ -179,6 +190,7 @@ static const struct virtio_dispatch remoteproc_virtio_dispatch_funcs = {
 	.get_features = rproc_virtio_get_features,
 	.read_config = rproc_virtio_read_config,
 	.notify = rproc_virtio_virtqueue_notify,
+	.wait_tx_buffer = rproc_virtio_wait_tx_buffer,
 #ifndef VIRTIO_SLAVE_ONLY
 	/*
 	 * We suppose here that the vdev is in a shared memory so that can
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -304,6 +304,72 @@ static void rpmsg_virtio_release_rx_buffer(struct rpmsg_device *rdev,
 	metal_mutex_release(&rdev->lock);
 }
 
+/**
+ * rpmsg_virtio_rx_callback
+ *
+ * Rx callback function.
+ *
+ * @param vq - pointer to virtqueue on which messages is received
+ *
+ */
+static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
+{
+	struct virtio_device *vdev = vq->vq_dev;
+	struct rpmsg_virtio_device *rvdev = vdev->priv;
+	struct rpmsg_device *rdev = &rvdev->rdev;
+	struct rpmsg_endpoint *ept;
+	struct rpmsg_hdr *rp_hdr;
+	uint32_t len;
+	uint16_t idx;
+	int status;
+
+	metal_mutex_acquire(&rdev->lock);
+
+	/* Process the received data from remote node */
+	rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
+
+	metal_mutex_release(&rdev->lock);
+
+	while (rp_hdr) {
+		rp_hdr->reserved = idx;
+
+		/* Get the channel node from the remote device channels list. */
+		metal_mutex_acquire(&rdev->lock);
+		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
+		metal_mutex_release(&rdev->lock);
+
+		if (ept) {
+			if (ept->dest_addr == RPMSG_ADDR_ANY) {
+				/*
+				 * First message received from the remote side,
+				 * update channel destination address
+				 */
+				ept->dest_addr = rp_hdr->src;
+			}
+			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
+					 rp_hdr->len, rp_hdr->src, ept->priv);
+
+			RPMSG_ASSERT(status >= 0,
+				     "unexpected callback status\r\n");
+		}
+
+		metal_mutex_acquire(&rdev->lock);
+
+		/* Check whether callback wants to hold buffer */
+		if (!(rp_hdr->reserved & RPMSG_BUF_HELD)) {
+			/* No, return used buffers. */
+			rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
+		}
+
+		rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
+		if (!rp_hdr) {
+			/* tell peer we return some rx buffer */
+			virtqueue_kick(rvdev->rvq);
+		}
+		metal_mutex_release(&rdev->lock);
+	}
+}
+
 static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 						uint32_t *len, int wait)
 {
@@ -333,6 +399,8 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		metal_mutex_release(&rdev->lock);
 		if (rp_hdr || !tick_count)
 			break;
+		if (rpmsg_virtio_wait_tx_buffer(rvdev) >= 0)
+			continue;
 		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
 		tick_count--;
 	}
@@ -456,72 +524,6 @@ static int rpmsg_virtio_send_offchannel_raw(struct rpmsg_device *rdev,
 static void rpmsg_virtio_tx_callback(struct virtqueue *vq)
 {
 	(void)vq;
-}
-
-/**
- * rpmsg_virtio_rx_callback
- *
- * Rx callback function.
- *
- * @param vq - pointer to virtqueue on which messages is received
- *
- */
-static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
-{
-	struct virtio_device *vdev = vq->vq_dev;
-	struct rpmsg_virtio_device *rvdev = vdev->priv;
-	struct rpmsg_device *rdev = &rvdev->rdev;
-	struct rpmsg_endpoint *ept;
-	struct rpmsg_hdr *rp_hdr;
-	uint32_t len;
-	uint16_t idx;
-	int status;
-
-	metal_mutex_acquire(&rdev->lock);
-
-	/* Process the received data from remote node */
-	rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
-
-	metal_mutex_release(&rdev->lock);
-
-	while (rp_hdr) {
-		rp_hdr->reserved = idx;
-
-		/* Get the channel node from the remote device channels list. */
-		metal_mutex_acquire(&rdev->lock);
-		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
-		metal_mutex_release(&rdev->lock);
-
-		if (ept) {
-			if (ept->dest_addr == RPMSG_ADDR_ANY) {
-				/*
-				 * First message received from the remote side,
-				 * update channel destination address
-				 */
-				ept->dest_addr = rp_hdr->src;
-			}
-			status = ept->cb(ept, RPMSG_LOCATE_DATA(rp_hdr),
-					 rp_hdr->len, rp_hdr->src, ept->priv);
-
-			RPMSG_ASSERT(status >= 0,
-				     "unexpected callback status\r\n");
-		}
-
-		metal_mutex_acquire(&rdev->lock);
-
-		/* Check whether callback wants to hold buffer */
-		if (!(rp_hdr->reserved & RPMSG_BUF_HELD)) {
-			/* No, return used buffers. */
-			rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
-		}
-
-		rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
-		if (!rp_hdr) {
-			/* tell peer we return some rx buffer */
-			virtqueue_kick(rvdev->rvq);
-		}
-		metal_mutex_release(&rdev->lock);
-	}
 }
 
 /**


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -333,6 +333,8 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		metal_mutex_release(&rdev->lock);
```
edmooring, line 333: Do we need to hold the lock, or re-lock the device during rpmsg_virtio_wait_tx_buffer()?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -333,6 +333,8 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
...
 		if (rp_hdr || !tick_count)
 			break;
+		if (rpmsg_virtio_wait_tx_buffer(rvdev) >= 0)
```
edmooring, line 336: This looks wrong. If it fails, the program waits twice, once in rpmsg_virtio_wait_tx_buffer(), and then in metal_sleep_usec().

edmooring (changes requested):
This needs a lot more documentation explaining what the new operation must do, and what values it returns.

lib/rpmsg/rpmsg_virtio.c
GUIDINGLI: No need, you can assume the simplest user's implementation of wait_tx_buffer() is metal_sleep_usec().

lib/rpmsg/rpmsg_virtio.c
GUIDINGLI: If wait fails, that means hasn't wait, then we still need do metal_sleep_usec(), or will busyloop.

lib/include/openamp/remoteproc.h
```diff
@@ -432,6 +432,16 @@ struct remoteproc_ops {
...
+	/**
+	 * wait_tx_buffer
+	 *
+	 * Wait for a tx buffer to become available.
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 *
+	 * return 0 means there is a tx buffer available, otherwise negative value.
+	 */
+	int (*wait_tx_buffer)(struct remoteproc *rproc);
```
arnopo, line 444: This API is for the rpmsg protocol not the remoteproc part

lib/rpmsg/rpmsg_virtio.c
arnopo: Please could you explain why you don't use `rpmsg_trysend` API instead  that set `wait` to `false` and so `tick_count` to `0`?

arnopo (changes requested):
if `rpmsg_trysend` can do the job please detail what is missing

lib/rpmsg/rpmsg_virtio.c
GUIDINGLI: There is only one difference between a looping query mechanism and a waiting mechanism.
Of course, a waiting mechanism is more suitable for low power device.

arnopo: My point here is that I don't understand why you don't use "rpmsg_trysend" in an application loop instead.
you would generate your loop without waiting mechanism.

GUIDINGLI:
There is only one difference between a looping query mechanism and a waiting mechanism.
Of course, a waiting mechanism is more suitable for low power device.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: But,  rpmsg_send/rpmsg_get_tx_payload_buffer is a public api, we can't limit how other team write the application.

edmooring: > If wait fails, that means hasn't wait, then we still need do metal_sleep_usec(), or will busyloop.

So a function with "wait" in its name doesn't wait sometimes? This needs to be documented in the Doxygen comments. There needs to be enough documentation that a new user of the library can understand why they might need to implement this operation, and how to implement it properly.

edmooring: > My point here is that I don't understand why you don't use "rpmsg_trysend" in an application loop instead.
> you would generate your loop without waiting mechanism.
I agree with @arnopo. I don't understand why this needs to be in the library, rather than in application code.

edmooring:
An operation that simply checked if there is a tx buffer available could be useful to application writers. For example, there could be a special thread that monitored buffer usage, and if there were no buffers available, did some sort of maintenance on the buffer pool, or restarted the application. The point is that the handling of the out of buffers situation is best done by the application programmer.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: > > If wait fails, that means hasn't wait, then we still need do metal_sleep_usec(), or will busyloop.
> 

Yes, wait fail mean it can't finish the work in some special case, so the common code need do the fallback action. In most case, wait success mean the buffer available and the framework can try again immediately.

> So a function with "wait" in its name doesn't wait sometimes? This needs to be documented in the Doxygen comments. There needs to be enough documentation that a new user of the library can understand why they might need to implement this operation, and how to implement it properly.

Yes.

xiaoxiang781216:
> An operation that simply checked if there is a tx buffer available could be useful to application writers.

It's still a busy loop.

> For example, there could be a special thread that monitored buffer usage, and if there were no buffers available, did some sort of maintenance on the buffer pool, or restarted the application. The point is that the handling of the out of buffers situation is best done by the application programmer.

Are you sure this should be done by the application programmer, not BSP programmer?!
The team write rpmsg service is different from the team port OpenAMP to the new hardware. How to know the buffer available just like how to receive the remote request is the platform/hardware dependent. On the other hand, the service team could write a general rpmsg service which can work on any hardware platform with the correct OpenAMP porting.

I remember that there is a work group to define the general rpmsg service, could you tell me how the work group can avoid the busy loop in a platform dependent way by your propose.

xiaoxiang781216:
From the architecture perspective, the busy polling can be replaced by waiting a semaphore in OpenAMP framework, just like how Linux side do the same thing, but this require to change libmetal and open-amp. From our upstream experience, all my team member don't want to make any big change to libmetal or open-amp.

arnopo:
> From the architecture perspective, the busy polling should be replaced by waiting a semaphore in OpenAMP framework, just like how Linux side do the same thing, but this require to change libmetal and open-amp. From our upstream experience, all my team member don't want to make any big change to libmetal or open-amp.

Behavior depends on:
- OS : linux offers capability of an active wait creating a work queue, bare metal and some OS can do that. In bare metal using rpmsg_send is not a good idea as it will block all the system.
- hardware: Polling vrings without any inter processor signaling is an implementation that makes semaphore unusable. 
- rpmsg services: in multi service mode, some services can want to sleep waiting TX buffer is available, some want  to inform application, some wants to have the priority, ....
Supporting all platforms, OSes, services  usecases in the library itself seems not a good idea as too complex.

As you mentioned a rpmsg service layer is something that makes sense in OSes. But this layer is quite OS/project dependent. Make it generic would also introduce new design tradeoffs.
This layer exists in [Zephyr](https://github.com/zephyrproject-rtos/zephyr/tree/main/subsys/ipc/rpmsg_service).
Look like implementing such kind of layer in NuttX would allow you to avoid such cross dependencies, managing this use case and https://github.com/OpenAMP/open-amp/pull/238, with out impacting other OpenAMP lib users.

Focusing on the PR itself:
What about exposing  `RPMSG_TICKS_PER_INTERVAL` as project config,  has done  for `RPMSG_BUFFER_SIZE`?
If set to 0 => polling with no wait.

xiaoxiang781216:
> > From the architecture perspective, the busy polling should be replaced by waiting a semaphore in OpenAMP framework, just like how Linux side do the same thing, but this require to change libmetal and open-amp. From our upstream experience, all my team member don't want to make any big change to libmetal or open-amp.
> 
> Behavior depends on:
> 
> * OS : linux offers capability of an active wait creating a work queue, bare metal and some OS can do that. In bare metal using rpmsg_send is not a good idea as it will block all the system.
> * hardware: Polling vrings without any inter processor signaling is an implementation that makes semaphore unusable.
> * rpmsg services: in multi service mode, some services can want to sleep waiting TX buffer is available, some want  to inform application, some wants to have the priority, ....
>   Supporting all platforms, OSes, services  usecases in the library itself seems not a good idea as too complex.
> 

Yes, that's why this patch provide the callback to give the porting layer a chance to handle this situation in a better way other than the poll loop.

> As you mentioned a rpmsg service layer is something that makes sense in OSes. But this layer is quite OS/project dependent. Make it generic would also introduce new design tradeoffs. This layer exists in [Zephyr](https://github.com/zephyrproject-rtos/zephyr/tree/main/subsys/ipc/rpmsg_service). Look like implementing such kind of layer in NuttX would allow you to avoid such cross dependencies, managing this use case and #238, with out impacting other OpenAMP lib users.
> 

Yes, but the framework shouldn't forbid the porting layer to do some optimization if they want.
“Programming Perl” give a good suggestion about this topic:
Easy things should be easy, and hard things should be possible.

> Focusing on the PR itself: What about exposing `RPMSG_TICKS_PER_INTERVAL` as project config, has done for `RPMSG_BUFFER_SIZE`? If set to 0 => polling with no wait.

First, we want to remove all poll loop for battery powered device. Second, the callback give many possibility other than simply sleep(e.g. wait semaphore, recursive dispatch...).

arnopo:
More an library integration point that a feature point. From feature point of view be able to optimize the waiting is completly valid, that's why the rpmsg_trysend had been created.
The main advantage of your proposal is to simplify the support of different virtio backend. remoteproc_virtio is the default one some other such as static vrings(Zephyr) or virtio mmio can be used.

If @edmooring  is ok let's move forward on your proposed implementation and see how smoothly this could be inserted.

arnopo (changes requested):

edmooring:
> > An operation that simply checked if there is a tx buffer available could be useful to application writers.
> 
> It's still a busy loop.
> 
> > For example, there could be a special thread that monitored buffer usage, and if there were no buffers available, did some sort of maintenance on the buffer pool, or restarted the application. The point is that the handling of the out of buffers situation is best done by the application programmer.
> 
> Are you sure this should be done by the application programmer, not BSP programmer?! The team write rpmsg service is different from the team port OpenAMP to the new hardware. How to know the buffer available just like how to receive the remote request is the platform/hardware dependent. On the other hand, the service team could write a general rpmsg service which can work on any hardware platform with the correct OpenAMP porting.

Yes, because the behavior that is needed can change, depending on the application. In many cases, the application/service developers work for a different company than the team that develops the BSP, and there are dozens of different teams from multiple organizations using the same BSP.

> 
> I remember that there is a work group to define the general rpmsg service, could you tell me how the work group can avoid the busy polling in a platform dependent way by your propose.

As @arnopo points out down below, this can be done at a higher level than the OpenAMP transport layer. Attempting to handle application-level policy decisions at the transport layer risks causing more problems later.

edmooring:
> > An operation that simply checked if there is a tx buffer available could be useful to application writers.
> 
> It's still a busy loop.
> 
> > For example, there could be a special thread that monitored buffer usage, and if there were no buffers available, did some sort of maintenance on the buffer pool, or restarted the application. The point is that the handling of the out of buffers situation is best done by the application programmer.
> 
> Are you sure this should be done by the application programmer, not BSP programmer?! The team write rpmsg service is different from the team port OpenAMP to the new hardware. How to know the buffer available just like how to receive the remote request is the platform/hardware dependent. On the other hand, the service team could write a general rpmsg service which can work on any hardware platform with the correct OpenAMP porting.

Yes, because the behavior that is needed can change, depending on the application. In many cases, the application/service developers work for a different company than the team that develops the BSP, and there are dozens of different teams from multiple organizations using the same BSP.

> 
> I remember that there is a work group to define the general rpmsg service, could you tell me how the work group can avoid the busy polling in a platform dependent way by your propose.

As @arnopo points out down below, this can be done at a higher level than the OpenAMP transport layer. Attempting to handle application-level policy decisions at the transport layer risks causing more problems later.

lib/include/openamp/remoteproc.h
```diff
@@ -433,6 +433,17 @@ struct remoteproc_ops {
 					  metal_phys_addr_t da,
 					  void *va, size_t size,
 					  struct remoteproc_mem *buf);
+	/**
+	 * wait_notified
+	 *
+	 * Wait for remote notified, when there is no TX buffer anymore.
+	 * Set to NULL means use usleep to TX buffer available.
```
xiaoxiang781216, line 440: Set to NULL means use usleep to wait TX buffer available.

lib/include/openamp/remoteproc.h
```diff
@@ -433,6 +433,17 @@ struct remoteproc_ops {
...
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 *
+	 * return 0 means there is notify available, otherwise negative value.
+	 */
+	int (*wait_notified)(struct remoteproc *rproc);
```
xiaoxiang781216, line 446: add uint32_t id

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -22,6 +22,7 @@ extern "C" {
 
 /* define vdev notification function user should implement */
 typedef int (*rpvdev_notify_func)(void *priv, uint32_t id);
+typedef int (*rpvdev_wait_notified_func)(void *priv);
```
xiaoxiang781216, line 25: reuse rpvdev_notify_func?

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -37,6 +38,7 @@ struct remoteproc_virtio {
 	void *vdev_rsc;
 	struct metal_io_region *vdev_rsc_io;
 	rpvdev_notify_func notify;
+    rpvdev_wait_notified_func wait_notified;
```
xiaoxiang781216, line 41: align

lib/include/openamp/virtio.h
```diff
@@ -137,6 +137,7 @@ struct virtio_dispatch {
 			     void *src, int length);
 	void (*reset_device)(struct virtio_device *dev);
 	void (*notify)(struct virtqueue *vq);
+	int (*wait_notified)(struct virtio_device *dev);
```
xiaoxiang781216, line 140: virtio_device ->virtqueue

lib/include/openamp/rpmsg.h
```diff
@@ -42,6 +42,7 @@ extern "C" {
 #define RPMSG_ERR_BUFF_SIZE		(RPMSG_ERROR_BASE - 5)
 #define RPMSG_ERR_INIT			(RPMSG_ERROR_BASE - 6)
 #define RPMSG_ERR_ADDR			(RPMSG_ERROR_BASE - 7)
+#define RPMSG_ERR_NXIO			(RPMSG_ERROR_BASE - 8)
```
xiaoxiang781216, line 46: NXIO->NO_SUPPORT?

GUIDINGLI:
@arnopo as your suggestion, modify `rproc_virtio_create_vdev`, but now zephyr compile failed, how could I do ?

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -22,6 +22,7 @@ extern "C" {
 
 /* define vdev notification function user should implement */
 typedef int (*rpvdev_notify_func)(void *priv, uint32_t id);
+typedef int (*rpvdev_wait_notified_func)(void *priv, uint32_t id);
```
xiaoxiang781216, line 25: can we change to rpvdev_notify_wait_func to align with rpvdev_notify_func?

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -37,6 +38,7 @@ struct remoteproc_virtio {
 	void *vdev_rsc;
 	struct metal_io_region *vdev_rsc_io;
 	rpvdev_notify_func notify;
+	rpvdev_wait_notified_func wait_notified;
```
xiaoxiang781216, line 41: wait_notified to notify_wait

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -140,6 +140,15 @@ rpmsg_virtio_create_virtqueues(struct rpmsg_virtio_device *rvdev,
 					callbacks);
 }
 
+static inline int
+rpmsg_virtio_wait_notified(struct rpmsg_virtio_device *rvdev,
```
xiaoxiang781216, line 144: rpmsg_virtio_notify_wait

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -140,6 +140,15 @@ rpmsg_virtio_create_virtqueues(struct rpmsg_virtio_device *rvdev,
...
+					struct virtqueue *vq)
```
xiaoxiang781216, line 145: alignment

lib/include/openamp/virtio.h
```diff
@@ -162,6 +162,7 @@ struct virtio_dispatch {
 			     void *src, int length);
 	void (*reset_device)(struct virtio_device *dev);
 	void (*notify)(struct virtqueue *vq);
+	int (*wait_notified)(struct virtio_device *dev, struct virtqueue *vq);
```
xiaoxiang781216, line 165: notify_wait(struct virtqueue *vq);

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -179,6 +192,7 @@ static const struct virtio_dispatch remoteproc_virtio_dispatch_funcs = {
 	.get_features = rproc_virtio_get_features,
 	.read_config = rproc_virtio_read_config,
 	.notify = rproc_virtio_virtqueue_notify,
+	.wait_notified = rproc_virtio_wait_notified,
```
xiaoxiang781216, line 195: .notify_wait = rproc_virtio_virtqueue_notify_wait

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -30,6 +30,19 @@ static void rproc_virtio_virtqueue_notify(struct virtqueue *vq)
 	rpvdev->notify(rpvdev->priv, vring_info->notifyid);
 }
 
+static int rproc_virtio_wait_notified(struct virtio_device *vdev,
+				struct virtqueue *vq)
```
xiaoxiang781216, line 34: static int rproc_virtio_notify_wait(struct virtqueue *vq)

lib/remoteproc/remoteproc.c
```diff
@@ -880,6 +880,16 @@ static int remoteproc_virtio_notify(void *priv, uint32_t id)
 	return 0;
 }
 
+static int remoteproc_virtio_wait_notified(void *priv, uint32_t id)
```
xiaoxiang781216, line 883: remoteproc_virtio_notify_wait

lib/include/openamp/remoteproc.h
```diff
@@ -433,6 +433,18 @@ struct remoteproc_ops {
...
+	 *
+	 * Wait for remote notified, when there is no TX buffer anymore.
+	 * Set to NULL means use usleep to wait TX buffer available.
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 * @id    - the notifyid
+	 *
+	 * return 0 means there is notify available, otherwise negative value.
+	 */
+	int (*wait_notified)(struct remoteproc *rproc, uint32_t id);
```
xiaoxiang781216, line 447: notify_wait> should we move before line 414?

GUIDINGLI:
@edmooring @arnopo @jjmcdn
Could help review this ?

GUIDINGLI:
@edmooring @arnopo @jjmcdn
Could help review this ?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -347,6 +347,13 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		metal_mutex_release(&rdev->lock);
 		if (rp_hdr || !tick_count)
 			break;
+
+		status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
```
arnopo, line 351: create a more generic function that implement also the metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
something such as:

```
static inline int
rpmsg_virtio_wait_for_buffer(struct rpmsg_virtio_device *rvdev,
			 struct virtqueue *vq, int timeout)
{
	if (rvdev->vdev->func->notify_wait) {
			return rvdev->vdev->func->notify_wait(rvdev->vdev, vq, timeout) :
	} else {
			metal_sleep_usec(timeout);
			return RPMSG_SUCCESS;
	}
}
```

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -140,6 +140,15 @@ rpmsg_virtio_create_virtqueues(struct rpmsg_virtio_device *rvdev,
 					callbacks);
 }
 
+static inline int
+rpmsg_virtio_notify_wait(struct rpmsg_virtio_device *rvdev,
```
arnopo, line 144: any need to expose this API to the users?

lib/include/openamp/remoteproc.h
```diff
@@ -411,6 +411,18 @@ struct remoteproc_ops {
...
+	 *
+	 * Wait for remote notified, when there is no TX buffer anymore.
+	 * Set to NULL means use usleep to wait TX buffer available.
+	 *
+	 * @rproc - pointer to remoteproc instance
+	 * @id    - the notifyid
+	 *
+	 * return 0 means there is notify available, otherwise negative value.
+	 */
+	int (*notify_wait)(struct remoteproc *rproc, uint32_t id);
```
arnopo, line 425: Here we don't wait that the remote is notified, but that the local is notified when the remote releases a buffer. 
So we wait for remote notification. your previous definition `wait_notified`or seems to me more explicit (  comment is valid for the whole commit)

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -22,6 +22,7 @@ extern "C" {
 
 /* define vdev notification function user should implement */
 typedef int (*rpvdev_notify_func)(void *priv, uint32_t id);
+typedef int (*rpvdev_notify_wait)(void *priv, uint32_t id);
```
arnopo, line 25: "rpvdev_wait_vring_update" seems more explicit to me

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -179,6 +192,7 @@ static const struct virtio_dispatch remoteproc_virtio_dispatch_funcs = {
 	.get_features = rproc_virtio_get_features,
 	.read_config = rproc_virtio_read_config,
 	.notify = rproc_virtio_virtqueue_notify,
+	.notify_wait = rproc_virtio_notify_wait,
```
arnopo, line 195: As mentioned the feature is optional. seems better to declare it in `remoteproc_virtio` (similar to rpvdev_notify_func) and add a new API to set it.

arnopo (changes requested):

lib/rpmsg/rpmsg_virtio.c
GUIDINGLI: we need pass param timeout to rpmsg_virtio_wait_for_buffer, and need `tick_count--`.
So I change the caller.

lib/include/openamp/rpmsg_virtio.h
GUIDINGLI: Don't need

lib/include/openamp/remoteproc.h
GUIDINGLI: done

lib/remoteproc/remoteproc_virtio.c
GUIDINGLI: It is a element of `remoteproc_ops`, if user don't want this feature, then don't set the `remoteproc_ops->wait_notified`

arnopo: > It is a element of `remoteproc_ops`, if user don't want this feature, then don't set the `remoteproc_ops->wait_notified`

That's true only if the remoteproc_ops is used. Here is an exemple of the use of rpmsg without remoteproc_ops:
https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/ipc/openamp_rsc_table/src/main_remote.c#L228

CV-Bowen: @arnopo Hi, I want to continue this PR. You mean add new api to set the `rpvdev->wait_notifed` is add a new function called `rproc_virtio_set_wait_notifed()` or add the new argument in` rproc_virtio_create_vdev()` like this:
```
struct virtio_device *
rproc_virtio_create_vdev(unsigned int role, unsigned int notifyid,
			 void *rsc, struct metal_io_region *rsc_io,
			 void *priv,
			 rpvdev_notify_func notify,
                         rpvdev_wait_notify_func wait_notifed,
			 virtio_dev_reset_cb rst_cb)
```
Both are good for me, but I think we should set the notify and wait_notifed function in same way.

arnopo: Hi @CV-Bowen ,

I will need probably few time to remind all discussions on this PR, but concerning our point:

We are trying  to not modify the existing API. As the wait_notifed is optional, I prefer the `rproc_virtio_set_wait_notifed()` API. 

Please notice also that we added more Doxygen documentation, so this PR needs rebase and some refresh .

CV-Bowen:
@arnopo Hi, I add a new PR https://github.com/OpenAMP/open-amp/pull/518 for this feature, and did some modifications based on your comments in this PR. Could you take a look?

arnopo:
I close this one to avoid confusion.

PR #364, opened 2022-03-28
Removed unused code from rpmsg_get_endpoint

tammyleino:
Case for src and dest address is unused since the first case matches src only.


arnopo (approved):
LGTM,
If this test is needed in future, this could be reverted

edmooring (approved):
Looks good to me.

PR #365, opened 2022-03-28
Enable OpenAMP Demo configuration via CMake

bentheredonethat:
Currently the demos' interrupt, shared memory and IPC structure values are hard-coded.

Allow these to configured via CMake configure by 
(a) wrapping C symbols in #ifdef / #ifndef
(b) enabling linker script defined resource table location to be set in CMake.

### DIFF
--- a/apps/machine/zynqmp/platform_info.c
+++ b/apps/machine/zynqmp/platform_info.c
@@ -36,12 +36,20 @@
                                * one CPU in the exmaple. We set the CPU
                                * index to 0. */
 #ifdef versal
+#ifndef IPI_CHN_BITMASK
 #define IPI_CHN_BITMASK     0x08 /* IPI channel bit mask for IPI
 					* from/to RPU0 */
+#endif /* !IPI_CHN_BITMASK */
+#ifndef IPI_DEV_NAME
 #define IPI_DEV_NAME        "ff360000.ipi" /* IPI device name */
+#endif /* !IPI_DEV_NAME */
 #else
+#ifndef IPI_CHN_BITMASK
 #define IPI_CHN_BITMASK	    0x00000100
+#endif /* !IPI_CHN_BITMASK */
+#ifndef IPI_DEV_NAME
 #define IPI_DEV_NAME	    "ff340000.ipi"
+#endif /* !IPI_DEV_NAME */
 #endif /* versal */
 #define DEV_BUS_NAME        "platform" /* device bus name. "platform" bus
                                         * is used in Linux kernel for generic
@@ -49,15 +57,27 @@
 /* libmetal devices names used in the examples.
  * They are platform devices, you find them in Linux sysfs
  * sys/bus/platform/devices */
+#ifndef SHM_DEV_NAME
 #define SHM_DEV_NAME        "3ed20000.shm" /* shared device name */
-
+#endif /* SHM_DEV_NAME */
+#ifndef RSC_MEM_PA
 #define RSC_MEM_PA          0x3ED20000UL
+#endif /* !RSC_MEM_PA */
+#ifndef RSC_MEM_SIZE
 #define RSC_MEM_SIZE        0x2000UL
+#endif /* !RSC_MEM_SIZE */
+#ifndef VRING_MEM_PA
 #define VRING_MEM_PA        0x3ED40000UL
+#endif /* !VRING_MEM_PA */
+#ifndef VRING_MEM_SIZE
 #define VRING_MEM_SIZE      0x8000UL
+#endif /* !VRING_MEM_SIZE */
+#ifndef SHARED_BUF_PA
 #define SHARED_BUF_PA       0x3ED48000UL
+#endif /* !SHARED_BUF_PA */
+#ifndef SHARED_BUF_SIZE
 #define SHARED_BUF_SIZE     0x40000UL
-
+#endif /* !SHARED_BUF_SIZE */
 
 struct remoteproc_priv rproc_priv = {
 	.shm_name = SHM_DEV_NAME,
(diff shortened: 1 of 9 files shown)


apps/CMakeLists.txt
```diff
@@ -8,6 +8,11 @@ set (APPS_ROOT_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
 
 set (APPS_SHARE_DIR "${CMAKE_CURRENT_BINARY_DIR}/share")
 
+# If not provided, add default value for resource table
+if(NOT DEFINED ${RSC_TABLE})
+  set(RSC_TABLE 0x3ed20000 CACHE STRING "")
+endif(NOT DEFINED ${RSC_TABLE})
+
```
arnopo, line 15: This part is related to the machine, the resource table is optional

arnopo (changes requested):

apps/CMakeLists.txt
bentheredonethat: ok thanks for the review. is it if we keep this but move it to the machine-specific CMakeLists.txt files?

bentheredonethat: updated so that now this logic is only in the machine

arnopo (changes requested):

bentheredonethat:
@arnopo  updated so that only linker scripts and the 2 CMakeLists.txt files are updated. thanks

apps/machine/zynqmp_r5/CMakeLists.txt
```diff
@@ -3,3 +3,7 @@ collect (APP_COMMON_SOURCES rsc_table.c)
 collect (APP_COMMON_SOURCES zynqmp_r5_a53_rproc.c)
 collect (APP_INC_DIRS "${CMAKE_CURRENT_SOURCE_DIR}")
 
+# If not provided, add default value for resource table
+if(NOT DEFINED ${RSC_TABLE})
+  set(RSC_TABLE 0x3ed20000 CACHE STRING "")
+endif(NOT DEFINED ${RSC_TABLE})
```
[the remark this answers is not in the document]
bentheredonethat: moved to the file you suggestd

apps/examples/echo/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ foreach (_app rpmsg-echo-ping rpmsg-echo)
       add_executable (${_app}.out ${_sources})
       set_source_files_properties(${_sources} PROPERTIES COMPILE_FLAGS "${_cflags}")
 
-      target_link_libraries(${_app}.out -Wl,-Map=${_app}.map -Wl,--gc-sections ${_linker_opt} -Wl,--start-group ${OPENAMP_LIB}-static ${_deps} -Wl,--end-group)
+      target_link_libraries(${_app}.out -Wl,-Map=${_app}.map  -Wl,--defsym,_rsc_table=${RSC_TABLE} -Wl,--gc-sections ${_linker_opt} -Wl,--start-group ${OPENAMP_LIB}-static ${_deps} -Wl,--end-group)
```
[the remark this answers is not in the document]
bentheredonethat: updated so that it is in this file

apps/machine/zynqmp/platform_info.c
```diff
@@ -36,28 +36,48 @@
...
 #define DEV_BUS_NAME        "platform" /* device bus name. "platform" bus
                                         * is used in Linux kernel for generic
 					* devices */
 /* libmetal devices names used in the examples.
  * They are platform devices, you find them in Linux sysfs
  * sys/bus/platform/devices */
+#ifndef SHM_DEV_NAME
 #define SHM_DEV_NAME        "3ed20000.shm" /* shared device name */
-
+#endif /* SHM_DEV_NAME */
```
arnopo, line 62: Do you really need to give a name that depend on the address?
Same for POLL_DEV_NAME and IPI_DEV_NAME

arnopo (approved):
I wonder if the application folder should not be in a separate repo as code is for demo and test and doesn't request same level of maturity.
@edmooring what is your feeling on this?

That's said for time being, LGTM with just a comment/question

apps/machine/zynqmp/platform_info.c
bentheredonethat: for linux a name is needed as the device is accessed via UIO and device tree node name

edmooring:
> I wonder if the application folder should not be in a separate repo as code is for demo and test and doesn't request same level of maturity. @edmooring what is your feeling on this?
> 
> That's said for time being, LGTM with just a comment/question

I think having the apps in a separate repository will make it harder for newcomers to work with the code. Their lives are hard enough already:)

apps/system/generic/machine/zynqmp_r5/linker_large_text.ld
```diff
@@ -312,10 +312,10 @@ _SDA2_BASE_ = __sdata2_start + ((__sbss2_end - __sdata2_start) / 2 );
    __undef_stack = .;
 } > psu_r5_tcm_ram_0_S_AXI_BASEADDR
 
-.resource_table 0x3ed20000 : {
+.resource_table _rsc_table : {
 	. = ALIGN(4);
 	*(.resource_table)
-} > psu_ddr_S_AXI_BASEADDR
+}
```
edmooring, line 318: Why remove the ' > psu_ddr_S_AXI_BASEADDR'?

bentheredonethat: its possible that the demo is having IPC structures outside of DDR. For example TCM.

edmooring (approved):
Looks OK to me. Seems to work.

PR #368, opened 2022-04-12
virtqueue_add_buffer must check that there is room for multiple buffers

tammyleino:
Ensure there is enough room for all the buffers being added to the vring.
Signed-off-by: Tammy Leino tammy_leino@mentor.com

### DIFF
--- a/lib/virtio/virtqueue.c
+++ b/lib/virtio/virtqueue.c
@@ -131,7 +131,7 @@ int virtqueue_add_buffer(struct virtqueue *vq, struct virtqueue_buf *buf_list,
 
 	VQ_PARAM_CHK(vq == NULL, status, ERROR_VQUEUE_INVLD_PARAM);
 	VQ_PARAM_CHK(needed < 1, status, ERROR_VQUEUE_INVLD_PARAM);
-	VQ_PARAM_CHK(vq->vq_free_cnt == 0, status, ERROR_VRING_FULL);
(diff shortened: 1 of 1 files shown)


lib/virtio/virtqueue.c
```diff
@@ -129,7 +129,7 @@ int virtqueue_add_buffer(struct virtqueue *vq, struct virtqueue_buf *buf_list,
 
 	needed = readable + writable;
 
-	VQ_PARAM_CHK(vq == NULL, status, ERROR_VQUEUE_INVLD_PARAM);
+	VQ_PARAM_CHK(vq < needed, status, ERROR_VQUEUE_INVLD_PARAM);
```
arnopo, line 132: The comparison is not valid, you compare an integer with a structure pointer.
What about following check?
`VQ_PARAM_CHK((needed < 1 || needed > vq->vq_free_cnt), status, ERROR_VQUEUE_INVLD_PARAM);`

arnopo (changes requested):

lib/virtio/virtqueue.c
tammyleino: I had actually fixed this as follows:

VQ_PARAM_CHK(vq->vq_free_cnt < needed, status, ERROR_VRING_FULL);

Not sure how I truncated vq_free_cnt on my commit.

arnopo (approved):
LGTM

edmooring (approved):
Looks good to go.

PR #369, opened 2022-04-12
_rpmsg_virtio_get_buffer_size should return error on failure

tammyleino:
The code wrongly returns 0 if a descriptor is not available.

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -276,11 +276,12 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
 		length =
 		    (int)virtqueue_get_desc_size(rvdev->svq) -
 		    sizeof(struct rpmsg_hdr);
-		if (length < 0) {
-			length = 0;
-		}
 	}
 #endif /*!VIRTIO_MASTER_ONLY*/
+
+	if (length <= 0) {
+		length = RPMSG_ERR_NO_BUFF;
+	}
 
 	return length;
 }


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -276,8 +276,8 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
 		length =
 		    (int)virtqueue_get_desc_size(rvdev->svq) -
 		    sizeof(struct rpmsg_hdr);
-		if (length < 0) {
-			length = 0;
+		if (length <= 0) {
+			length = RPMSG_ERR_NO_BUFF;
```
arnopo, line 280: Regarding virtqueue_get_desc_size seems that 
```
if (length = 0) {
	length = RPMSG_ERR_NO_BUFF;
}
```
is enough, and I would move it at the end of the function to also treat the` (role == RPMSG_MASTER)` case

Could you also provide more details in the commit message explaining that this commit is to fix the API?
Indeed this fix mismatch between code and documentation (see https://github.com/OpenAMP/open-amp/blob/main/lib/include/openamp/rpmsg_virtio.h#L133).
But this laso means that application can observe a side effect depending on how the rpmsg_virtio_get_buffer_size was tested.

arnopo (changes requested):

lib/rpmsg/rpmsg_virtio.c
tammyleino: If virtqueue_get_desc_size(rvdev->svq) returns zero, the code still goes on to subtract sizeof(struct rpmsg_hdr) from the length, so I think we need to check <= 0 

I will move the check to the end of the routine.

edmooring (approved):
Looks good to me.

arnopo (approved):
Same than https://github.com/OpenAMP/open-amp/pull/336
to apply on top of for the 2022.04 release as impact API

PR #371, opened 2022-04-13
remoteproc_virtio_notify could use null notify operation

tammyleino:
remoteproc_virtio_notify updated to check notify operation before invoking.


arnopo (approved):

edmooring (approved):
Good catch.

Issue #370, opened 2022-04-13
remoteproc_get_mem could match on wrong memory region name


tammyleino:
The remoteproc_get_mem function compares names using the length of the given string (excluding "end-of-string" character) as a limit:
```
   if (name) {

      if (!strncmp(name, mem->name, strlen(name)))

         return mem;
```
As a result, it is possible to get a match when the given name is a subset of mem->name. For example, name of "shared" will match the mem->name "shared_memory".

Proposed code change:
```
			if ((strlen(name) == strlen(mem->name)) && (!strncmp(name, mem->name, RPROC_MAX_NAME_LEN)))
				return mem;
```

arnopo:
`strncmp(name, mem->name, RPROC_MAX_NAME_LEN)` seems sufficient as `RPROC_MAX_NAME_LEN` is the max size of both strings.
The strncmp function will stop comparing if a null character is encountered in either in `name`  or `mem->name`.

tammyleino:
mem = "shared"
mem->name = "shared_memory"

Without the strlen comparison, the code will incorrectly return a match, because it will encounter a null character in mem before getting to the end of mem->name

arnopo:
I made a short test on my Linux PC
```
	char s1[20] = "shared_ram";
	char s2[10] = "shared";
	ret = strncmp(s1,s2, 6);
	LPRINTF("strncmp result1: %d\r\n", ret);
	ret = strncmp(s1,s2, 8);
	LPRINTF("strncmp result2: %d\r\n", ret);
```
the result is 
```
strncmp result1: 0
strncmp result2: 95
```
unless the behavior depends on the C library, it seems that the comparison using strncmp with RPROC_MAX_NAME_LEN size should detect the difference between the two strings.

tammyleino:
@arnopo  I was concerned about C lib compatibility, which is why I proposed this solution, but after some experimentation with different compilers, I don't see a problem with excluding the strlen from the comparison.

PR #374, opened 2022-04-14
remoteproc_get_mem could wrongly match on wrong name

tammyleino:
Modified remoteproc_get_mem to bound the name comparison by RPROC_MAX_NAME_LEN instead of the length of the incoming search string

### DIFF
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -48,7 +48,7 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 	metal_list_for_each(&rproc->mems, node) {
 		mem = metal_container_of(node, struct remoteproc_mem, node);
 		if (name) {
-			if (!strncmp(name, mem->name, strlen(name)))
+			if (!strncmp(name, mem->name, RPROC_MAX_NAME_LEN))
 				return mem;
 		} else if (pa != METAL_BAD_PHYS) {
 			metal_phys_addr_t pa_start, pa_end;


lib/remoteproc/remoteproc.c
```diff
@@ -48,7 +48,7 @@ remoteproc_get_mem(struct remoteproc *rproc, const char *name,
 	metal_list_for_each(&rproc->mems, node) {
 		mem = metal_container_of(node, struct remoteproc_mem, node);
 		if (name) {
-			if (!strncmp(name, mem->name, strlen(name)))
+			if (!strncmp(name, mem->name, RPROC_MAX_NAME_LEN))
```
edmooring, line 51: I believe the current behavior is correct. If strlen(name) > RPROC_MAX_NAME_LEN, we will get an out-of-bounds access to mem->name.

Quoting an old BSD man page I had handy:

> The strcmp() and strncmp() functions lexicographically compare the null-ter-
     minated strings s1 and s2.

     The strncmp() function compares not more than n characters.  Because
     strncmp() is designed for comparing strings rather than binary data, charac-
     ters that appear after a `\0' character are not compared.

arnopo: @edmooring: 
The issue here is strlen(name) < RPROC_MAX_NAME_LEN
Please have a look to my comment in associated issue: 
https://github.com/OpenAMP/open-amp/issues/370#issuecomment-1099401883

arnopo (approved):

lib/remoteproc/remoteproc.c
edmooring: After further discussion with @arnopo, I agree that this change is an improvement.

edmooring (approved):
Given that an earlier check constrains strlen(name) to < RPROC_MAX_NAME_LEN, this change makes things clearer.

PR #375, opened 2022-04-18
openamp: divide shram to TX shram & RX shram

Donny9:
openamp: divide shram to TX shram & RX shram 

Signed-off-by: Jiuzhu Dong <dongjiuzhu1@xiaomi.com>

### DIFF
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -49,10 +49,12 @@ struct rpmsg_virtio_shm_pool {
  *
  * @h2r_buf_size: the size of the buffer used to send data from host to remote
  * @r2h_buf_size: the size of the buffer used to send data from remote to host
+ * @split_shpool: the flag that splitting share memory pool to TX and RX
  */
 struct rpmsg_virtio_config {
 	uint32_t h2r_buf_size;
 	uint32_t r2h_buf_size;
+	bool split_shpool;
 };
 
 /**
@@ -196,8 +198,13 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
  * @param ns_bind_cb  - callback handler for name service announcement without
  *                      local endpoints waiting to bind.
  * @param shm_io - pointer to the share memory I/O region.
- * @param shpool - pointer to shared memory pool. rpmsg_virtio_init_shm_pool has
- *                 to be called first to fill this structure.
+ * @param shpool - pointer to shared memory pool array.
+ *                 If the config->split_shpool is turn on, the array will contain
+ *                 two elements, the shpool of txshpool and rxshpool, Otherwise,
+ *                 the array has only one element, and txshpool rxshpool shares
+ *                 a shpool.
+ *                 And rpmsg_virtio_init_shm_pool has to be called first to fill
+ *                 each shpool in this array.
  * @param config - pointer to configuration structure
  *
  * @return - status of function execution
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -32,6 +32,7 @@
 	(&(const struct rpmsg_virtio_config) {     \
 		.h2r_buf_size = RPMSG_BUFFER_SIZE, \
 		.r2h_buf_size = RPMSG_BUFFER_SIZE, \
+		.split_shpool = false,             \
 	})
 #else
 #define RPMSG_VIRTIO_DEFAULT_CONFIG          NULL
@@ -689,11 +690,11 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 		 * Since device is RPMSG Remote so we need to manage the
 		 * shared buffers. Create shared memory pool to handle buffers.
 		 */
+		rvdev->shpool = config->split_shpool ? shpool + 1 : shpool;
 		if (!shpool)
 			return RPMSG_ERR_PARAM;
-		if (!shpool->size)
+		if (!shpool->size || !rvdev->shpool->size)
 			return RPMSG_ERR_NO_BUFF;
-		rvdev->shpool = shpool;
 
 		vq_names[0] = "rx_vq";
 		vq_names[1] = "tx_vq";


edmooring (changes requested):
This needs a detailed explanation as to why the check that the other end is ready is unnecessary.

Donny9:
> This needs a detailed explanation as to why the check that the other end is ready is unnecessary.

Okay.

Patch 1. In the multi core of lower power device, when one of core enters sleep, it needs to put its corresponding share memory into retention mode to save power consumption. Based on the limitations of the chip design, when the CPU to which share memory belongs goes to sleep, the share memory enters the retention mode, and other cores will not be able to access it. When the share memory divides tx shm and rx shm and the core of tx shm and rx shm are different, so that when one CPU sleeps, the other CPU can still access its own tx shm.

Patch 2. Base 1, The call of rproc_virtio_get_status will access the resource table, but the resource table only belongs to the shm of one of the CPUs, so after sleep, the other CPU cannot access it, so we don't to check the status for rpmsg_virtio_get_tx_payload_buffer every times.

xiaoxiang781216:
BTW, the boot already check the status flag:
https://github.com/OpenAMP/open-amp/blob/main/lib/rpmsg/rpmsg_virtio.c#L675-L678
https://github.com/OpenAMP/open-amp/blob/main/lib/rpmsg/rpmsg_virtio.c#L226-L242
so, the check in get_tx_payload is redundant.

arnopo:
To help could you detail your memory mapping to understand in which memory are your shared structures?
- The resource table
- The vrings
- RX buffers
-TX buffers
- log trace buffer ( if declared in the resource table ?

which CPU put the memory in retention:  both or only one core ( in this case is it the host or the remote processor?)
Thanks in advance

xiaoxiang781216:
> To help could you detail your memory mapping to understand in which memory are your shared structures?

Let's name CPU A(master) and B(slave).

> 
> * The resource table

SRAM A

> * The vrings

tx: SRAM A
rx: SRAM B

> * RX buffers

SRAM B

>   -TX buffers

SRAM A

> * log trace buffer ( if declared in the resource table ?
> 

SRAM B

> which CPU put the memory in retention: both or only one core ( in this case is it the host or the remote processor?) Thanks in advance

both, SRAM A and B can enter the retention state independently.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -161,6 +161,38 @@ int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
...
+ *
+ * @return - status of function execution
+ */
+int rpmsg_init_vdev_ext(struct rpmsg_virtio_device *rvdev,
+			struct virtio_device *vdev,
+			rpmsg_ns_bind_cb ns_bind_cb,
+			struct metal_io_region *shm_io,
+			struct rpmsg_virtio_shm_pool *tx_shpool,
+			struct rpmsg_virtio_shm_pool *rx_shpool);
+
```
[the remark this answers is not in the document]
Donny9: If shpool is a array, we need to modify the prototype of rpmsg_init_vdev to 
`int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,    
                    struct virtio_device *vdev,
                    rpmsg_ns_bind_cb ns_bind_cb,
                    struct metal_io_region *shm_io,
                    struct rpmsg_virtio_shm_pool **shpool)
`
This leads to many changes.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -324,16 +324,10 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
...
-	int status;
 
 	/* Get the associated remote device for channel. */
 	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
 
-	/* Validate device state */
-	status = rpmsg_virtio_get_status(rvdev);
-	if (!(status & VIRTIO_CONFIG_STATUS_DRIVER_OK))
-		return NULL;
-
```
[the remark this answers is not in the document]
Donny9: yes, i do after the above plan is confirmed

lib/include/openamp/rpmsg_virtio.h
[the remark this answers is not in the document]
Donny9: Done!

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -788,6 +767,52 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 	return status;
 }
 
+int rpmsg_virtio_get_buffer_size(struct rpmsg_device *rdev)
```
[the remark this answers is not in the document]
Donny9: revert

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -788,6 +767,52 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
...
+	if (!rdev)
+		return RPMSG_ERR_PARAM;
+	metal_mutex_acquire(&rdev->lock);
+	rvdev = (struct rpmsg_virtio_device *)rdev;
+	size = _rpmsg_virtio_get_buffer_size(rvdev);
+	metal_mutex_release(&rdev->lock);
+	return size;
+}
+
+int rpmsg_init_vdev(struct rpmsg_virtio_device *rvdev,
```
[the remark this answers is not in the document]
Donny9: Revert

lib/rpmsg/rpmsg_virtio.c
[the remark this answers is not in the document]
Donny9: @arnopo Done, thank you. https://github.com/OpenAMP/open-amp/pull/387

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -53,6 +53,7 @@ struct rpmsg_virtio_shm_pool {
 struct rpmsg_virtio_config {
 	uint32_t h2r_buf_size;
 	uint32_t r2h_buf_size;
+	bool txrx_shpool;
```
xiaoxiang781216, line 56: splitted_shpool?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -743,7 +744,8 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 		vqbuf.len = rvdev->config.r2h_buf_size;
 		for (idx = 0; idx < rvdev->rvq->vq_nentries; idx++) {
 			/* Initialize TX virtqueue buffers for remote device */
-			buffer = rpmsg_virtio_shm_pool_get_buffer(shpool,
+			buffer = rpmsg_virtio_shm_pool_get_buffer(
+					shpool[config->txrx_shpool ? 1 : 0],
```
xiaoxiang781216, line 748: config->txrx_shpool

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -191,7 +197,7 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 				struct virtio_device *vdev,
 				rpmsg_ns_bind_cb ns_bind_cb,
 				struct metal_io_region *shm_io,
-				struct rpmsg_virtio_shm_pool *shpool,
+				struct rpmsg_virtio_shm_pool **shpool,
```
arnopo, line 200: We should avoid to modify API as impact all users.
I was expecting to keep `rpmsg_virtio_shm_pool *`  pointer which points to the first element and use `shpool++` to internally address next array element . 
In addition I quite surprised that you don't need also to duplicate the `shm_io`

An alternative is to replace  in struct `rpmsg_virtio_config` the `txrx_shpool` by a 
`struct rpmsg_virtio_shm_pool **txrx_shpool`,
(and add a `struct metal_io_region **shm_io `)

Donny9: Done!

lib/include/openamp/rpmsg_virtio.h
Donny9: split_shpool.Done!

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -689,11 +690,11 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 		 * Since device is RPMSG Remote so we need to manage the
 		 * shared buffers. Create shared memory pool to handle buffers.
 		 */
+		rvdev->shpool = config->split_shpool ? shpool + 1 : shpool;
```
edmooring, line 693: If config->split_shpool is set to a true value other than 1, this will give bad values for rvdev->shpool. I suggest making the value explicitly 1 or 2.

edmooring (changes requested):
Otherwise, this looks reasonable.

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -53,6 +53,7 @@ struct rpmsg_virtio_shm_pool {
 struct rpmsg_virtio_config {
 	uint32_t h2r_buf_size;
 	uint32_t r2h_buf_size;
+	bool split_shpool;
```
arnopo, line 56: The field description is missing  above. Please also update the commit subject according to the new config name

lib/rpmsg/rpmsg_virtio.c
arnopo: @edmooring, could you re-check? for me it Ok as split_shpool is a boolean, but perhaps I missed something

arnopo (changes requested):
minor fixes on comments and commit message  else LGTM

lib/include/openamp/rpmsg_virtio.h
Donny9: Done!

arnopo (approved):

lib/rpmsg/rpmsg_virtio.c
edmooring: @arnopo, you are correct. I didn't look at enough of the context.

edmooring (approved):
Looks good to me. I appreciate the commit message explaining why this code is needed.

PR #376, opened 2022-04-18
openamp: add ns_unbind_cb to rpmsg_device

xiaoxiang781216:
it's useful to notify the port layer and also symmetry with ns_bind_cb callback.

### DIFF
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -109,6 +109,8 @@ struct rpmsg_device_ops {
  * @lock: mutex lock for rpmsg management
  * @ns_bind_cb: callback handler for name service announcement without local
  *              endpoints waiting to bind.
+ * @ns_unbind_cb: callback handler for name service announcement, called when
+ *                remote ept is destroyed.
  * @ops: RPMsg device operations
  * @support_ns: create/destroy namespace message
  */
@@ -118,6 +120,7 @@ struct rpmsg_device {
 	unsigned long bitmap[metal_bitmap_longs(RPMSG_ADDR_BMP_SIZE)];
 	metal_mutex_t lock;
 	rpmsg_ns_bind_cb ns_bind_cb;
+	rpmsg_ns_bind_cb ns_unbind_cb;
 	struct rpmsg_device_ops ops;
 	bool support_ns;
 };
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -585,6 +585,8 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 		metal_mutex_release(&rdev->lock);
 		if (_ept && _ept->ns_unbind_cb)
 			_ept->ns_unbind_cb(_ept);
+		if (rdev->ns_unbind_cb)
+			rdev->ns_unbind_cb(rdev, name, dest);
 	} else {
 		if (!_ept) {
 			/*
--- a/docs/rpmsg-design.md
+++ b/docs/rpmsg-design.md
@@ -101,6 +101,12 @@ running on two processors.
   void (*rpmsg_ns_bind_cb)(struct rpmsg_device *rdev,
 			   const char *name, uint32_t dest)
   ```
+* RPMsg name service unbind callback. If user defines such callback, when
+  there is name service destroy arrives, it will call this callback.:
+  ```
+  void (*rpmsg_ns_unbind_cb)(struct rpmsg_device *rdev,
+			   const char *name, uint32_t dest)
+  ```
 * RPMsg endpoint name service unbind callback. If user defines such callback,
   when there is name service destroy arrives, it will call this callback to
   notify the user application about the remote has destroyed the service.:


arnopo:
Could you detail your need? in which use case you need to notify the port layer in addition of the endpoint?

xiaoxiang781216:
> Could you detail your need? in which use case you need to notify the port layer in addition of the endpoint?

In some case, the service/apps may start later than open-amp framework, and then lose the creation notification from remote peer. To fix this issue, the porting layer could catch the notification and forward it to the client/service at the launch.
Since the client/server may create and destroy the endpointer dynamically and unbound, the porting layer need some mechanism(unbind callback here) to reclaim the resource from cache.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -584,6 +584,8 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 		metal_mutex_release(&rdev->lock);
 		if (_ept && _ept->ns_unbind_cb)
 			_ept->ns_unbind_cb(_ept);
+		if (rdev->ns_unbind_cb)
```
arnopo, line 587: At this point you can not ensure that the endpoint is released, 
Seems more safe to call the call back in rpmsg_destroy_ept

arnopo (approved):

arnopo:
> > Could you detail your need? in which use case you need to notify the port layer in addition of the endpoint?
> 
> In some case, the service/apps may start later than open-amp framework, and then lose the creation notification from remote peer. To fix this issue, the porting layer could catch the notification and forward it to the client/service at the launch. 
Since the client/server may create and destroy the endpointer dynamically and unbound, the porting layer need some mechanism(unbind callback here) to reclaim the resource from cache.

In such case, is the endpoint created by the porting layer?
Seems quite tricky to address:
- your port layer could free resource that are still used by the application.
- you have to manage error case where an unbind is receive after the bind, while the application is not ready.

What about having the `endpoint->ns_unbind_cb` calling a porting layer function that would forward to the application?

xiaoxiang781216:
> > > Could you detail your need? in which use case you need to notify the port layer in addition of the endpoint?
> > 
> > 
> > In some case, the service/apps may start later than open-amp framework, and then lose the creation notification from remote peer. To fix this issue, the porting layer could catch the notification and forward it to the client/service at the launch.
> > Since the client/server may create and destroy the endpointer dynamically and unbound, the porting layer need some mechanism(unbind callback here) to reclaim the resource from cache.
> 
> In such case, is the endpoint created by the porting layer?

No, the endpoint is still created by service/application. The resource I mean here is simple channel name and port number:
https://github.com/apache/incubator-nuttx/blob/master/drivers/rptun/rptun.c#L581-L582

> Seems quite tricky to address:
> 
> * your port layer could free resource that are still used by the application.

No, the house keeping data is managed by the port layer internally, not shared with application at all.

> * you have to manage error case where an unbind is receive after the bind, while the application is not ready.
> 

Of course, the application doesn't get any notification in this case since the channel name and port number is removed from cache in ns_unbind_cb callback.

> What about having the `endpoint->ns_unbind_cb` calling a porting layer function that would forward to the application?

As I explain before, the porting layer doesn't create the endpoint behalf the application so it can't hook the notification from  `endpoint->ns_unbind_cb`.

BTW, it's also wrong to create the endpoint in porting layer because:

1. The service represented by the endpoint mayn't really exist at all. It's hard to know from the porting layer in a dynamical environment.
2. Even the service exist, it's wrong to create the endpoint before the service is ready. e.g. How to handle the message send from client before the service start?

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: This callback just notify the porting layer the endpoint at the remote side was destroyed, not relate to the life cycle of the local endpoint, and it's also symmetric with ns_bind_cb at line 599.

arnopo (changes requested):
Thanks for the link to the code,  easier to understand your need.
The addition of this callback seems reasonable to me.
just documentation to address

lib/include/openamp/rpmsg.h
```diff
@@ -118,6 +118,7 @@ struct rpmsg_device {
 	unsigned long bitmap[metal_bitmap_longs(RPMSG_ADDR_BMP_SIZE)];
 	metal_mutex_t lock;
 	rpmsg_ns_bind_cb ns_bind_cb;
+	rpmsg_ns_bind_cb ns_unbind_cb;
```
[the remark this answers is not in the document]
xiaoxiang781216: Done.

lib/rpmsg/rpmsg_virtio.c
xiaoxiang781216: Both rpmsg and remoteproc layer need bind/unbind in many case, but bind is only sent to rpmsg_device_ops and unbind is only sent to rpmsg_endpoint in the current code base. PR https://github.com/OpenAMP/open-amp/pull/246 and https://github.com/OpenAMP/open-amp/pull/376 make both layer can receive the bind/unbind notification which make the notification more symmetric and functional.

arnopo: Could you add callback description in https://github.com/OpenAMP/open-amp/blob/main/docs/rpmsg-design.md#rpmsg-user-defined-callbacks documentation?
To help user Please, please add sub chapter for endpoint callbacks and service announcement callbacks

xiaoxiang781216: Done.

arnopo (approved):

arnopo (approved):

edmooring (approved):
This seems like a sensible addition.

PR #378, opened 2022-04-19
checkpatch: Increase the default limit to 100 characters

arnopo:
The 80 character limit is a warning which sometime affect the readability.
let's relax this limit to 100 as done in some other projects

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -2,7 +2,7 @@
 --no-tree
 --summary-file
 --show-types
---max-line-length=80
+--max-line-length=100
 --min-conf-desc-length=1
 
 --ignore BRACES


.checkpatch.conf
```diff
@@ -2,7 +2,7 @@
 --no-tree
 --summary-file
 --show-types
---max-line-length=80
+--max-line-length=100
```
edmooring, line 5: Does this make it more difficult to integrate with Zephyr? If it does, I would keep the old value and work around it in OpenAMP.

arnopo: No impact on Zephyr integration, the zephyr checkpatch rules are only for the Zephyr Os repository not the modes.

edmooring (commented):
Good to go.

edmooring (approved):
Good to go.

PR #387, opened 2022-05-06
openamp: don't need check status when get_tx_payload

Donny9:
Remove redundant check, because this status is already check in
rpmsg_init_vdev_with_config and rpmsg_virtio_wait_remote_ready.

Signed-off-by: Jiuzhu Dong <dongjiuzhu1@xiaomi.com>

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -324,15 +324,9 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 	struct rpmsg_hdr *rp_hdr;
 	uint16_t idx;
 	int tick_count;
-	int status;
 
 	/* Get the associated remote device for channel. */
 	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
-
-	/* Validate device state */
-	status = rpmsg_virtio_get_status(rvdev);
-	if (!(status & VIRTIO_CONFIG_STATUS_DRIVER_OK))
-		return NULL;
 
 	if (wait)
 		tick_count = RPMSG_TICK_COUNT / RPMSG_TICKS_PER_INTERVAL;


xiaoxiang781216:
> Hi, Thanks @Donny9 for creating this separate PR
> 
> @wmamills, @bentheredonethat This is the RP mentioned yesterday at the OpenAMP RP meeting, during discussion around the recovery from a host processor crash . We probably need to clarify first the host crash use case ( including Linux kernel), to ensure that this PR is compatible
> 
> * Should we implement the DEVICE_NEEDS_RESET?
> * Using a cached status, updated when notified?
> * ...

We have done the crash recovery before, basically it's very hardware and rpmsg service specific, OpenAMP self doesn't need too much work/adjust in this case. 
To recovery virito/rpmsg self, we need:

1. Detect the remote crash/hang(could be watchdog, status bit or keepalive ping/pong...)
2. Stop/restart remoteproc in the abnormal case(may generate dump before restart)

All this work can(should) be done in the porting layer, this is also how Linux remoteproc done:

1. remoteproc driver detect the abnormal condition
2. remoteproc framework handle restart process

bentheredonethat:
In theory i am on board. can you show a test or demo setup with this working?

CC: @arnopo  @wmamills  @Donny9  @xiaoxiang781216

xiaoxiang781216:
> In theory i am on board. can you show a test or demo setup with this working?
> 
> CC: @arnopo @wmamills @Donny9 @xiaoxiang781216

please see how stm32 do the check with wdog here:
https://github.com/torvalds/linux/blob/master/drivers/remoteproc/stm32_rproc.c#L280

xiaoxiang781216:
@arnopo what's about this patch? I think this change doesn't impact other advanced recovery mechanism.

arnopo:
Hi @xiaoxiang781216, @bentheredonethat 

if I'm looking to the initialisation step:
- on virtio device side: The endpoint should not be created if `VIRTIO_CONFIG_STATUS_DRIVER_OK` is not set so test should not needed, but no guaranty that the `rproc_virtio_wait_remote_ready` is called. So if the test was suppressed here, we should at least have a check around the `rpmsg_create_ept` function.
- on virtio driver side: as vdev is created every thing should be ok. 

My concern is more around the robustness on runtime or on shutdown. The functions `rpmsg_get_tx_payload_buffer` and `rpmsg_virtio_send_offchannel_raw` are called by the users. We can not prevent a call while the the virtio is no more properly configured. 
  
2 questions:
- What would be improvement in term of performance if we suppress this check?
- What  happen if status is not VIRTIO_CONFIG_STATUS_DRIVER_OK? Seems to me that we could have an illegal access trying to access to vrings that are no more valid.

bentheredonethat:
Hi @arnopo 

What would be improvement in term of performance if we suppress this check?
- not sure
What happen if status is not VIRTIO_CONFIG_STATUS_DRIVER_OK? Seems to me that we could have an illegal access trying to access to vrings that are no more valid.
- if this is not set then we don't have guarantee that both parties are aligned in setup. also if there is pending buffer something could go awry.

xiaoxiang781216:
> Hi @xiaoxiang781216, @bentheredonethat
> 
> if I'm looking to the initialisation step:
> 
> * on virtio device side: The endpoint should not be created if `VIRTIO_CONFIG_STATUS_DRIVER_OK` is not set so test should not needed, but no guaranty that the `rproc_virtio_wait_remote_ready` is called. So if the test was suppressed here, we should at least have a check around the `rpmsg_create_ept` function.

But remoteproc_create_virtio arleady call rproc_virtio_wait_remote_ready before return.

> * on virtio driver side: as vdev is created every thing should be ok.
> 
> My concern is more around the robustness on runtime or on shutdown. The functions `rpmsg_get_tx_payload_buffer` and `rpmsg_virtio_send_offchannel_raw` are called by the users. We can not prevent a call while the the virtio is no more properly configured.
> 

Even the code passes rproc_virtio_wait_remote_ready and continue to the next action, the remote side may enter reset state immediately after the check. So, the check can't really fix the problem.

> 2 questions:
> 
> * What would be improvement in term of performance if we suppress this check?
> * What  happen if status is not VIRTIO_CONFIG_STATUS_DRIVER_OK? Seems to me that we could have an illegal access trying to access to vrings that are no more valid.

We made this change not for performance, but for power saving. Since the hardware command shared RAM in low power mode when there is no pending message to process, checking the status may hang in this case.

arnopo:
> > Hi @xiaoxiang781216, @bentheredonethat
> > if I'm looking to the initialisation step:
> > 
> > * on virtio device side: The endpoint should not be created if `VIRTIO_CONFIG_STATUS_DRIVER_OK` is not set so test should not needed, but no guaranty that the `rproc_virtio_wait_remote_ready` is called. So if the test was suppressed here, we should at least have a check around the `rpmsg_create_ept` function.
> 
> But remoteproc_create_virtio arleady call rproc_virtio_wait_remote_ready before return.

Unfortunately remoteproc_create_virtio is not always used. For instance Zephyr does not use it for IPC service.

> 
> > * on virtio driver side: as vdev is created every thing should be ok.
> > 
> > My concern is more around the robustness on runtime or on shutdown. The functions `rpmsg_get_tx_payload_buffer` and `rpmsg_virtio_send_offchannel_raw` are called by the users. We can not prevent a call while the the virtio is no more properly configured.
> 
> Even the code passes rproc_virtio_wait_remote_ready and continue to the next action, the remote side may enter reset state immediately after the check. So, the check can't really fix the problem.

That's a good point. seems at least the function should be reworked to better use `metal_mutex_acquire`.

> 
> > 2 questions:
> > 
> > * What would be improvement in term of performance if we suppress this check?
> > * What  happen if status is not VIRTIO_CONFIG_STATUS_DRIVER_OK? Seems to me that we could have an illegal access trying to access to vrings that are no more valid.
> 
> We made this change not for performance, but for power saving. Since the hardware command shared RAM in low power mode when there is no pending message to process, checking the status may hang in this case.

That's clarify your need. 
I wonder if a `weak` function to check the status on runtime should not be a good compromise here, between your constraint and the legacy support. Is such implementation would  respond to your need ?

xiaoxiang781216:
> > > Hi @xiaoxiang781216, @bentheredonethat
> > > if I'm looking to the initialisation step:
> > > 
> > > * on virtio device side: The endpoint should not be created if `VIRTIO_CONFIG_STATUS_DRIVER_OK` is not set so test should not needed, but no guaranty that the `rproc_virtio_wait_remote_ready` is called. So if the test was suppressed here, we should at least have a check around the `rpmsg_create_ept` function.
> > 
> > 
> > But remoteproc_create_virtio arleady call rproc_virtio_wait_remote_ready before return.
> 
> Unfortunately remoteproc_create_virtio is not always used. For instance Zephyr does not use it for IPC service.
> 

But Zephyr code must do the similar check during the initialize phase insteading rpmsg_send_xxx, otherwise the driver will hit the race condition sooner or later.

One example is that the remote core need time to change state to VIRTIO_CONFIG_STATUS_DRIVER_OK, but the check in rpmsg_send_xxx may do the wrong response if it is called too early before the remote core set VIRTIO_CONFIG_STATUS_DRIVER_OK.

Another related race condition we hit and made the following change in mainline:
```
commit 832c5a6b4987ee67e1d3038a6310e20f62a40e82
Author: Xiang Xiao <xiaoxiang@xiaomi.com>
Date:   Thu Jan 3 23:07:26 2019 +0800

    call rproc_virtio_wait_remote_ready instead the adhoc loop
    
diff --git a/lib/remoteproc/remoteproc.c b/lib/remoteproc/remoteproc.c
index 46f35882..9068ae23 100644
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -7,7 +7,6 @@
  */
 
 #include <metal/alloc.h>
-#include <metal/cpu.h>
 #include <metal/log.h>
 #include <metal/utilities.h>
 #include <openamp/elf_loader.h>
@@ -906,17 +905,12 @@ remoteproc_create_virtio(struct remoteproc *rproc,
                return NULL;
        }
 
+       rproc_virtio_wait_remote_ready(vdev);
+
        rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
        metal_list_add_tail(&rproc->vdevs, &rpvdev->node);
        num_vrings = vdev_rsc->num_of_vrings;
 
-       while (1) {
-               if ((vdev->func->get_status(vdev) &
-                   VIRTIO_CONFIG_STATUS_DRIVER_OK)) {
-                       break;
-               }
-               metal_cpu_yield();
-       }
```
You can see rproc_virtio_wait_remote_ready need be done before we fetch num_of_vrings from resource table, because the remote core also need time to prepare the resource table(e.g. copy the table from flash to ram). Only the check done in rproc_virtio_wait_remote_ready can guarantee this.

BTW, virtio spec define the initialize handshaking in detail, any implementation must follow this sequence, otherwise you will hit  
mysterious and random error when you ship enough device which use OpenAMP in the very complex envrionment.

> > > * on virtio driver side: as vdev is created every thing should be ok.
> > > 
> > > My concern is more around the robustness on runtime or on shutdown. The functions `rpmsg_get_tx_payload_buffer` and `rpmsg_virtio_send_offchannel_raw` are called by the users. We can not prevent a call while the the virtio is no more properly configured.
> > 
> > 
> > Even the code passes rproc_virtio_wait_remote_ready and continue to the next action, the remote side may enter reset state immediately after the check. So, the check can't really fix the problem.
> 
> That's a good point. seems at least the function should be reworked to better use `metal_mutex_acquire`.

Yes, but even we add metal_mutex_acquire here and in the interrupt handler, the race condition still exists because the remote can't hold the same mutex cross the different OS(except we use the hardware spinlock).

> 
> > > 2 questions:
> > > 
> > > * What would be improvement in term of performance if we suppress this check?
> > > * What  happen if status is not VIRTIO_CONFIG_STATUS_DRIVER_OK? Seems to me that we could have an illegal access trying to access to vrings that are no more valid.
> > 
> > 
> > We made this change not for performance, but for power saving. Since the hardware command shared RAM in low power mode when there is no pending message to process, checking the status may hang in this case.
> 
> That's clarify your need. I wonder if a `weak` function to check the status on runtime should not be a good compromise here, between your constraint and the legacy support. Is such implementation would respond to your need ?

Yes, weak doesn't work here since the first waiting in remoteproc_create_virtio is required to do the correct sync.

arnopo:
> Yes, but even we add metal_mutex_acquire here and in the interrupt handler, the race condition still exists because the
> remote can't hold the same mutex cross the different OS(except we use the hardware spinlock).

I full agree that this test does not prevent from race condition, and today there not solution in existing virtio spec for IPC. as you mention it a shared resource that should be probably be protected by hardware spinlocks

> You can see rproc_virtio_wait_remote_ready need to be done before we fetch num_of_vrings from resource table,
> because the remote core also need time to prepare the resource table(e.g. copy the table from flash to ram). Only the ready check done in initial phase can guarantee this.

>BTW, virtio spec define the initialize handshaking in detail, any implementation must follow this sequence, otherwise >you will hit mysterious and random error when you ship enough devices which use OpenAMP in the very complex envrionment.

Yes for the init but here the check could also be related to the deinit detection, that seems not treated in the spec... 
- Does the device shouldd sent a  VIRTIO_CONFIG_STATUS_NEEDS_RESET ( that is not handled yet in OpenAMP and could also lead to hang in your low power use case)?
- Does  driver stops the communicationby cleaning the VIRTIO_CONFIG_STATUS_DRIVER_OK bit? how managethis  in virtio? 

That said, the only need we have to suppress the check, is for your low power use case. Else if I put in balance the risk on legacy implementation vs the benefit, I'm not in favor of suppressing it.

So before suppressing this check. I would prefer that we try in a first step to properly implement the management of this status in the remoteproc_virtio  to manage the deinit.
- status register cache? with update on notification?
- application Callback to notify applications on a virtio deinit/reset ?

get_status remoteproc dispatcher could rely on cached status instead of the register status.

PR #388, opened 2022-05-06
lib: Remove warnings from -Wcast-align

edmooring:
There were several places in the library where a char* was used to point
to a wider type (uint32_t or a struct). Casting the char* to the wider
type resulted in a compiler warning.

This commit adds a cast to void* to remove the compiler warnings. This
results in no functional change.

### DIFF
--- a/lib/proxy/rpmsg_retarget.c
+++ b/lib/proxy/rpmsg_retarget.c
@@ -178,7 +178,7 @@ int _open(const char *filename, int flags, int mode)
 		return -EINVAL;
 
 	/* Construct rpc payload */
-	syscall = (struct rpmsg_rpc_syscall *)tmpbuf;
+	syscall = (void *)tmpbuf;
 	syscall->id = OPEN_SYSCALL_ID;
 	syscall->args.int_field1 = flags;
 	syscall->args.int_field2 = mode;
@@ -228,7 +228,7 @@ int _read(int fd, char *buffer, int buflen)
 	syscall.args.int_field2 = buflen;
 	syscall.args.data_len = 0;	/*not used */
 
-	resp = (struct rpmsg_rpc_syscall *)tmpbuf;
+	resp = (void *)tmpbuf;
 	resp->id = 0;
 	ret = rpmsg_rpc_send(rpc, (void *)&syscall, payload_size,
 			     tmpbuf, sizeof(tmpbuf));
@@ -281,7 +281,7 @@ int _write(int fd, const char *ptr, int len)
 	if (fd == 1)
 		null_term = 1;
 
-	syscall = (struct rpmsg_rpc_syscall *)tmpbuf;
+	syscall = (void *)tmpbuf;
 	syscall->id = WRITE_SYSCALL_ID;
 	syscall->args.int_field1 = fd;
 	syscall->args.int_field2 = len;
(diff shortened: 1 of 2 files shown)


lib/remoteproc/rsc_table_parser.c
```diff
@@ -24,7 +24,7 @@ int handle_rsc_table(struct remoteproc *rproc,
 		     struct metal_io_region *io)
 {
 	char *rsc_start;
-	unsigned int rsc_type;
+	uint32_t rsc_type;
```
arnopo, line 27: I wonder here if we should not take oportunity o clean up the code using struct fw_rsc_hdr hdr

lib/remoteproc/rsc_table_parser.c
```diff
@@ -60,7 +60,7 @@ int handle_rsc_table(struct remoteproc *rproc,
 		if (io &&
 		    metal_io_virt_to_offset(io, rsc_start) == METAL_BAD_OFFSET)
 			return -RPROC_ERR_RSC_TAB_TRUNC;
-		rsc_type = *((uint32_t *)rsc_start);
+		rsc_type = *((uint32_t *)(void *)rsc_start);
```
arnopo, line 63: I wonder here if we should not take oportunity to clean up the code using `fw_rsc_hdr` structure
```
struct fw_rsc_hdr  hdr = (void *)((char *)rsc_table + rsc_table->offset[idx]);
rsc_type = hdr->type;
void * rsc_start = (void *)hdr;
```
rsc_start assignment seems to me magic without `fw_rsc_hdr` reference.

lib/proxy/rpmsg_retarget.c
```diff
@@ -228,7 +228,7 @@ int _read(int fd, char *buffer, int buflen)
 	syscall.args.int_field2 = buflen;
 	syscall.args.data_len = 0;	/*not used */
 
-	resp = (struct rpmsg_rpc_syscall *)tmpbuf;
+	resp = (struct rpmsg_rpc_syscall *)(void *)tmpbuf;
```
arnopo, line 231: is `resp = (void *)tmpbuf;`  ok?
if yes few optimizations are possible in your PR

lib/remoteproc/rsc_table_parser.c
```diff
@@ -209,7 +209,7 @@ size_t find_rsc(void *rsc_table, unsigned int rsc_type, unsigned int index)
 	for (i = 0; i < r_table->num; i++) {
 		rsc_start = (char *)r_table;
 		rsc_start += r_table->offset[i];
-		lrsc_type = *((uint32_t *)rsc_start);
+		lrsc_type = *((uint32_t *)(void *)rsc_start);
```
arnopo, line 212: same remarque here about the use of "fw_rsc_hdr" struct

lib/remoteproc/rsc_table_parser.c
edmooring: Agreed. I did the minimum necessary to remove the warning. The problem that this code is trying to resolve is that the two programs sharing this memory are running under different CPU architectures and built with different compilers. This will take some thought.

lib/proxy/rpmsg_retarget.c
edmooring: Yes, it is OK. I will simplify these bits. It should make the code easier to read.

arnopo (approved):

PR #391, opened 2022-05-13
apps: machines: clean-up the virtio device features definition

arnopo:
Provide a more explicit name and value for the virtio device
features.


edmooring (commented):
Looks good to me.

arnopo:
> Looks good to me.

could you change your review status to approved, in this case ?

edmooring (approved):

PR #392, opened 2022-05-13
doc: clean-up data structures

arnopo:
Realign data structure description with the code.
The Description of the libmetal API is removed.


edmooring (approved):
Looks good to me.

PR #394, opened 2022-05-20
Add flow control service 

arnopo:
This series proposes an implementation for the rpmsg virtio transport backend, of the signaling API proposed by Deepak Kumar Singh on Linux:
"rpmsg and glink signaling API support" [1]

The aim of the pull request is to offer the possibility for an endpoint to inform a remote endpoint about its state, based on a software flow control[2].

For this a new rpmsg service( with a fixed address 64) is proposed.
It is responsible for:
- transmitting local endpoint flow control information to the remote side,
- informing a local endpoint about a remote endpoint flow control.

For the rpmsg virtio transport layer the service is negotiated thanks to the virtio feature flag: VIRTIO_RPMSG_F_FC

Notice that this pull request introduces new feature in the rpmsg protocol,  So it has to be aligned with Linux implementation.
For this a pull request will sent to the linux remoteproc mailing to implement pending feature. 

[1]https://lkml.org/lkml/2022/1/18/867
[2]https://en.wikipedia.org/wiki/Software_flow_control


arnopo:
The series sent on Linux mailing list is available here:
https://lore.kernel.org/lkml/20220520082940.2984914-1-arnaud.pouliquen@foss.st.com/T/#t

lib/rpmsg/rpmsg_fc.c
```diff
@@ -0,0 +1,95 @@
...
+	return RPMSG_SUCCESS;
+}
+
+int rpmsg_set_flow_control(struct rpmsg_endpoint *ept, uint32_t flags)
+{
+	struct rpmsg_ept_msg  msg;
+
+	msg.flags = flags;
+	msg.src = ept->addr;
+	msg.dst = ept->dest_addr;
```
xiaoxiang781216, line 92: As I mention in other PR, the separated flow control isn't good because:

1. it isn't efficient to send a dedicated message for the flow control, since the same info could piggyback as a flag field in normal data transfer in most case.
2. ON/OFF info isn't enough in the advanced flow control since the additional info is required(e.g. the slide window, round trip delay, congestion etc..).

So I would expect the flow control is part of the protocol design, and the general flow control service is useless.

arnopo: Thanks for your feedback, could please send  your feedback on Linux mailing list instead. As this feature impact both, the Linux and the OpenAMP. Having a single place would ease discussion.

If you have not subscribed to the Linux remote proc mailing list you can get the "mbox" on patchwork 
https://patchwork.kernel.org/project/linux-remoteproc/list/?series=643501

lib/rpmsg/rpmsg_fc.c
```diff
@@ -0,0 +1,95 @@
...
+	metal_mutex_release(&rdev->lock);
+
+	if (!_ept)
+		return RPMSG_ERR_PARAM;
+
+	/*
+	 * An NS announcement has been sent to the remote side which answer with flow control.
+	 * Register the remote endpoint as default destination endpoint.
+	 */
+	if (_ept->dest_addr == RPMSG_ADDR_ANY)
```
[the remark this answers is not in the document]
arnopo: Agree, NS_ack would help for this

PR #396, opened 2022-05-20
Terminology: Replace potentially offensive terms in documentation

arnopo:
In line with the terminology changes proposed by the OpenAMP
Technical Steering committee. Suppress "master and "slave terms
in comments and docs.


arnopo:
Please Ignore compliance check failure, checkpatch returns false positive:
```
Error: -:981: WARNING:TYPO_SPELLING: 'doesn'' may be misspelled - perhaps 'doesn't'?
22
#981: FILE: docs/img-src/coprocessor-rpmsg-static-ep.gv:66:
23
+    s_rpmsg_vdev_init_comment [label="\l* As vdev doesn't support name service, it will not create name service endpoint;\l* It will not return until the host set status to DRIVER READY\l"];
```
the "As vdev doesn't support" is not misspelled

edmooring (approved):
Looks good to me.

PR #401, opened 2022-06-10
Introduce rpmsg_release_tx_buffer API

arnopo:
Add an API to be able to release unused TX buffers that will not be used. For example, this API can be called in case of an error between the buffer reservation and the sending to the remote side.

Associated with this API, a mechanism is proposed to temporarily store the virtio buffer that can not be put back into the vrings.
These buffers are stored in a "recycle" table to be reused as a priority on the next `rpmsg_get_tx_payload_buffer` call.

Note that the recycler is not based on a metal list but on a static array to respect the static memory allocation constraint.


lib/include/openamp/rpmsg_virtio.h
```diff
@@ -73,6 +95,8 @@ struct rpmsg_virtio_device {
 	struct virtqueue *svq;
 	struct metal_io_region *shbuf_io;
 	struct rpmsg_virtio_shm_pool *shpool;
+	uint32_t desc_recycler_cnt;
+	struct desc_recycler_t desc_recycler[RPMSG_RECYCLER_SIZE];
```
xiaoxiang781216, line 99: It's better to move them to rpmsg_virtio_shm_pool and add new api:
rpmsg_virtio_shm_pool_free_buffer or rpmsg_virtio_shm_pool_return_buffer

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -414,6 +432,40 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
...
+	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
+
+	metal_mutex_acquire(&rdev->lock);
+
+	if (rvdev->desc_recycler_cnt == RPMSG_RECYCLER_SIZE) {
+		status = RPMSG_ERR_NO_MEM;
+		goto ret_status;
+	}
+
+	rp_hdr = RPMSG_LOCATE_HDR(rxbuf);
```
xiaoxiang781216, line 451: since rxbuf is free, why not reuse rxbuf to form a free link list, so:

1. We can remove desc_recycler and the limitation of RPMSG_RECYCLER_SIZE
2. Avoid to increase the size of struct rpmsg_virtio_device

apps/tests/msg/CMakeLists.txt
```diff
@@ -15,10 +15,14 @@ collector_list (_deps PROJECT_LIB_DEPS)
 
 set (OPENAMP_LIB open_amp)
 
-foreach (_app msg-test-rpmsg-ping msg-test-rpmsg-update msg-test-rpmsg-flood-ping)
+foreach (_app msg-test-rpmsg-ping msg-test-rpmsg-nocopy-ping msg-test-rpmsg-nocopy-echo  msg-test-rpmsg-update msg-test-rpmsg-flood-ping )
```
xiaoxiang781216, line 18: remove the space before )

apps/tests/msg/rpmsg-nocopy-echo.c
```diff
@@ -0,0 +1,219 @@
...
+ * in the rpmsg virtio layer.
+ */
+
+#include <stdio.h>
+#include <openamp/open_amp.h>
+#include <metal/alloc.h>
+#include "platform_info.h"
+
+#define SHUTDOWN_MSG	0xEF56A55A
+#define RPMSG_SERVICE_NAME         "rpmsg-openamp-demo-channel"
```
xiaoxiang781216, line 19: align 0xEF56A55A with  "rpmsg-openamp-demo-channel"

lib/include/openamp/rpmsg_virtio.h
arnopo: The rpmsg_virtio_shm_pool structure is used to manage the shared memory for new buffer allocation. i would prefer in this case to create a new structure

lib/rpmsg/rpmsg_virtio.c
arnopo: yes would be nice to be able to remove limitation. But here rxbuf is pointing to the rpmsg payload, so I don't see how I can reuse it as I should give the pointer on `rpmsg_get_tx_payload_buffer`.
Please, could you detail what you have in mind, if I missed something?

xiaoxiang781216: we can directly cast rxbuf to desc_recycler_t * and link into the list since nobody use rxbuf after this point.

lib/include/openamp/rpmsg_virtio.h
xiaoxiang781216: it's very common and natural that shm pool support the free function, like malloc is paired with free.

lib/rpmsg/rpmsg_virtio.c
arnopo: Got it!
Your proposal is to reuse the rpmsg payload to temporay store the desc_recycler_t structure.
That's seems a good idea, I will rework the PR to implement in this way.
Thanks for the advice!

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -25,6 +25,21 @@
 /* Time to wait - In multiple of 1 msecs. */
 #define RPMSG_TICKS_PER_INTERVAL                1000
 
+/**
+ * struct vbuff_reclaimer_t - vring buffer recycler
+ *
+ * This structure is used by the rpmsg virtio to store unused virtio buffer, as the
+ * virtqueue structure has been already updated and memory allocated.
+ *
+ * @desc_idx: the size of the buffer used to send data from host to remote
```
xiaoxiang781216, line 34: the name and description doesn't match the usage

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -25,6 +25,21 @@
...
+ * @addr: address of the virtio buffer
+ * @node: node in reclaimer list.
+ */
+struct vbuff_reclaimer_t {
+	uint16_t idx;
+	void *addr;
+	struct metal_list node;
```
xiaoxiang781216, line 41: it's better to put the node to the first

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -414,6 +449,33 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 	return len;
 }
 
+static int rpmsg_virtio_release_tx_buffer(struct rpmsg_device *rdev, void *rxbuf)
```
xiaoxiang781216, line 452: rxbuf->txbuf

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -414,6 +449,33 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
...
+{
+	struct rpmsg_virtio_device *rvdev;
+	struct rpmsg_hdr *rp_hdr;
+	struct vbuff_reclaimer_t *r_desc = rxbuf;
```
xiaoxiang781216, line 456: it's better to let r_desc point to rp_hdr and remove addr field from vbuff_reclaimer_t

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -157,25 +172,45 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
...
-					rvdev->config.h2r_buf_size);
+		if (role == RPMSG_HOST)
 			*len = rvdev->config.h2r_buf_size;
-			*idx = 0;
+#endif /*!VIRTIO_DEVICE_ONLY*/
+#ifndef VIRTIO_DRIVER_ONLY
+		if (role == RPMSG_REMOTE)
+			*len = virtqueue_get_buffer_length(rvdev->svq, *idx);
+#endif /*!VIRTIO_DRIVER_ONLY*/
+	} else {
```
xiaoxiang781216, line 195: change to:
```
#ifndef VIRTIO_DEVICE_ONLY
else if (role == RPMSG_HOST) {
...
}
#endif /*!VIRTIO_DEVICE_ONLY*/
#ifndef VIRTIO_DRIVER_ONLY
else if (role == RPMSG_REMOTE) {
...
}
#endif /*!VIRTIO_DRIVER_ONLY*/
```
could remove the unnecessary change.

lib/include/openamp/rpmsg.h
```diff
@@ -99,6 +101,7 @@ struct rpmsg_device_ops {
 	int (*send_offchannel_nocopy)(struct rpmsg_device *rdev,
 				      uint32_t src, uint32_t dst,
 				       const void *data, int len);
+	int (*release_tx_buffer)(struct rpmsg_device *rdev, void *rxbuf);
```
xiaoxiang781216, line 104: rxbuf->txbuf

xiaoxiang781216: but not change

arnopo: The [compare ](https://github.com/OpenAMP/open-amp/compare/806f192983475d06cb64930e832b2cdc8d4853ea..148ae1d1da65d48415eb485fccb0c413c0daf9d6) button associated to my force-update shows the  delta
Please, could check it again?

xiaoxiang781216: It look good now.

lib/include/openamp/rpmsg.h
```diff
@@ -334,6 +340,26 @@ void rpmsg_release_rx_buffer(struct rpmsg_endpoint *ept, void *rxbuf);
...
+ * remote side.
+ *
+ * @ept: the rpmsg endpoint
+ * @txbuf: tx buffer with message payload
+ *
+ * @return:
+ *   - RPMSG_SUCCESS on success
+ *   - RPMSG_ERR_PARAM on invalid parameter
+ *   - RPMSG_ERR_PERM if service not implemented
+ *   - RPMSG_ERR_NO_MEM is the TX buffer recycler if full.
```
carlocaione, line 357: is -> if

lib/include/openamp/rpmsg.h
carlocaione: uhm, I still see `rxbuf` here.
Screenshot from 2022-07-13 11-56-43

lib/virtio/virtqueue.c
```diff
@@ -226,6 +226,12 @@ uint32_t virtqueue_get_buffer_length(struct virtqueue *vq, uint16_t idx)
 	return vq->vq_ring.desc[idx].len;
 }
 
+void *virtqueue_get_buffer_addr(struct virtqueue *vq, uint16_t idx)
+{
+	VRING_INVALIDATE(vq->vq_ring.desc[idx].len);
```
carlocaione, line 231: Should we invalidate `vq->vq_ring.desc[idx].addr`?

lib/include/openamp/rpmsg.h
```diff
@@ -101,7 +101,7 @@ struct rpmsg_device_ops {
 	int (*send_offchannel_nocopy)(struct rpmsg_device *rdev,
 				      uint32_t src, uint32_t dst,
 				       const void *data, int len);
-	int (*release_tx_buffer)(struct rpmsg_device *rdev, void *rxbuf);
+	int (*release_tx_buffer)(struct rpmsg_device *rdev, void *txbuf);
```
carlocaione, line 104: ok, this is fixed in a different commit then. Can you squash this change into the first commit?

lib/include/openamp/rpmsg.h
arnopo: My apologize i was not fix the good commit, done should be ok

carlocaione (approved):

edmooring (changes requested):

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -414,6 +444,32 @@ static int rpmsg_virtio_send_offchannel_nocopy(struct rpmsg_device *rdev,
 	return len;
 }
 
+static int rpmsg_virtio_release_tx_buffer(struct rpmsg_device *rdev, void *txbuf)
```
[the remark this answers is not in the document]
arnopo: After more investigations, the detection of a double release is not simple. If it quite simple to test that the buffer is not in the reclaimer list, it is not possible to detect the case if the buffer as been recycled ( so no more in the list) or even already used to send a rpmsg.

I added a note in `rpmsg_virtio_release_tx_buffer`  header: 
https://github.com/OpenAMP/open-amp/pull/401/commits/c172737c44b12ad3f9edda8ffb650fd326f356cd#diff-145cb24aa08983b6c2538aedc198375a454527d8311596cb87d970b917168971R347

lib/include/openamp/rpmsg.h
```diff
@@ -334,6 +337,29 @@ void rpmsg_release_rx_buffer(struct rpmsg_endpoint *ept, void *rxbuf);
...
+/**
+ * @brief Releases unused buffer.
+ *
+ * This API can be called at process context when the Tx buffer reserved by
+ * rpmsg_get_tx_payload_buffer needs to be released without been sent to the
+ * remote side.
+ *
+ * Note that the rpmsg virtio is not able to detect if a buffer has already been released.
+ * The user must prevent a double release (e.g. by resetting its buffer pointer to zero after
+ * the release).
```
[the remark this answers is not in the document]
arnopo: As mentioned [in a previous comment](https://github.com/OpenAMP/open-amp/pull/401#discussion_r923485526) , seems to me no way to guaranty, for instance in  this in following cases:
- the user send the buffer and try to release it after
- the process A release a buffer, the proccess B get it from the recycler, the process A release it again.

This is quite similar to using a pointer after its associated memory has been freed.

lib/include/openamp/rpmsg.h
```diff
@@ -334,6 +337,29 @@ void rpmsg_release_rx_buffer(struct rpmsg_endpoint *ept, void *rxbuf);
 void *rpmsg_get_tx_payload_buffer(struct rpmsg_endpoint *ept,
 				  uint32_t *len, int wait);
 
+/**
+ * @brief Releases unused buffer.
+ *
+ * This API can be called at process context when the Tx buffer reserved by
```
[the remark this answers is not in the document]
arnopo: Copy past from rpmsg_get_tx_payload_buffer header. but I can suppress "at process context" if it is confusing.

lib/include/openamp/rpmsg.h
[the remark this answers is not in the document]
edmooring: I see your point. In resource-constrained environments, this is about all that can be done.

edmooring (approved):

PR #406, opened 2022-07-25
Application-supplied buffer addresses not checked for validity

tammyleino:
Code must check for valid buffer address to avoid potential corruption

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -288,9 +288,14 @@ static int _rpmsg_virtio_get_buffer_size(struct rpmsg_virtio_device *rvdev)
 
 static void rpmsg_virtio_hold_rx_buffer(struct rpmsg_device *rdev, void *rxbuf)
 {
+	struct rpmsg_virtio_device *rvdev;
 	struct rpmsg_hdr *rp_hdr;
 
-	(void)rdev;
+	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
+
+	if (!metal_io_is_block_valid(rvdev->rvq->shm_io, sizeof(struct rpmsg_hdr),
+				     rxbuf, 0))
+		return;
 
 	rp_hdr = RPMSG_LOCATE_HDR(rxbuf);
 
@@ -307,6 +312,11 @@ static void rpmsg_virtio_release_rx_buffer(struct rpmsg_device *rdev,
 	uint32_t len;
 
 	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
+
+	if (!metal_io_is_block_valid(rvdev->rvq->shm_io, sizeof(struct rpmsg_hdr),
+				     rxbuf, 0))
+		return;
+
 	rp_hdr = RPMSG_LOCATE_HDR(rxbuf);
 	/* The reserved field contains buffer index */
 	idx = (uint16_t)(rp_hdr->reserved & ~RPMSG_BUF_HELD);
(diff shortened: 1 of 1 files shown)


edmooring (commented):
Would it make sense to make these chunks into a macro? At first glance, they are all doing the same thing.

tammyleino:
@edmooring Yes, I can do that.  Thank you.

tammyleino:
I have not ported my API test application to OpenAMP, but here are the tests I am running against our implementation:

    /*** rpmsg_hold_rx_buffer ***/

    /* null rpmsg_endpoint struct */
    /* null buffer address */
    /* locally scoped memory */
    /* one byte before the valid shared memory range */
    /* within the rpmsg_hdr section */
    /* one byte outside shared memory */

    /*** rpmsg_release_rx_buffer ***/

    /* null rpmsg_endpoint struct */
    /* null buffer address */
    /* locally scoped memory */
    /* one byte before the valid shared memory range */
    /* within the rpmsg_hdr section */
    /* one byte outside shared memory */

    /*** rpmsg_send_offchannel_nocopy ***/

    /* null rpmsg_endpoint struct */
    /* null tx buffer */
    /* invalid length */
    /* bad destination address */
    /* use a bad buffer to send the data - locally scoped memory */
    /* use a bad buffer to send the data - one byte past the valid shared memory range */
    /* use a bad buffer to send the data - one byte before the valid shared memory range */
    /* use a bad buffer to send the data - within the rpmsg_hdr section */

lib/rpmsg/rpmsg_internal.h
```diff
@@ -34,6 +34,12 @@ extern "C" {
 	((struct rpmsg_hdr *)((unsigned char *)(p) - sizeof(struct rpmsg_hdr)))
 #define RPMSG_LOCATE_DATA(p) ((unsigned char *)(p) + sizeof(struct rpmsg_hdr))
 
+#define RPMSG_VALIDATE_BUF_ADDR(addr, rvdev, len) \
+	(((addr >= (((struct metal_io_region *)rvdev->rvq->shm_io)->virt + \
+	sizeof(struct rpmsg_hdr))) && \
+	((addr + len) < (((struct metal_io_region *)rvdev->rvq->shm_io)->virt + \
+	((struct metal_io_region *)rvdev->rvq->shm_io)->size))) ? 1 : 0)
+
```
arnopo, line 42: The metal_io_region structure should be manipulated in libmetal.
please create `metal_io_block_check ` or `metal_io_is_block_valid` API in libmetal instead

arnopo (changes requested):

lib/rpmsg/rpmsg_internal.h
tammyleino: @arnopo I have some questions since this will cause a dependency for OpenAMP on a specific version of libmetal.  From a PR merging standpoint, we can just ensure the two PRs in the different repos are merged at the same time, but from a release standpoint, how do we denote that OpenAMP version x is dependent on version y?  Or, do we strive to keep OpenAMP interoperable with older verisons of libmetal?  In which case, I would need to #define the libmetal macro to nothing within OpenAMP if it doesn't already exist in libmetal.

arnopo: Good question:
In term of merging, I will integrate the libmetal first, and rerun the OpenAMP build test to ensure that is working.

For the release. I don't expect to maintain interoperability with older versions of libmetal. 
I think we should only have a dependency message in the release note and offer to send a PR on the dedicated branch for the release if backwards compatibility is needed for a specific libmetal version.
I assume that libmetal APIs will be independent enough to be backported to a legacy branch...
@edmooring: what is your feeling?

edmooring: @arnopo: I agree. I think there has always been an implicit assumption that the libmetal and open amp libraries are tightly coupled. We should probably document this.

tammyleino:
Build will fail until https://github.com/OpenAMP/libmetal/pull/222 is merged, but I wanted to commit these changes so they can be reviewed in sync with the libmetal PR.

edmooring (approved):
Looks good to go.

PR #407, opened 2022-07-26
Vring descriptor count must be limited to 65k per device

tammyleino:
Code must ensure we don't overflow the 16-bit count in vring_alloc_info

### DIFF
--- a/lib/include/openamp/remoteproc_virtio.h
+++ b/lib/include/openamp/remoteproc_virtio.h
@@ -20,6 +20,9 @@
 extern "C" {
 #endif
 
+/* maximum number of vring descriptors for a vdev limited by 16-bit data type */
+#define	RPROC_MAX_VRING_DESC	USHRT_MAX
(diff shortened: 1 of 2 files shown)


lib/include/openamp/remoteproc_virtio.h
```diff
@@ -20,6 +20,9 @@
 extern "C" {
 #endif
 
+/* maximum number of vring descriptors for a vdev */
+#define	RPROC_MAX_VRING_DESC	65535
```
arnopo, line 24: could USHRT_MAX be used instead ?  and please specify in comment that limitation is linked to 16-bit format

tammyleino: @arnopo  Noted and addressed - please review - thanks!

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #408, opened 2022-07-27
Additional input parameter checks for APIs

tammyleino:
Added additional input parameter checks to APIs

### DIFF
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -161,6 +161,9 @@ int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
 static inline int rpmsg_send(struct rpmsg_endpoint *ept, const void *data,
 			     int len)
 {
+	if (!ept)
+		return RPMSG_ERR_PARAM;
+
 	return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
 					 len, true);
 }
@@ -184,6 +187,9 @@ static inline int rpmsg_send(struct rpmsg_endpoint *ept, const void *data,
 static inline int rpmsg_sendto(struct rpmsg_endpoint *ept, const void *data,
 			       int len, uint32_t dst)
 {
+	if (!ept)
+		return RPMSG_ERR_PARAM;
+
 	return rpmsg_send_offchannel_raw(ept, ept->addr, dst, data, len, true);
 }
 
@@ -229,6 +235,9 @@ static inline int rpmsg_send_offchannel(struct rpmsg_endpoint *ept,
 static inline int rpmsg_trysend(struct rpmsg_endpoint *ept, const void *data,
 				int len)
 {
+	if (!ept)
+		return RPMSG_ERR_PARAM;
+
 	return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
 					 len, false);
 }
@@ -252,6 +261,9 @@ static inline int rpmsg_trysend(struct rpmsg_endpoint *ept, const void *data,
 static inline int rpmsg_trysendto(struct rpmsg_endpoint *ept, const void *data,
 				  int len, uint32_t dst)
 {
+	if (!ept)
+		return RPMSG_ERR_PARAM;
+
 	return rpmsg_send_offchannel_raw(ept, ept->addr, dst, data, len, false);
 }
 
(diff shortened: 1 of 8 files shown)


tammyleino:
Discussion regarding these checks is greatly appreciated - some may be too stringent for all systems.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -797,15 +800,17 @@ void rpmsg_deinit_vdev(struct rpmsg_virtio_device *rvdev)
...
+		rdev = &rvdev->rdev;
+		while (!metal_list_is_empty(&rdev->endpoints)) {
+			node = rdev->endpoints.next;
+			ept = metal_container_of(node, struct rpmsg_endpoint, node);
+			rpmsg_destroy_ept(ept);
+		}
 
-	rvdev->rvq = 0;
-	rvdev->svq = 0;
+		rvdev->rvq = 0;
```
edmooring, line 811: Why the indentation change here?

edmooring (changes requested):
Aside from the indentation issue, looks good to go.

lib/rpmsg/rpmsg_virtio.c
tammyleino: The code is within the if conditional now, so it should be lined up as such.

lib/include/openamp/rpmsg.h
```diff
@@ -161,8 +161,11 @@ int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
 static inline int rpmsg_send(struct rpmsg_endpoint *ept, const void *data,
 			     int len)
 {
-	return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
-					 len, true);
+	if (ept)
+		return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
+						 len, true);
+	else
+		return RPMSG_ERR_PARAM;
```
arnopo, line 168: suggestion:
```
if (!ept)
	return RPMSG_ERR_PARAM;

return rpmsg_send_offchannel_raw(ept, ept->addr, ept->dest_addr, data,
						 len, true);

```
comment to apply for the other function below. when possible test input parameters first and return on error

lib/remoteproc/remoteproc.c
```diff
@@ -142,18 +142,23 @@ static int remoteproc_parse_rsc_table(struct remoteproc *rproc,
...
 			     struct resource_table *rsc_table,
 			     size_t rsc_size)
 {
 	int ret;
-	struct metal_io_region *io;
+	struct metal_io_region *io = NULL;
 
-	io = remoteproc_get_io_with_va(rproc, rsc_table);
+	if (rproc && rsc_table)
+		io = remoteproc_get_io_with_va(rproc, rsc_table);
```
arnopo, line 161: what about?

```
if (!rproc || !rsc_table || ! rsc_size)
	return -RPROC_EINVAL;

io = remoteproc_get_io_with_va(rproc, rsc_table);
if (!io)
	return -RPROC_EINVAL;
```

arnopo (changes requested):
Thank you for this robustness improvement?
To align code style would prefer that you test input parameter first. i put an example in comment for the` rpmsg_send_offchannel_raw` function

lib/rpmsg/rpmsg_virtio.c
edmooring: Got it. Looks good to go.

edmooring (approved):

lib/include/openamp/rpmsg.h
tammyleino: I am happy to change this per your request, but I generally test the most likely condition first in order to better optimize the code.  In your example, since it is more likely that ept will not be NULL, I am testing the most likely condition first.  Please let me know your thoughts.

lib/remoteproc/remoteproc.c
tammyleino: Sure, will change.

lib/include/openamp/rpmsg.h
arnopo: I''m wonder if something change in term of code optimization as it is a simple condition in which you have to test the ept for both case 
Do you check the delta?
that said defining "likely" & "unlikely" macro as done in Linux cool be something interesting for this kind of check.

arnopo (approved):

PR #409, opened 2022-07-28
Add missing macros for driver/device-only compilation

tammyleino:
Added macros to omit code when compiling for device-only.


edmooring (approved):
Looks good to me.

arnopo (approved):

PR #413, opened 2022-08-02
openamp: firstly take all buffer from shram pool

GUIDINGLI:
openamp: firstly take all buffer from shram pool

1. the memory already alloced for shram
2. this can help dump the right free cnt

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -161,12 +161,13 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
 
 #ifndef VIRTIO_DEVICE_ONLY
 	if (role == RPMSG_HOST) {
-		data = virtqueue_get_buffer(rvdev->svq, len, idx);
-		if (!data && rvdev->svq->vq_free_cnt) {
-			data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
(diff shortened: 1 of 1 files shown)


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -161,12 +161,13 @@ static void *rpmsg_virtio_get_tx_buffer(struct rpmsg_virtio_device *rvdev,
 
 #ifndef VIRTIO_DEVICE_ONLY
 	if (role == RPMSG_HOST) {
-		data = virtqueue_get_buffer(rvdev->svq, len, idx);
-		if (!data && rvdev->svq->vq_free_cnt) {
-			data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
-					rvdev->config.h2r_buf_size);
+		data = rpmsg_virtio_shm_pool_get_buffer(rvdev->shpool,
```
arnopo, line 164: What happens if  the size of the shared ram allow to allocate more than the max number of descriptors in vrings?
my concern is that some implementations shared the pool between different vdev instances, or simply use a part of the pool for some other topics.

And the check of` svq->vq_free_cnt` is obsolete since the introduction of the no-copy...

Could you introduce a counter to ensure that we do not allocate more than the number of vq descriptors defined?

arnopo (changes requested):

edmooring (changes requested):
The commit message doesn't explain why this change is good. Please expand the commit message to provide this explanation.

PR #429, opened 2022-10-12
Update rpc_demo description

TanmayShah-xilinx:
rpc_demo.c file creates firmware that runs on remote processor. The current description implies that proxy service is implemented on remote side which is not true. Fix description and explicitly mention that proxy service is implemented on host side.


arnopo (approved):

arnopo (approved):

edmooring (approved):
Looks good to me.

PR #431, opened 2022-10-13
README: Remove deprecated method to build open-amp lib for zephyr

arnopo:
The instructions to build open-amp for zephyr do not work. Instead of trying to maintain a cross dependency between the zephyr and the open-amp, just reference the zephyr project and documentation


carlocaione (approved):

wmamills (approved):
I am OK with this deletion as it is not the way to go.
In the future we need to show how to build Zephyr with upstream open-amp and libmetal and possible local changes.

wmamills:
BTW: I created an issue to track the above

edmooring (approved):
Looks good to me.

Issue #430, opened 2022-10-13
Possible problem of using strncmp to match endpoint name


Luoject:
In rpmsg_get_endpoint function, it uses strncmp to match endpoint name, the implementation is shown below:
`		if (name)
			name_match = !strncmp(ept->name, name,
					      sizeof(ept->name));`

It dose not match the whole name, which may be a problem. For example, the local ept->name is "ept", the param name is "ept_1" or "ept_2", both of the param name will match the local one. This is not as expected.

Should we use strcmp instead?

Luoject:
Here is my pull request: https://github.com/OpenAMP/open-amp/pull/432

Please take time to review, thanks!

Luoject:
sizeof(ept->name) is the same as RPMSG_NAME_SIZE. There's no problem. I will close the pull request.

PR #432, opened 2022-10-13
lib: Use strcmp to match endpoint name instead of strncmp

Luoject:
It's not accurate to use strncmp to match endpoint name, because it dosen't match the whole word.
Instead, it's recommended to use strcmp.

Related issue: https://github.com/OpenAMP/open-amp/issues/430

### DIFF
--- a/lib/rpmsg/rpmsg.c
+++ b/lib/rpmsg/rpmsg.c
@@ -219,8 +219,7 @@ struct rpmsg_endpoint *rpmsg_get_endpoint(struct rpmsg_device *rdev,
 			return ept;
 		/* else use name service and destination address */
 		if (name)
-			name_match = !strncmp(ept->name, name,
-					      sizeof(ept->name));
+			name_match = !strcmp(ept->name, name);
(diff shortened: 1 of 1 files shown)


lib/rpmsg/rpmsg.c
```diff
@@ -219,8 +219,7 @@ struct rpmsg_endpoint *rpmsg_get_endpoint(struct rpmsg_device *rdev,
 			return ept;
 		/* else use name service and destination address */
 		if (name)
-			name_match = !strncmp(ept->name, name,
-					      sizeof(ept->name));
+			name_match = !strcmp(ept->name, name);
```
arnopo, line 222: What about using 
`name_match = !strncmp(ept->name, name, RPMSG_NAME_SIZE)`?
We cannot guaranty that the names will finish with the `'\0'` ...

Luoject: sizeof(ept->name) is the same as RPMSG_NAME_SIZE. There's no problem.

PR #438, opened 2022-10-29
zynqmp_r5 resource table: Change notifyid for the virtio device.

edmooring:
The notifyid for the virtio device in the resource table was set to 0, and the notifyids for the vrings were set to 1 and 2. However, the Linux remoteproc implementation changes those values to 0 and 1 when the firmware image is loaded. With the enhanced error checking introduced in commit 2016ff1e, the duplicate assignment of notifyid 0 to the rvdev and vring0 resulted in a failure.

Set the value of notifyid for the virtio device (rvdev) to 31 to avoid possible conflicts with the Linux imposed values.

### DIFF
--- a/apps/machine/zynqmp_r5/rsc_table.c
+++ b/apps/machine/zynqmp_r5/rsc_table.c
@@ -51,7 +51,7 @@ struct remote_resource_table __resource resources = {
 
 	/* Virtio device entry */
 	{
-	 RSC_VDEV, VIRTIO_ID_RPMSG_, 0, RPMSG_VDEV_DFEATURES, 0, 0, 0,
+	 RSC_VDEV, VIRTIO_ID_RPMSG_, 31, RPMSG_VDEV_DFEATURES, 0, 0, 0,
 	 NUM_VRINGS, {0, 0},
 	 },
 


arnopo:
Hi Ed,

To share the information: this PR is related to an issue detected during non regression test for the **V2022.10 releas**e.

That said,  the issue is not clear for me and  I do not reproduce the  issue in my test with same resource table configuration.

I wonder if your PR is not  a workaround that masks the real issue.
I would prefer instead to revert the commit that introduce this and create an issue to properly fix this later .

**Anyway I think it is important here to understand the issue before we release the v2022.10.** 

Could you details the issue you are facing. Which function returns the error? 
Please also confirm that you are using the remoteproc framework of the Linux Kernel to load the firmware?
and which version of the Linux Kernel?

bentheredonethat:
can we check if this affects the current demos for both kernel and rpmsg-userspace demos?

edmooring:
@bentheredonethat I think this will affect all firmware that is loaded via the Linux remoteproc infrastructure, and uses 0 for the notifyid for the rpmsg_vdev in the resource table.

edmooring:
This is a more complete explanation of what I think is going on:
1. At firmware build time, the resource table is set up with notifyid==0 for the rpmsg virtio device. The vrings are set up with notifyid==1 for vring0 and notifyid==2 for vring1. I have verified this with objdump on the ELF file.
2. When the firmware is loaded onto the R5 via the Linux remoteproc sysfs interface, the kernel remoteproc implementation ignores the notifyid values for the vrings, and reallocates them, changing the resource table. On the Xilinx configuration I tested on, this results in vring0 being allocated notifyid 0, and vring1 being allocated notifyid 1. I have verified this with debugging prints.
3. When the remote processor is started, it parses the resource table.  When it gets to the virtio device, it calls handle_vdev_rsc()https://github.com/OpenAMP/open-amp/blob/a411066468ca4ebea855640542b71253600b3453/lib/remoteproc/rsc_table_parser.c#L127, which in turn calls remoteproc_allocate_id()https://github.com/OpenAMP/open-amp/blob/a411066468ca4ebea855640542b71253600b3453/lib/remoteproc/remoteproc.c#L870 with the vdev's notifyid of 0. This successfully returns 0.
Then handle_vdev_rsc() goes through the vrings. On vring0, it calls remoteproc_allocate_id(), with notifyid set to 0, because Linux changed vring0.notifyid from 1 to 0 in (2) above. This results in remoteproc_allocate_id() returning RSC_NOTIFY_ID_ANY. This causes handle_vdev_rsc() to return -RPROC_ERR_RSC_TAB_NP, which causes the firmware startup to fail.
This appeared to work in previous versions, because remoteproc_allocate_id() returning RSC_NOTIFY_ID_ANY was effectively ignored, leaving the notifyid unchanged and allowing execution to continue. I don't know what effect, if any, the duplicate notifyid in the vdev and vring0 has. The Linux remoteproc implementation does not appear to use the notifyid in the device at all.
Note, I was wrong about which commit triggers the error, it's actually 03c80a13.

arnopo:
> This is a more complete explanation of what I think is going on:

Thanks for the clarifications!

> 
> 1. At firmware build time, the resource table is set up with notifyid==0 for the rpmsg virtio device. The vrings are set up with notifyid==1 for vring0 and notifyid==2 for vring1. I have verified this with objdump on the ELF file.
> 2. When the firmware is loaded onto the R5 via the Linux remoteproc sysfs interface, the kernel remoteproc implementation ignores the notifyid values for the vrings, and reallocates them, changing the resource table. On the Xilinx configuration I tested on, this results in vring0 being allocated notifyid 0, and vring1 being allocated notifyid 1. I have verified this with debugging prints.

Right, this is specified here: https://elixir.bootlin.com/linux/latest/source/drivers/remoteproc/remoteproc_core.c#L367

> 3. When the remote processor is started, it parses the resource table.  When it gets to the virtio device, it calls handle_vdev_rsc()https://github.com/OpenAMP/open-amp/blob/a411066468ca4ebea855640542b71253600b3453/lib/remoteproc/rsc_table_parser.c#L127
>    , which in turn calls remoteproc_allocate_id()https://github.com/OpenAMP/open-amp/blob/a411066468ca4ebea855640542b71253600b3453/lib/remoteproc/remoteproc.c#L870
>     with the vdev's notifyid of 0. This successfully returns 0.
>    Then handle_vdev_rsc() goes through the vrings. On vring0, it calls remoteproc_allocate_id(), with notifyid set to 0, because Linux changed vring0.notifyid from 1 to 0 in (2) above. This results in remoteproc_allocate_id() returning RSC_NOTIFY_ID_ANY. This causes handle_vdev_rsc() to return -RPROC_ERR_RSC_TAB_NP, which causes the firmware startup to fail.

Ok, Now I think I understand. The behavior seems related to the Xilinx implementation.
In the remote fw  the `platform_create_proc()` function calls `remoteproc_set_rsc_table()` function. 
This supposes that the Linux Kernel has finished to update the resource table when the remote firmware parses it.
This seems working if the remote processor firmware is loaded by Linux. This does not work if the coprocessor is started before the Linux.
Seems to me not safe for the remote processor to trust the content of the vdev and vrings structures in the resource table until the` vdev->status` is set to `VIRTIO_CONFIG_STATUS_DRIVER_OK`.

And that's why I do not reproduce the issue on the STM32MP1 platform with Zephyr or in baremetal. Indeed, we don't use the remoteproc rproc structure on remote. we directly call `rproc_virtio_create_vdev`, no `remoteproc_allocate_id() `call.

That said the point remains valid.  And from my POV, if we would like to fix it, it should be in Linux.

>    This appeared to work in previous versions, because remoteproc_allocate_id() returning RSC_NOTIFY_ID_ANY was effectively ignored, leaving the notifyid unchanged and allowing execution to continue. I don't know what effect, if any, the duplicate notifyid in the vdev and vring0 has. The Linux remoteproc implementation does not appear to use the notifyid in the device at all.
>    Note, I was wrong about which commit triggers the error, it's actually [03c80a1](https://github.com/OpenAMP/open-amp/commit/03c80a13417f67f94852cf7d5cb160f1bdf8271e).

Regarding the analysis, I think the issue in not blocker from open-amp library perspective.
I propose to discuss the point today during the remoteproc bi-weekly meeting.

arnopo (approved):

PR #439, opened 2022-11-06
Fix warn: declaration of 'vring_rsc' shadows a previous local

xiaoxiang781216:
```
 regressed by commit 03c80a13417f67f94852cf7d5cb160f1bdf8271e
    Author: Tammy Leino <tammy_leino@mentor.com>
    Date:   Mon Sep 26 08:14:35 2022 -0700
    
        handle_vdev_rsc must return error if notifyid cannot be assigned
    
        Updated handle_vdev_rsc to return error if a unique id cannot be assigned
```


arnopo (approved):
Good catch! thanks!
We need to add add the `-Wshadow` build flag in our test environment

edmooring (approved):
Good catch.

arnopo:
I also updated the v2022.10 branch, we will create a 2022.10.1 tag with a version 1.3.1 at the end of the month to fix the release

PR #451, opened 2022-11-22
CI: fix Zephyr build issue resulting from the introduction of sysbuild

arnopo:
"Sysbuild is a higher-level build system that can be used to combine multiple other build systems together"

Add the option by default for all platform as it is compatible with a single image.


danmilea (approved):
This fixes the build issue.

PR #465, opened 2023-02-14
BUG FIX: rpmsg: add cache flush when hold rx buffer

GUIDINGLI:
rpmsg: add cache flush when hold rx buffer

Here is the bug:

Assume we have 2 cpus, and use cached shram buffer

```
CPU0                        CPU1

1. send tx bufferX  
                              2. recv rx bufferX  
                              3. set idx to hdr->reserved 
                              4. handled rx bufferX 
                              5. return bufferX
6. reuse tx bufferX 
                              7. dirty cache auto flushed cause by step 3, hdr changed by CPU1
8. bufferX meet error
```

So, can we use the cache flush to resolve the issue or other ways ?

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -93,6 +93,11 @@ static void rpmsg_virtio_return_buffer(struct rpmsg_virtio_device *rvdev,
 				       uint16_t idx)
 {
 	unsigned int role = rpmsg_virtio_get_role(rvdev);
+
+#ifdef VIRTIO_CACHED_BUFFERS
+	metal_cache_invalidate(buffer, len);
+#endif
+
 #ifndef VIRTIO_DEVICE_ONLY
 	if (role == RPMSG_HOST) {
 		struct virtqueue_buf vqbuf;


GUIDINGLI:
@arnopo 

Please help to review this, thanks

GUIDINGLI:
@jjmcdn @edgarigl @ulfalizer @carlocaione 

Please help to review this, thanks

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -583,7 +581,15 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
...
 
 		/* Check whether callback wants to hold buffer */
-		if (!(rp_hdr->reserved & RPMSG_BUF_HELD)) {
+		if (rp_hdr->reserved & RPMSG_BUF_HELD) {
+			/* Yes, save idx. */
+			rp_hdr->reserved |= idx;
+
+#ifdef VIRTIO_CACHED_BUFFERS
+			metal_cache_flush(rp_hdr, sizeof(struct rpmsg_hdr));
+#endif
```
carlocaione, line 590: I'm a bit confused. Why do you need to flush the header here if the buffer is being held? I would expect that to be flushed when the buffer will be released and ready again for consumption.

GUIDINGLI: Because we modify the RX hdr data when hold the RX buffer.
Details please see this PR's description.

carlocaione: The PR's description is really confusing to me. If we are holding the RX buffer that means that the buffer is being used by (for example) CPU1 and it will be released at a later time, so why it is important to flush the header at this stage?

GUIDINGLI: If we are holding the RX buffer that means that the buffer is being used by (for example) CPU1
Yes, this time we do:
3. set idx to hdr->reserved 
And this time, the hdr is cache-able for CPU1, the step 3 write action probably just write to cache (non write to real RAM), 
Then after step 5, the buffer returned to CPU0, but the CPU1 dirty cache (which caused by step 3 write action) auto flushed.
Then we meet error.

arnopo: The need seems valid to me,  but could not be sufficient. If the RPMsg client updates the rpmsg payload this could lead to  payload corruption on auto flush.
Seems to me that flushing the cache associated to the full buffer in `rpmsg_virtio_return_buffer` would be more reliable...

GUIDINGLI:
@arnopo 
I use the cache_invalidate to resolve this issue, please review

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -93,6 +93,11 @@ static void rpmsg_virtio_return_buffer(struct rpmsg_virtio_device *rvdev,
 				       uint16_t idx)
 {
 	unsigned int role = rpmsg_virtio_get_role(rvdev);
+
+#ifdef VIRTIO_CACHED_BUFFERS
+	metal_cache_invalidate(buffer, len);
```
arnopo, line 98: Any reason to use `metal_cache_invalidate` instead of `metal_cache_flush`?
Both seems Ok to me as the buffers are returned after their use...

arnopo (approved):

edmooring (approved):
Looks good to me.

PR #466, opened 2023-02-16
BUGFIX: rpmsg: do cache_invalidate() when real data returned

GUIDINGLI:
rpmsg: do cache_invalidate() when real data returned


arnopo (approved):

edmooring (approved):
Looks good to go.

PR #472, opened 2023-03-01
BUGFIX: change unsuitable atomic_int to atomic_flag

mzyx-hnu:
This patch tries to fix issue#471

### DIFF
--- a/apps/examples/linux_rpc_demo/linux_rpc_demo.c
+++ b/apps/examples/linux_rpc_demo/linux_rpc_demo.c
@@ -36,7 +36,7 @@
 static struct rpmsg_rpc_clt *rpmsg_default_rpc;
 static int file_d, bytes_written, bytes_read;
 static struct polling poll;
-static atomic_int wait_resp;
+static atomic_flag wait_resp;
 
 static void rpmsg_rpc_shutdown(struct rpmsg_rpc_clt *rpc)
 {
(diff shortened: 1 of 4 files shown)


apps/examples/linux_rpc_demo/linux_rpc_demo.c
```diff
@@ -464,7 +464,7 @@ int app(struct rpmsg_device *rdev, void *priv)
 	/* redirect I/Os */
 	LPRINTF("Initializating I/Os redirection...\r\n");
 	table_len = (int)sizeof(rpc_table) / sizeof(struct rpmsg_rpc_services);
-	atomic_init(&wait_resp, 1);
+	atomic_flag_test_and_set(&wait_resp);
```
arnopo, line 467: I would expect here  &wait_resp = ATOMIC_FLAG_INIT 
```
wait_resp = (atomic_flag)ATOMIC_FLAG_INIT;
atomic_flag_test_and_set(&wait_resp)
```

lib/proxy/rpmsg_retarget.c
```diff
@@ -85,7 +85,7 @@ int rpmsg_rpc_init(struct rpmsg_rpc_data *rpc,
 	rpc->ept_destroyed = 0;
 	rpc->respbuf = NULL;
 	rpc->respbuf_len = 0;
-	atomic_init(&rpc->nacked, 1);
+	atomic_flag_test_and_set(&rpc->nacked);
```
arnopo, line 88: same here

arnopo (changes requested):
The fix seems valid to me. I wonder if we need to have a `atomic_flag_test_and_set` on init,  but that seems inline with the initial value set by `atomic_init`...

One cosmetic remark:
Please, split in 2 commits: one for the lib one` for the app. have look to existing commit to see commit subjet prefix to use

edmooring (commented):
Not much to say here. I think this will be good to go with @arnopo comments addressed.

mzyx-hnu:
> The fix seems valid to me. I wonder if we need to have a `atomic_flag_test_and_set` on init, but that seems inline with the initial value set by `atomic_init`...
> 
> One cosmetic remark: Please, split in 2 commits: one for the lib one` for the app. have look to existing commit to see commit subjet prefix to use

updated as the above comments

arnopo (approved):

edmooring (approved):
Good to go.

PR #497, opened 2023-07-05
CI: Fix checkpatch

arnopo:
The CI does not consider "check" reported by checkpatch. This commit fixes it by:

- rebasing check_compliance.py compliance.yml to integrate Zephyr updates,
- add detection of the "check" in the report.

An exemple of the result is available here: https://github.com/arnopo/open-amp/pull/25

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -16,11 +16,15 @@ jobs:
     name: checkpatch review
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - name: Checkout the code
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
     - name: Install python dependencies
       run: |
         pip3 install setuptools
-        pip3 install junitparser==1.6.3 gitlint codespell
+        pip3 install python-magic junitparser gitlint codespell lxml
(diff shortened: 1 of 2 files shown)


.github/workflows/compliance.yml
```diff
@@ -16,11 +16,15 @@ jobs:
...
+    - name: Checkout the code
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
     - name: Install python dependencies
       run: |
         pip3 install setuptools
-        pip3 install junitparser==1.6.3 gitlint codespell
+        pip3 install junitparser gitlint codespell lxml
```
danmilea, line 27: Not mandatory: add python-magic to the list of python packages explicitly installed as a reference for check_compliance.py users. I don't think this happens on ubuntu-latest, so this action might work on github.

In a clean puthon3.11 virtual environment python-magic is also required, otherwise check_compliance.py fails.
./scripts/ci/check_compliance.py", line 20, in <module>
    import magic
ModuleNotFoundError: No module named 'magic'

Otherwise, good to merge.

danmilea (changes requested):
I have a single quality of life comment which is not mandatory. Otherwise, good to merge.

danmilea:
Thanks for the update @arnopo . Good to merge.

PR #500, opened 2023-07-12
remoteproc: do cache invalidation before reading rsc_table status

iuliana-prodan:
Do a cache invalidation before reading the resource table's status since this ca be in a cacheable region.


iuliana-prodan:
cc @carlocaione @arnopo

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -40,6 +41,11 @@ static unsigned char rproc_virtio_get_status(struct virtio_device *vdev)
 	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
 	vdev_rsc = rpvdev->vdev_rsc;
 	io = rpvdev->vdev_rsc_io;
+
+#ifdef VIRTIO_CACHED_RSC_TABLE
```
TanmayShah-xilinx, line 45: Can we move #ifdef part to header file ?
May be we can introduce function to check if cache is on or not and then based on that we can decide if we need to call metal_cache_invalidate or not.

iuliana-prodan: So far, in OpenAMP there are 2 ways of doing cache invalidation:
-  rpmsg: https://github.com/OpenAMP/open-amp/blob/main/lib/rpmsg/rpmsg_virtio.c#L138
- virtqueue: https://github.com/OpenAMP/open-amp/blob/main/lib/virtio/virtqueue.c#L32

I've chose the first one.
You're suggesting the second one - just to put the defines in a header, not in .c file, right?

iuliana-prodan: @carlocaione yes, will do like this.

carlocaione: > @carlocaione yes, will do like this.

bonus point if you can clean up these two ways of doing the same thing :stuck_out_tongue_winking_eye: (in a different PR)

cmake/options.cmake
```diff
@@ -90,6 +90,12 @@ if (WITH_DCACHE_BUFFERS)
   add_definitions(-DVIRTIO_CACHED_BUFFERS)
 endif (WITH_DCACHE_BUFFERS)
 
+option (WITH_DCACHE_RSC_TABLE "Build with resource table cache operations enabled" OFF)
+
+if (WITH_DCACHE_RSC_TABLE)
+  add_definitions(-DVIRTIO_CACHED_RSC_TABLE)
```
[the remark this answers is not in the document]
iuliana-prodan: Well.. this is tricky - why we have 2 options now? Why don't we use only one?
I believe it's better to treat them separately because each has it's own memory assigned in the dts.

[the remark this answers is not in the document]
arnopo: We can also assume that most of the project would enable the cache for vrings, buffers and resource table.
To simplify the use, perhaps it would be nice to introduce an option to enable cache for all memories.

```
option (WITH_DCACHE "Build with all cache operations enabled" OFF)

option (WITH_DCACHE_VRINGS "Build with vrings cache operations enabled" OFF)
if (WITH_DCACHE_VRINGS OR WITH_DCACHE)
  add_definitions(-DVIRTIO_CACHED_VRINGS)
endif (WITH_DCACHE_VRINGS)
option (WITH_DCACHE_BUFFERS "Build with vrings cache operations enabled" OFF)
if (WITH_DCACHE_BUFFERS OR WITH_DCACHE)
  add_definitions(-DVIRTIO_CACHED_BUFFERS)
endif (WITH_DCACHE_BUFFERS)

option (WITH_DCACHE_RSC_TABLE "Build with resource table cache operations enabled" OFF)

if (WITH_DCACHE_RSC_TABLE OR WITH_DCACHE)
  add_definitions(-DVIRTIO_CACHED_RSC_TABLE)
endif (WITH_DCACHE_RSC_TABLE)

```

i just also notice an error in description (It a bonus that can be fixed later, as @carlocaione  said)
```
- option (WITH_DCACHE_BUFFERS "Build with vrings cache operations enabled" OFF)
+ option (WITH_DCACHE_BUFFERS "Build with buffers cache operations enabled" OFF)
```

lib/remoteproc/remoteproc_virtio.c
TanmayShah-xilinx: Thanks. Yes second way works.

cmake/options.cmake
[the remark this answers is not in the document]
TanmayShah-xilinx: @arnopo I agree. We can assume for whole platform either cache is enabled or not. Anyway, There are many configurations already available for libmetal. It's good idea to combine all three in one. However, we can do that cleanup in separate PR, doesn't have to be as part of this one.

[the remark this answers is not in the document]
iuliana-prodan: @arnopo @carlocaione @TanmayShah-xilinx in this PR I'll just add a new option - VIRTIO_CACHED_RSC_TABLE.

I'll follow up with the other suggestions in a separate PR - updates & clean-up with:
- add a new option that enables all 3 cache options - RINGS, BUFFERS, RSC_TABLE;
- s/vrings/buffers for WITH_DCACHE_BUFFERS option;
- use defines for cache functions for BUFFERS, the same as for RINGS and RSC_TABLE.

carlocaione (approved):
LGTM but looking forward to the cleanup PR.

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -23,6 +24,15 @@ extern "C" {
 /* maximum number of vring descriptors for a vdev limited by 16-bit data type */
 #define	RPROC_MAX_VRING_DESC	USHRT_MAX
 
+/* cache invalidation helpers for resource table */
+#ifdef VIRTIO_CACHED_RSC_TABLE
+#define RSC_TABLE_FLUSH(x)		metal_cache_flush(&x, sizeof(x))
+#define RSC_TABLE_INVALIDATE(x)		metal_cache_invalidate(&x, sizeof(x))
```
arnopo, line 30: checkpatch check has ben reinforced. To fix the issue, this should work
```
#define RSC_TABLE_FLUSH(x) ({ void *_x = x; metal_cache_flush(&_x, sizeof(_x))})
#define RSC_TABLE_INVALIDATE(x) ({ void *_x = x; metal_cache_invalidate(&_x, sizeof(_x))}) 
```

iuliana-prodan:
Changes since v3:
- fix checkpatch warning;
- update README in new commit.

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -23,6 +24,19 @@ extern "C" {
 /* maximum number of vring descriptors for a vdev limited by 16-bit data type */
 #define	RPROC_MAX_VRING_DESC	USHRT_MAX
 
+/* cache invalidation helpers for resource table */
+#ifdef VIRTIO_CACHED_RSC_TABLE
+#define RSC_TABLE_FLUSH(x) \
+	({void *_x = (x); \
+	metal_cache_flush(&_x, sizeof(_x)); })
```
carlocaione, line 31: Are you sure this is doing what you think it is doing? Looking at it it looks like you are flushing the pointer here, not the actual struct also with a wrong `sizeof()`.

carlocaione (changes requested):

lib/include/openamp/remoteproc_virtio.h
iuliana-prodan: You're right. I've fixed it.
Also, to pass checkpatch I've added 2 arguments for RSC_TABLE_FLUSH and RSC_TABLE_INVALIDATE

iuliana-prodan:
Changes since v4:
- in order to fix checkpatch warnings I've added 2 arguments for RSC_TABLE_FLUSH and RSC_TABLE_INVALIDATE,

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -40,6 +40,9 @@ static unsigned char rproc_virtio_get_status(struct virtio_device *vdev)
 	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
 	vdev_rsc = rpvdev->vdev_rsc;
 	io = rpvdev->vdev_rsc_io;
+
+	RSC_TABLE_INVALIDATE(vdev_rsc, sizeof(vdev_rsc));
```
carlocaione, line 44: I don't think this is correct either.

This is expanding to:
```
struct fw_rsc_vdev *vdev_rsc;

metal_cache_flush(&vdev_rsc, sizeof(vdev_rsc))
```
That is you are flushing the pointer, not the actual table.

The correct expansion should be;
```
metal_cache_flush(vdev_rsc, sizeof(*vdev_rsc));
```

carlocaione (changes requested):

lib/remoteproc/remoteproc_virtio.c
iuliana-prodan: Right. 
I added a new version fixing everything.

iuliana-prodan:
Changes since v5:
- call RSC_TABLE_INVALIDATE with sizeof(struct..);
- do not use address of the pointer for metal_cache_flush/metal_cache_invalidate;

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -40,6 +40,9 @@ static unsigned char rproc_virtio_get_status(struct virtio_device *vdev)
 	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
 	vdev_rsc = rpvdev->vdev_rsc;
 	io = rpvdev->vdev_rsc_io;
+
+	RSC_TABLE_INVALIDATE(vdev_rsc, sizeof(struct fw_rsc_vdev));
+
```
carlocaione, line 45: This is fine now, thanks.

Shouldn't we also `RSC_TABLE_FLUSH` in `rproc_virtio_set_status()` though?

carlocaione (changes requested):

lib/remoteproc/remoteproc_virtio.c
carlocaione: @arnopo I'm not familiar with remoteproc at all. What are the structs that are actually r/w from both sides that needs to be flushed / invalidated?

TanmayShah-xilinx: @iuliana-prodan, @carlocaione 

Yes we need cache operation even in read config and write config as well.
One more question, should we perform only flush or invalidate and flush both during writing resource table? Or it is really platform dependent?

arnopo: "Flush" and "invalidate" should be sufficient if the structures are aligned on a cache line. We can improve this in the future if needed.

arnopo: Right @TanmayShah-xilinx, 
The cache shoule also be managed in `rproc_virtio_read_config` and`rproc_virtio_read_config`

iuliana-prodan:
Changes since v6:

- do flush/invalidate also for write/read config;

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -73,6 +77,9 @@ static uint32_t rproc_virtio_get_dfeatures(struct virtio_device *vdev)
...
```
carlocaione, line 82: nit: can you keep the newlines consistent among all the `RSC_TABLE_*` calls?

carlocaione (approved):
In general this LGTM. I'm a bit concerned about the alignment of these structs. If the structs are not aligned to the cache line, the invalidation will corrupt neighboring structs in memory. It really is a silent killer.

See for example https://github.com/zephyrproject-rtos/zephyr/pull/59456

iuliana-prodan:
Changes since v7:
keep the newlines consistent among all the RSC_TABLE_* calls;

carlocaione (approved):

arnopo (approved):
LGTM

PR #503, opened 2023-08-09
Support for baremetal R5 in TI K3 family devices 

kwillis01:
Added support for RPMsg on the baremetal R5 in the TI SK-AM64. This can be extended to other devices in the TI K3 family by changing addresses in the linker and platform_info files. A mailbox driver was added to kick RPMsg for these devices.


kwillis01:
@tnmysh what should I do with the ti_k3_r5.cmake and CMakeLists.txt I edited when moving to the system reference? Should I just add in the ti_k3_r5.cmake and add a note to change the CMakeLists.txt line?

kwillis01:
@tnmysh Okay I will do that. What should I do with the ti_k3_r5.cmake and CMakeLists.txt I edited when moving to the system reference? Should I just add in the ti_k3_r5.cmake and add a note to change the CMakeLists.txt line?

tnmysh:
> @tnmysh Okay I will do that. What should I do with the ti_k3_r5.cmake and CMakeLists.txt I edited when moving to the system reference? Should I just add in the ti_k3_r5.cmake and add a note to change the CMakeLists.txt line?

I did not get the question. We will link ope-amp library statically with baremetal code. So, I don't think you should modify CMakeLists.txt at all.

glneo:
I'll re-open this PR over on the `openamp-system-reference` repo once the current demos from here and libmetal are moved over. This can be closed.

PR #505, opened 2023-10-02
Refactor VRING macros for AMP VIRTIO

danmilea:
Renamed AMP VIRTIO macros to avoid conflicts.


lib/include/openamp/virtio.h
```diff
@@ -152,7 +152,7 @@ struct virtio_device_id {
 	sizeof(struct vring_used_elem) * n + sizeof(uint16_t))
 
 #define VRING_DECLARE(name, n, align) \
-static char __vrbuf_##name[VIRTIO_RING_SIZE(n, align)] __aligned(VRING_ALIGNMENT); \
+static char __vrbuf_##name[VIRTIO_RING_SIZE(n, align)] __aligned(VIRTIO_AMP_VRING_ALIGNMENT); \
```
arnopo, line 155: If i well understood here you expect to reserve vrings memory with a `VIRTIO_AMP_VRING_ALIGNMENT` Alignement. to guaranty that a 4K page is reserved for one vring.

What about  using a parammeter instead of hardcoding it? 
And how do you ensure that the host is aligned with the __vrbuf_##name address?
As there is not usage yet of this macro, I would be in favor of removing it and reintroduce it  when used.
else we should maintain it evenif not usable in a generic way.

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -286,7 +286,7 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 
 	vring_info->io = vmdev->shm_io;
 	vring_info->info.num_descs = virtio_mmio_get_max_elem(vdev, idx);
-	vring_info->info.align = VRING_ALIGNMENT;
+	vring_info->info.align = VIRTIO_AMP_VRING_ALIGNMENT;
```
arnopo, line 289: The `align` is used by `vq_ring_init`.  That's means that you align the `vring_used` struct on a new 4k page right?
That would mean that the vring is on two 4k page...

arnopo (commented):
Thank @danmilea for the reactivity. few comments/questions

lib/include/openamp/virtio.h
danmilea: I have removed the define in the newer version of the PR.

lib/include/openamp/virtio.h
```diff
@@ -136,29 +136,7 @@ struct virtio_device_id {
...
-
-#define VRING_DECLARE(name, n, align) \
-static char __vrbuf_##name[VIRTIO_RING_SIZE(n, align)] __aligned(VRING_ALIGNMENT); \
-static struct vring __vring_##name = { \
-	.desc = (void *)__vrbuf_##name, \
-	.avail = (void *)((unsigned long)__vrbuf_##name + n * sizeof(struct vring_desc)), \
-	.used = (void *)((unsigned long)__vrbuf_##name + ((n * sizeof(struct vring_desc) + \
-		(n + 1) * sizeof(uint16_t) + align - 1) & ~(align - 1))), \
-}
+#define VIRTIO_AMP_VRING_ALIGNMENT           4096
```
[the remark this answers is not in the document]
danmilea: Done.

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #508, opened 2023-10-09
rpmsg: add release cb and refcnt in endpoint to fix ept used-after-free

yintao707:
issue description:​

4nOaD0QVZj

In our case, the` rpmsg_virtio_rx_callback()` is called in a rpmsg thread. Let's assume a situation, when `my_cb `has already got the ept in rpmsg thread,  user service thread just right called `my_deinit` to excute rpmsg_destroy_ept and released ept, at this time, the execution of ept ->cb has not been completed in rpmsg thread. so , there is a used after free about the ept.

Assuming we choose to fix this situation at the user level, we can continue to refer to the following examples:
```
void my_cb(...)
{
   atomic_fetch_add(&g_priv->refcnt, 1); // 2*
   ....
  if (atomic_fetch_sub(&g_priv->refcnt, 1) == 1) {
    rpmsg_destroy_ept(&g_priv->ept);
    free(g_priv);
  }
}

void my_init(void)
{
  g_priv = malloc(sizeof(*g_priv));
  atomic_init(&g_priv->refcnt, 1);
  g_priv->ept.cb = my_cb;
  rpmsg_create_ept(&g_priv->ept);
}

void my_deinit(void)
{
  if (atomic_fetch_sub(&g_priv->refcnt, 1) == 1) { // 1*
    rpmsg_destroy_ept(&g_priv->ept);
    free(g_priv);
  }
}
```
1. rpmsg_virtio_rx_callback release lock here: https://github.com/OpenAMP/open-amp/blob/cd8823876fb6920571abbeecc570a9da0cb546ba/lib/rpmsg/rpmsg_virtio.c#L561C1-L561C36
2. os decide to suspend rpmsg_virtio_rx_callback and switch to my_deinit
3. my_deinit finish execution and release g_priv(1*)
4. rpmsg_virtio_rx_callback resume and then happen use-after-free

The root cause is that OpenAMP call cb after release lock
Therefore, reference counting in the application layer cannot solve this race condition. Therefore, to avoid race condition, we added refnt to the endpoint and call the release callback when ept callback finished

### DIFF
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -50,6 +50,7 @@ struct rpmsg_device;
 /* Returns positive value on success or negative error value on failure */
 typedef int (*rpmsg_ept_cb)(struct rpmsg_endpoint *ept, void *data,
 			    size_t len, uint32_t src, void *priv);
+typedef void (*rpmsg_ept_release_cb)(struct rpmsg_endpoint *ept);
 typedef void (*rpmsg_ns_unbind_cb)(struct rpmsg_endpoint *ept);
 typedef void (*rpmsg_ns_bind_cb)(struct rpmsg_device *rdev,
 				 const char *name, uint32_t dest);
@@ -72,6 +73,12 @@ struct rpmsg_endpoint {
 
 	/** Address of the default remote endpoint binded */
 	uint32_t dest_addr;
+
+	/** Reference count for determining whether the endpoint can be deallocated */
+	uint32_t refcnt;
+
+	/** Callback to inform the user that the endpoint allocation can be safely removed */
+	rpmsg_ept_release_cb release_cb;
 
 	/**
 	 * User rx callback, return value of this callback is reserved for future
--- a/lib/rpmsg/rpmsg.c
+++ b/lib/rpmsg/rpmsg.c
@@ -97,6 +97,25 @@ static int rpmsg_set_address(unsigned long *bitmap, int size, int addr)
 	}
 }
 
+void rpmsg_ept_incref(struct rpmsg_endpoint *ept)
+{
+	if (ept)
+		ept->refcnt++;
+}
+
+void rpmsg_ept_decref(struct rpmsg_endpoint *ept)
+{
+	if (ept) {
+		ept->refcnt--;
+		if (!ept->refcnt) {
+			if (ept->release_cb)
+				ept->release_cb(ept);
+			else
+				ept->rdev = NULL;
+		}
+	}
+}
+
 int rpmsg_send_offchannel_raw(struct rpmsg_endpoint *ept, uint32_t src,
 			      uint32_t dst, const void *data, int len,
 			      int wait)
@@ -245,7 +264,7 @@ static void rpmsg_unregister_endpoint(struct rpmsg_endpoint *ept)
 		rpmsg_release_address(rdev->bitmap, RPMSG_ADDR_BMP_SIZE,
 				      ept->addr);
 	metal_list_del(&ept->node);
-	ept->rdev = NULL;
+	rpmsg_ept_decref(ept);
 	metal_mutex_release(&rdev->lock);
 }
 
@@ -257,6 +276,7 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
 			     rpmsg_ns_unbind_cb ns_unbind_cb)
 {
 	strncpy(ept->name, name ? name : "", sizeof(ept->name));
+	ept->refcnt = 1;
 	ept->addr = src;
 	ept->dest_addr = dest;
 	ept->cb = cb;
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -514,6 +514,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		/* Get the channel node from the remote device channels list. */
 		metal_mutex_acquire(&rdev->lock);
 		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
+		rpmsg_ept_incref(ept);
 		metal_mutex_release(&rdev->lock);
 
 		if (ept) {
@@ -532,6 +533,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		}
 
 		metal_mutex_acquire(&rdev->lock);
+		rpmsg_ept_decref(ept);
 
 		/* Check whether callback wants to hold buffer */
 		if (!(rp_hdr->reserved & RPMSG_BUF_HELD)) {
@@ -571,6 +573,7 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	struct rpmsg_endpoint *_ept;
 	struct rpmsg_ns_msg *ns_msg;
 	uint32_t dest;
+	bool ept_to_release;
 	char name[RPMSG_NAME_SIZE];
 
 	(void)priv;
@@ -589,14 +592,27 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	metal_mutex_acquire(&rdev->lock);
 	_ept = rpmsg_get_endpoint(rdev, name, RPMSG_ADDR_ANY, dest);
 
+	/*
+	 * If ept-release callback is not implemented, ns_unbind_cb() can free the ept.
+	 * Test _ept->release_cb before calling ns_unbind_cb() callbacks.
+	 */
+	ept_to_release = _ept && _ept->release_cb;
+
 	if (ns_msg->flags & RPMSG_NS_DESTROY) {
 		if (_ept)
 			_ept->dest_addr = RPMSG_ADDR_ANY;
+		if (ept_to_release)
+			rpmsg_ept_incref(_ept);
 		metal_mutex_release(&rdev->lock);
 		if (_ept && _ept->ns_unbind_cb)
 			_ept->ns_unbind_cb(_ept);
 		if (rdev->ns_unbind_cb)
 			rdev->ns_unbind_cb(rdev, name, dest);
+		if (ept_to_release) {
+			metal_mutex_acquire(&rdev->lock);
+			rpmsg_ept_decref(_ept);
+			metal_mutex_release(&rdev->lock);
+		}
 	} else {
 		if (!_ept) {
 			/*
--- a/lib/rpmsg/rpmsg_internal.h
+++ b/lib/rpmsg/rpmsg_internal.h
@@ -109,6 +109,31 @@ rpmsg_get_ept_from_addr(struct rpmsg_device *rdev, uint32_t addr)
 	return rpmsg_get_endpoint(rdev, NULL, addr, RPMSG_ADDR_ANY);
 }
 
+/**
+ * @internal
+ *
+ * @brief Increase the endpoint reference count
+ *
+ * This function is used to avoid calling ept_cb after release lock causes race condition
+ * it should be called under lock protection.
+ *
+ * @param ept	pointer to rpmsg endpoint
+ *
+ */
+void rpmsg_ept_incref(struct rpmsg_endpoint *ept);
+
+/**
+ * @internal
+ *
+ * @brief Decrease the end point reference count
+ *
+ * This function is used to avoid calling ept_cb after release lock causes race condition
+ * it should be called under lock protection.
+ *
+ * @param ept	pointer to rpmsg endpoint
+ */
+void rpmsg_ept_decref(struct rpmsg_endpoint *ept);
+
 #if defined __cplusplus
 }
 #endif


yintao707:
if rpmsg service free the ept when has got the ept from the ept
list in rpmsg_virtio_rx_callback, there is a used after free about
the ept, so add refcnt to end point and call the rpmsg service
release callback when ept callback fininshed.

arnopo:
Hello @yintao707 

Your issue is not clear for me. 
Please, could you detail your issue ? what do you mean by "there is a used after free about the ept,"

Seems to me that you can call `rpmsg_destroy_ept()` at the end of your endpoint callback. The only constraint would be that you don't use the endpoint in your callback after the destroy.

lib/include/openamp/rpmsg.h
```diff
@@ -73,6 +77,18 @@ struct rpmsg_endpoint {
 	/** Address of the default remote endpoint binded */
 	uint32_t dest_addr;
 
+	/** Reference count of the endpoint */
+	atomic_uint refcnt;
+
+	/** Callback to increase the endpoint reference count for service */
+	rpmsg_ept_inc_ref_cb inc_ref_cb;
```
xiaoxiang781216, line 84: why need callback for ref

lib/include/openamp/rpmsg.h
```diff
@@ -73,6 +77,18 @@ struct rpmsg_endpoint {
...
+
+	/** Callback to decrease the endpoint reference count for service */
+	rpmsg_ept_dec_ref_cb dec_ref_cb;
+
+	/** Callback to free the endpoint */
+	rpmsg_ept_release_cb rel_cb;
```
xiaoxiang781216, line 90:
```suggestion
	rpmsg_ept_release_cb release_cb;
```

lib/include/openamp/rpmsg.h
```diff
@@ -142,6 +158,30 @@ struct rpmsg_device {
...
+ * @brief Increase the endpoint reference count
+ *
+ * This function increases reference count of the endpoint, if rpmsg service
+ * registered inc_ref_cb of @ept, the reference count is processed at the
+ * service; otherwise, it will directly increase refcnt of @ept.
+ *
+ * @ept: pointer to rpmsg endpoint
+ *
+ */
+void rpmsg_ept_inc_ref(struct rpmsg_endpoint *ept);
```
xiaoxiang781216, line 171:
```suggestion
void rpmsg_incref(struct rpmsg_endpoint *ept);
```

lib/include/openamp/rpmsg.h
```diff
@@ -142,6 +158,30 @@ struct rpmsg_device {
...
+ * @brief Decrease the end point reference count
+ *
+ * This function decreases reference count of the endpoint, if rpmsg service
+ * registered dec_ref_cb of @ept, the reference count is processed at the
+ * service; otherwise, it will directly increase refcnt of @ept. Release ept
+ * when the reference count decreases to zero.
+ *
+ * @ept: pointer to rpmsg endpoint
+ */
+void rpmsg_ept_dec_ref(struct rpmsg_endpoint *ept);
```
xiaoxiang781216, line 183:
```suggestion
void rpmsg_decref(struct rpmsg_endpoint *ept);
```

yintao707:
hi @arnopo ,
The problems currently encountered are as follows:
static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
{
	...
	while (rp_hdr) {
		rp_hdr->reserved = idx;

		/* Get the channel node from the remote device channels list. */
		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);

		if (ept) {
			if (ept->dest_addr == RPMSG_ADDR_ANY) {
				/*
				 * First message received from the remote side,
				 * update channel destination address
				 */
				ept->dest_addr = rp_hdr->src;
			}

			RPMSG_ASSERT(status >= 0,
				     "unexpected callback status\r\n");
		}
		...
	}
}
When the ept->cb is being executed, if other threads of the service release ept->priv, then use-afer-free will occur if the callback continues to execute; 
if rpsmg_unregister_endpoint can synchronously wait for ept->cb to be executed before ending, then the above problem will not occur. however there is no restriction on this in the current code. So I added refcnt to avoid this issue.

static void rpmsg_unregister_endpoint(struct rpmsg_endpoint *ept)
{
	struct rpmsg_device *rdev = ept->rdev;

	if (ept->addr != RPMSG_ADDR_ANY)
		rpmsg_release_address(rdev->bitmap, RPMSG_ADDR_BMP_SIZE,
				      ept->addr);
	metal_list_del(&ept->node);
	ept->rdev = NULL;
	rpmsg_decref(ept);
}

arnopo:
> When the ept->cb is being executed, if other threads of the service release ept->priv, then use-afer-free will occur if the callback continues to execute; if rpsmg_unregister_endpoint can synchronously wait for ept->cb to be executed before ending, then the above problem will not occur. however there is no restriction on this in the current code. So I added refcnt to avoid this issue.

Something doesn't seem correct to me (if I understood your use case correctly).

In your endpoint callback, you call rpmsg_destroy_ept(ept) to free the endpoint and then continue to use the ept structure. This is equivalent to using a pointer after freeing it.

Could you share your code with me so that I can better understand the issue you are trying to solve with this PR?"

yintao707:
> > When the ept->cb is being executed, if other threads of the service release ept->priv, then use-afer-free will occur if the callback continues to execute; if rpsmg_unregister_endpoint can synchronously wait for ept->cb to be executed before ending, then the above problem will not occur. however there is no restriction on this in the current code. So I added refcnt to avoid this issue.
> 
> Something doesn't seem correct to me (if I understood your use case correctly).
> 
> In your endpoint callback, you call rpmsg_destroy_ept(ept) to free the endpoint and then continue to use the ept structure. This is equivalent to using a pointer after freeing it.
> 
> Could you share your code with me so that I can better understand the issue you are trying to solve with this PR?"

hi, @arnopo , I have updated the description. Could you please take a look again

CV-Bowen:
> Thank you for clarifying the race condition! Please correct me if I am wrong, but the issue I see here is that the 'user thread service' releases the endpoint by freeing the ept pointer ("release ept"). I don't think it's a good idea to manage the free of this pointer in the library, as it is allocated and freed in the application, and we cannot ensure all application use cases.
> 
> For instance, the rpmsg callback can be called under an interrupt context. In such a case, a work can be created in the callback to treat the message in normal context. In such a scenario, your PR would not prevent the ept pointer from being freed between the callback call and the work execution.
> 
> Is there any particular reason why you are not using counters or mutexes in your application to prevent race conditions?"

@arnopo In our case, the `rpmsg_virtio_rx_callback()` is called in a thread.
And even the `rpmsg_virtio_rx_callback()` is called in interrupt and creates a work to process the message in normal context, we can call `rpmsg_incref()` first before creating work and call `rpmsg_decref()` to release the endpoint after the work finish to avoid the endpoint has been released after the `ept->cb()`.

Actually, Linux did the same thing to avoid the endpoint used-after-free issue, this is the commit:
https://github.com/torvalds/linux/commit/5a081caa0414b9bbb82c17ffab9d6fe66edbb72f
The different between OpenAMP and Linux is: Linux malloc the endpoint inside the virtio rpsmg bus, but in OpenAMP, the endpoint is maintained by the user, so I think we should add a callback `ept->release_cb()` to let user do this.

arnopo:
> > Thank you for clarifying the race condition! Please correct me if I am wrong, but the issue I see here is that the 'user thread service' releases the endpoint by freeing the ept pointer ("release ept"). I don't think it's a good idea to manage the free of this pointer in the library, as it is allocated and freed in the application, and we cannot ensure all application use cases.
> > For instance, the rpmsg callback can be called under an interrupt context. In such a case, a work can be created in the callback to treat the message in normal context. In such a scenario, your PR would not prevent the ept pointer from being freed between the callback call and the work execution.
> > Is there any particular reason why you are not using counters or mutexes in your application to prevent race conditions?"
> 
> @arnopo In our case, the `rpmsg_virtio_rx_callback()` is called in a thread. And even the `rpmsg_virtio_rx_callback()` is called in interrupt and creates a work to process the message in normal context, we can call `rpmsg_incref()` first before creating work and call `rpmsg_decref()` to release the endpoint after the work finish to avoid the endpoint has been released after the `ept->cb()`.

That's my point. Half of the solution is in the library and the other part is in the application. It seems to me that you are creating a open-amp interface to solve a problem in a complex way, which could be addressed with a mutex or acounter in the application in a safer way ( for instance you can face same issue if you free the ept->priv with another thread).

> 
> Actually, Linux did the same thing to avoid the endpoint used-after-free issue, this is the commit: [torvalds/linux@5a081ca](https://github.com/torvalds/linux/commit/5a081caa0414b9bbb82c17ffab9d6fe66edbb72f) The different between OpenAMP and Linux is: Linux malloc the endpoint inside the virtio rpsmg bus, but in OpenAMP, the endpoint is maintained by the user, so I think we should add a callback `ept->release_cb()` to let user do this.

Yes, in Linux, this has been implemented because the  endpoint allocation and free is also managed by the virtio rpmsg bus.. However, this is not the case for OpenAMP, as a requirement is to support static allocation.

The ept->release_cb() only makes sense with a counter mechanism. Therefore, I don't see a strong reason to complexify the rpmsg API, to partially answer to some memory allocations/free that should be managed by application .

@edmooring, @tnmysh : any opinion on this?

xiaoxiang781216:
> The ept->release_cb() only makes sense with a counter mechanism. Therefore, I don't see a strong reason to complexify the rpmsg API, to partially answer to some memory allocations/free that should be managed by application .
> 

@CV-Bowen has tried to add refcount inside application, but it can't fix this problem without any race condition, for example:
```
void my_cb(...)
{
   atomic_fetch_add(&g_priv->refcnt, 1); // 2*
   ....
  if (atomic_fetch_sub(&g_priv->refcnt, 1) == 1) {
    rpmsg_destroy_ept(&g_priv->ept);
    free(g_priv);
  }
}

void my_init(void)
{
  g_priv = malloc(sizeof(*g_priv));
  atomic_init(&g_priv->refcnt, 1);
  g_priv->ept.cb = my_cb;
  rpmsg_create_ept(&g_priv->ept);
}

void my_deinit(void)
{
  if (atomic_fetch_sub(&g_priv->refcnt, 1) == 1) { // 1*
    rpmsg_destroy_ept(&g_priv->ept);
    free(g_priv);
  }
}
```
Let's assume this sequence:

1. rpmsg_virtio_rx_callback release lock here:
    https://github.com/OpenAMP/open-amp/pull/508/files#diff-569d5f59c7dc7ac2fee7252d74df8af1ba6e366a30fae7e1b97de627bc0ce472L563
2. os decide to suspend rpmsg_virtio_rx_callback and switch to my_deinit
3.  my_deinit finish execution and release g_priv(1*)
4. rpmsg_virtio_rx_callback resume and then happen use-after-free

You can see the reference count inside my_cb can just reduce the race condition(safe after 2*), the gap between item 1 and 2* can just be handled by OpenAMP self.

xiaoxiang781216:
> > 
> > @CV-Bowen has tried to add refcount inside application, but it can't fix this problem without any race condition, for example:
> > ```
> > void my_cb(...)
> > {
> > ```
> 
> Here I suppose that you can not test g_priv ponter, else would be simple to exit with a mutex protection
> 

Yes, if it saves in a global variable and protected by a mutex in this simple demo, but the real case is that the endpoints are created and destroyed with dynamical number instances. Even this simple demo applies the change you suggest, the race condition still exists since OpenAMP access g_priv->ept's fields too before entering my_cb.

yintao707:
> * Can you send a version that fix the protection hole,  between the metal_mutex_release and the callbacks calls?
> * I would prefer something else that atomic_* functions in open-amp lib. We regularly face problems with compilers that not implements the atomic library.

hi @arnopo , Updated, please help review again.thanks

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -636,9 +639,13 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	if (ns_msg->flags & RPMSG_NS_DESTROY) {
 		if (_ept)
 			_ept->dest_addr = RPMSG_ADDR_ANY;
+		rpmsg_incref(_ept);
 		metal_mutex_release(&rdev->lock);
 		if (_ept && _ept->ns_unbind_cb)
 			_ept->ns_unbind_cb(_ept);
+		metal_mutex_acquire(&rdev->lock);
+		rpmsg_decref(_ept);
+		metal_mutex_release(&rdev->lock);
```
arnopo, line 648: Seem better to put block after `rdev->ns_unbind_cb`

lib/include/openamp/rpmsg.h
```diff
@@ -73,6 +74,12 @@ struct rpmsg_endpoint {
 	/** Address of the default remote endpoint binded */
 	uint32_t dest_addr;
 
+	/** Reference count of the endpoint */
+	uint32_t refcnt;
+
+	/** Callback to free the endpoint */
```
arnopo, line 80: Callback to inform the user that the endpoint allocation can be safely removed (optional).

lib/include/openamp/rpmsg.h
```diff
@@ -142,6 +149,21 @@ struct rpmsg_device {
 	bool support_ns;
 };
 
+/**
+ * @brief Increase the endpoint reference count
+ *
+ * @ept: pointer to rpmsg endpoint
+ *
+ */
+void rpmsg_incref(struct rpmsg_endpoint *ept);
```
arnopo, line 158: rpmsg_ept_incref and rpmsg_ept_decref

lib/include/openamp/rpmsg.h
```diff
@@ -142,6 +149,21 @@ struct rpmsg_device {
 	bool support_ns;
 };
 
+/**
+ * @brief Increase the endpoint reference count
```
arnopo, line 153: Add  ` @internal` label as it should not be used by user.

Please also add a sentence to explain that the function should be called under `rdev->lock` protection
the comment is also valid for `rpmsg_decref`

lib/rpmsg/rpmsg.c
```diff
@@ -97,6 +97,23 @@ static int rpmsg_set_address(unsigned long *bitmap, int size, int addr)
...
+	if (ept) {
+		ept->refcnt++;
+	}
+}
+
+void rpmsg_decref(struct rpmsg_endpoint *ept)
+{
+	if (ept) {
+		ept->refcnt--;
+		if (ept->refcnt == 0 && ept->release_cb) {
```
arnopo, line 111: nitpicking s/`ept->refcnt == 0`/`!ept->refcnt`/

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -568,6 +569,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 				 */
 				ept->dest_addr = rp_hdr->src;
 			}
+
```
arnopo, line 572: remove extra line

arnopo (changes requested):
Minor comments., else LGTM
Please, also update the commit message to better explain the race condition issue, and fix typo errors.

yintao707:
> Minor comments., else LGTM Please, also update the commit message to better explain the race condition issue, and fix typo errors.

hi， @arnopo ，Thank you very much for your patient review， I have updated the PR based on your modification suggestions. Could you please help me review again， thanks

lib/rpmsg/rpmsg.c
```diff
@@ -256,6 +274,7 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
 			     rpmsg_ept_cb cb,
 			     rpmsg_ns_unbind_cb ns_unbind_cb)
 {
+	rpmsg_ept_incref(ept);
```
xiaoxiang781216, line 277: change to directly assignment

lib/rpmsg/rpmsg.c
```diff
@@ -97,6 +97,23 @@ static int rpmsg_set_address(unsigned long *bitmap, int size, int addr)
 	}
 }
 
+void rpmsg_ept_incref(struct rpmsg_endpoint *ept)
+{
+	if (ept) {
```
xiaoxiang781216, line 102: check release_cb before increasing refcnt

lib/rpmsg/rpmsg.c
```diff
@@ -97,6 +97,23 @@ static int rpmsg_set_address(unsigned long *bitmap, int size, int addr)
...
+		ept->refcnt++;
+	}
+}
+
+void rpmsg_ept_decref(struct rpmsg_endpoint *ept)
+{
+	if (ept) {
```
xiaoxiang781216, line 109: check ept->release_cb before decreasing refcnt

lib/rpmsg/rpmsg.c
```diff
@@ -246,6 +263,7 @@ static void rpmsg_unregister_endpoint(struct rpmsg_endpoint *ept)
 				      ept->addr);
 	metal_list_del(&ept->node);
 	ept->rdev = NULL;
+	rpmsg_ept_decref(ept);
```
xiaoxiang781216, line 266: move before line 265

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -616,6 +618,7 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	struct rpmsg_ns_msg *ns_msg;
 	uint32_t dest;
 	char name[RPMSG_NAME_SIZE];
+	bool ept_release;
```
xiaoxiang781216, line 621:
```suggestion
	bool ept_decref;
```

lib/rpmsg/rpmsg.c
yintao707: move before line 265 probably occur use-after-free

xiaoxiang781216: ok

lib/rpmsg/rpmsg.c
```diff
@@ -257,6 +275,7 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
 			     rpmsg_ns_unbind_cb ns_unbind_cb)
 {
 	strncpy(ept->name, name ? name : "", sizeof(ept->name));
+	ept->refcnt++;
```
xiaoxiang781216, line 278: ept->refcnt = 1

lib/include/openamp/rpmsg.h
```diff
@@ -142,6 +149,31 @@ struct rpmsg_device {
...
+ *
+ * @brief Increase the endpoint reference count
+ *
+ * This function is used to avoid calling ept_cb after release lock causes race condition
+ * it should be called under lock protection.
+ *
+ * @param ept	pointer to rpmsg endpoint
+ *
+ */
+void rpmsg_ept_incref(struct rpmsg_endpoint *ept);
```
xiaoxiang781216, line 163: let's move the function declaration to lib/rpmsg/rpmsg_internal.h

xiaoxiang781216:
LGTM

lib/rpmsg/rpmsg.c
arnopo, line 266: Here we can have an issue if `ept->rdev` is used in `ept->release_cb`
move `ept->rdev = NULL` in rpmsg_ept_decref() should avoid issue

```
void rpmsg_ept_decref(struct rpmsg_endpoint *ept)
{
	if (ept && ept->release_cb) {
		ept->refcnt--;
		if (!ept->refcnt) 
			ept->release_cb(ept);
		else 
			ept->rdev = NULL;
	}
}
```

lib/rpmsg/rpmsg.c
```diff
@@ -97,6 +97,23 @@ static int rpmsg_set_address(unsigned long *bitmap, int size, int addr)
...
+}
+
+void rpmsg_ept_decref(struct rpmsg_endpoint *ept)
+{
+	if (ept && ept->release_cb) {
+		ept->refcnt--;
+		if (!ept->refcnt) {
+			ept->release_cb(ept);
+		}
+	}
```
arnopo, line 114: nitpicking: useless brackets for the if condition

arnopo (changes requested):

lib/rpmsg/rpmsg.c
yintao707: thanks, but I think there may be issues with such changes. firstly, not all endpoints require registration for release_cb; Secondly, the first call to rpmsg_ept_cb should be after rpmsg_xxx_ept_cb， but at that time, refcnt will not decrease to 0, setting ept ->rdev=NULL directly can also cause problems.

xiaoxiang781216: after release_cb, it's invalid to touch any fields inside ept.

arnopo: oops sorry my proposal contains a bug....
The idea is to either call the` release_cb()` or to set `ept->rdev` to `NULL`

```
void rpmsg_ept_incref(struct rpmsg_endpoint *ept)
{
	if (ept) {
		ept->refcnt++;
	}
}

void rpmsg_ept_decref(struct rpmsg_endpoint *ept)
{
	if (ept) {
		ept->refcnt--;
		if (!ept->refcnt) 
			if (ept->release_cb)
				ept->release_cb(ept);
			else 
				ept->rdev = NULL;
	}
}
```

lib/include/openamp/rpmsg.h
```diff
@@ -50,6 +50,8 @@ struct rpmsg_device;
 /* Returns positive value on success or negative error value on failure */
 typedef int (*rpmsg_ept_cb)(struct rpmsg_endpoint *ept, void *data,
 			    size_t len, uint32_t src, void *priv);
+typedef void (*rpmsg_ept_incref_cb)(void *priv);
```
xiaoxiang781216, line 53:
```suggestion
typedef void (*rpmsg_ept_incref_cb)(struct rpmsg_endpoint *ept);
```

lib/include/openamp/rpmsg.h
```diff
@@ -50,6 +50,8 @@ struct rpmsg_device;
...
+typedef void (*rpmsg_ept_decref_cb)(void *priv);
```
xiaoxiang781216, line 54:
```suggestion
typedef void (*rpmsg_ept_decref_cb)(struct rpmsg_endpoint *ept);
```

yintao707:
hi, @arnopo , I updated this PR because I found that if the reference count is only handled in the openamp layer, there may be the following issues.
Assuming this is my rpmsg_service thread, under normal circumstances, I would increment the reference count in my_init(). If no ept_cb is currently executing, I would release ept->priv in my_deinit().
```
rpmsg_service thread

void my_init()
{
    ...
    rpmsg_create_ept(ept);
    //ept->ref = 1;
    ...
}

void my_deinit()
{
    ...
    //Assuming ept_cb is not currently executing
    rpmsg_destroy_ept(ept);
        call release_cb(ept);
             ept->refcnt-1=0;
             free(ept->priv);
}
```
But if there is another thread that actively calls rpmsg_destroy_ept through my_stop_remote, and the previous reference count is 1, after rpmsg_destroy_ept is executed, release_cb will release ept->priv. If the rpmsg_service thread continues to access ept->priv, a problem may occur.
```
other thread

void my_stop_remote()
{
    /* Traversing all registered services.*/
    metal_list_for_each(g_rp_service, node)
    {
        rpmsg_xxx_device_destroy();
            call rpmsg_destroy_ept(ept);
                 call release_cb free ept->priv;
    }
}
```
Therefore, I have changed the original release_cb to incref_cb and decref_cb, allowing the service to control the usage of reference counts. This can mitigate the aforementioned problem.

Please @xiaoxiang781216 @arnopo  help review this modification，thanks

xiaoxiang781216 (approved):
LGTM.

arnopo:
> hi, @arnopo , I updated this PR because I found that if the reference count is only handled in the openamp layer, there may be the following issues. Assuming this is my rpmsg_service thread, under normal circumstances, I would increment the reference count in my_init(). If no ept_cb is currently executing, I would release ept->priv in my_deinit().
> 
> ```
> rpmsg_service thread
> 
> void my_init()
> {
>     ...
>     rpmsg_create_ept(ept);
>     //ept->ref = 1;
>     ...
> }
> 
> void my_deinit()
> {
>     ...
>     //Assuming ept_cb is not currently executing
>     rpmsg_destroy_ept(ept);
>         call release_cb(ept);
>              ept->refcnt-1=0;
>              free(ept->priv);
> }
> ```
> 
> But if there is another thread that actively calls rpmsg_destroy_ept through my_stop_remote, and the previous reference count is 1, after rpmsg_destroy_ept is executed, release_cb will release ept->priv. If the rpmsg_service thread continues to access ept->priv, a problem may occur.

No cristal clear  to me: "If the rpmsg_service thread continues to access ept->priv, a problem may occur."
When the release callback is called this means that no one is using the ept, right? 
If you set   `ept->priv` to `NULL` in release callback does it prevent the issue?
> 
> ```
> other thread
> 
> void my_stop_remote()
> {
>     /* Traversing all registered services.*/
>     metal_list_for_each(g_rp_service, node)
>     {
>         rpmsg_xxx_device_destroy();
>             call rpmsg_destroy_ept(ept);
>                  call release_cb free ept->priv;
>     }
> }
> ```
> 
> Therefore, I have changed the original release_cb to incref_cb and decref_cb, allowing the service to control the usage of reference counts. This can mitigate the aforementioned problem.
> 
> Please @xiaoxiang781216 @arnopo help review this modification，thanks

Your last version looks more like a hack to me. If the release_cb does not solve the issue, we probably have to consider another approach

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -636,11 +639,20 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 	if (ns_msg->flags & RPMSG_NS_DESTROY) {
 		if (_ept)
 			_ept->dest_addr = RPMSG_ADDR_ANY;
+		if (_ept && _ept->release_cb) {
```
arnopo, line 642: already tested in  `rpmsg_ept_incref`

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -636,11 +639,20 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
...
+			rpmsg_ept_incref(_ept);
+			ept_decref = true;
```
arnopo, line 644: useless local variable

arnopo (changes requested):
2 remaining comments not yet addressed + 2 new one

lib/rpmsg/rpmsg_virtio.c
yintao707: because some serives will free ept in _ept->ns_unbind_cb， if not added ept_decref 
```
if (_ept && _ept->ns_unbind_cb)
   _ept->ns_unbind_cb(_ept);
rpmsg_ept_decref(_ept);
```
rpmsg_ept_decref will use-afrer-free

lib/rpmsg/rpmsg_virtio.c
yintao707: 642 is to judge local variable ept_decref

yintao707:
> 2 remaining comments not yet addressed + 2 new one

Done, thanks！

arnopo (approved):
LGTM, thanks!

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -515,6 +515,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		/* Get the channel node from the remote device channels list. */
 		metal_mutex_acquire(&rdev->lock);
 		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
+		rpmsg_ept_incref(ept);
```
tnmysh, line 518: @yintao707, @arnopo does that make sense to move rpmsg_ept_incref() API within rpmsg_get_endpoint API ? If endpoint is retrieved successfully then we increase refcount. It is possible that get_endpoint API is called in future, so we increase refcount notifying endpoint is being used.

tnmysh: @arnopo Looks like refcnt is tracking more if callback is in progress or not rather than how many times endpoint is being used my multiple threads using rpmsg_get_endpoint correct ?

If this is the case, can we update documentation accordingly ?

arnopo: > @yintao707, @arnopo does that make sense to move rpmsg_ept_incref() API within rpmsg_get_endpoint API ? If endpoint is retrieved successfully then we increase refcount. It is possible that get_endpoint API is called in future, so we increase refcount notifying endpoint is being used.

I would prefer not to hide it in `rpmsg_get_endpoint` and address this only if we need to export the rpmsg_get_endpoint API in the future.



> @arnopo Looks like refcnt is tracking more if callback is in progress or not rather than how many times endpoint is being used my multiple threads using rpmsg_get_endpoint correct ?
> 
> If this is the case, can we update documentation accordingly ?

Is the documentation header for rpmsg_ept_incref not explicit enough for you? 
Could you provide more details on which part of the documentation you would  like to see updated ?

tnmysh: documentation for ept_incref looks good. But same documentation for refcnt variable should be updated: 

	/** Reference count of the endpoint */
	uint32_t refcnt;

Above comment gives impression that refcnt variable is used for endpoint object reference counts. But, refcnt variable is used to track if callback execution is in progress or not. So, above variable documentation should be updated accordingly.

GUIDINGLI:
LGTM

lib/rpmsg/rpmsg_virtio.c
arnopo: > documentation for ept_incref looks good. But same documentation for refcnt variable should be updated:
> 
> ```
> /** Reference count of the endpoint */
> uint32_t refcnt;
> ```
> 
> Above comment gives impression that refcnt variable is used for endpoint object reference counts. But, refcnt variable is used to track if callback execution is in progress or not. So, above variable documentation should be updated accordingly.

@yintao707 , please could you address @tnmysh comment that we can merge it?

yintao707: @arnopo @tnmysh , Thank you for your suggestion, I modified the comments about refcnt. Can you help me review whether this modification is appropriate

tnmysh (approved):

CV-Bowen:
> @yintao707:
> Can you send a version that fix the protection hole, between the metal_mutex_release and the callbacks calls?
> I would prefer something else that atomic_* functions in open-amp lib. We regularly face problems with compilers that not implements the atomic library. 

@arnopo Hi, revisit this PR, I found that you mentioned that do not use the atomic operations in open-amp lib. I'm a litte confused, because libmetal has implemented the atomic operations and all the platforms should port to libmetal and then the atomic operations should be available in all platforms. Could you discribe this more clear? Thanks.

arnopo:
Hi @CV-Bowen 

> > @yintao707:
> > Can you send a version that fix the protection hole, between the metal_mutex_release and the callbacks calls?
> > I would prefer something else that atomic_* functions in open-amp lib. We regularly face problems with compilers that not implements the atomic library.
> 
> @arnopo Hi, revisit this PR, I found that you mentioned that do not use the atomic operations in open-amp lib. I'm a litte confused, because libmetal has implemented the atomic operations and all the platforms should port to libmetal and then the atomic operations should be available in all platforms. Could you discribe this more clear? Thanks.

I was referring to the OpenAMP library only. Currently, the only atomic function used is `atomic_thread_fence`.
Since this is not always well supported by some compilers, atomic operations should be used only in the libmetal library and not called directly in the OpenAMP library.

CV-Bowen:
@arnopo Thanks, actually, what I want to do is try to make the `ept->cb` can be called in interrupt to improve the communication latency (reduce the context switch time), so I do:
1. change the nuttx implemented mutex to spinlock in Libmetal;
2. move the `ept->release_cb` out side the `rdev->lock` region because the `mutex_lock` or some other functions that will lead context switch will be called in `ept->release_cb`, and this will cause the dead lock because context switch in a spinlock lock region. So I wonder if we can move the `ept->release_cb` out of the `rdev->lock` region, and seems change the reference count to atomic is a good method.

But after listening to your explanation, I decide to still use the `rdev->lock` to protect the ept reference count and will push a PR later.

PR #509, opened 2023-10-09
virtio: make unused function virtio_describe() deprecated

arnopo:
The virtio_describe() function does not seem usable. Therefore, it should be marked as deprecated to initiate the removal process.


arnopo:
@edmooring @tnmysh, please add a look, I would like to merge this ne for the coming release

tnmysh (approved):
LGTM. 

I don't have too deep knowledge in virtio space to understand how function should be used however, if function isn't used anywhere for long-time we can deprecate it. And if needed bring-it back later.

PR #514, opened 2023-10-12
rpmsg: set ept address to incease num when alloc from the bitmap

GUIDINGLI:
rpmsg: set ept address to incease num when alloc from the bitmap

    
    CPU0                                CPU1
    create_ept1:addr1                   create_ept1
    OK                       <======    msg1
    OK                       <======    msg2
    OK                       <======    msg3
                               msg4
                             <======    msg4 on the virtioqueue
    close_ept1                          close_ept1
    
    create_ept2:addr1                   create_ept1
    (same addr with ept1)
                               msg4
    ept2 recv ept1 msg ERROR <======

The msg4 which belong to ep1, error received by ept2.

For the issue, I give a resolve method,
for the rpmsg_get_address(), always return a new increased num.

This should compile with https://github.com/OpenAMP/libmetal/pull/263

### DIFF
--- a/lib/rpmsg/rpmsg.c
+++ b/lib/rpmsg/rpmsg.c
@@ -24,12 +24,12 @@
  *
  * @return A unique address
  */
-static uint32_t rpmsg_get_address(unsigned long *bitmap, int size)
+static uint32_t rpmsg_get_address(unsigned long *bitmap, unsigned int start, int size)
 {
 	unsigned int addr = RPMSG_ADDR_ANY;
 	unsigned int nextbit;
 
-	nextbit = metal_bitmap_next_clear_bit(bitmap, 0, size);
+	nextbit = metal_bitmap_next_clear_bit(bitmap, start, size);
 	if (nextbit < (uint32_t)size) {
 		addr = RPMSG_RESERVED_ADDRESSES + nextbit;
 		metal_bitmap_set_bit(bitmap, nextbit);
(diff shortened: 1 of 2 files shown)


lib/rpmsg/rpmsg.c
```diff
@@ -24,12 +24,12 @@
...
  * @return A unique address
  */
-static uint32_t rpmsg_get_address(unsigned long *bitmap, int size)
+static uint32_t rpmsg_get_address(unsigned long *bitmap, unsigned int start, int size)
 {
 	unsigned int addr = RPMSG_ADDR_ANY;
 	unsigned int nextbit;
 
-	nextbit = metal_bitmap_next_clear_bit(bitmap, 0, size);
+	nextbit = metal_bitmap_loopnext_clear_bit(bitmap, start % size, size);
```
arnopo, line 32: regarding libmetal PR
```c
static inline unsigned int
metal_bitmap_loopnext_clear_bit(unsigned long *bitmap, unsigned int start,
				unsigned int max)
{
	unsigned int bit;

	for (bit = start;
	     bit < max + start && !metal_bitmap_is_bit_clear(bitmap, bit % max);
	     bit++)
		;
	return bit % max;
}
```
Not clear to me why you have to use a new metal function
`metal_bitmap_next_clear_bit(bitmap, start % size, size) `does not work?

arnopo (changes requested):
The issue is well described in the commit message.
Could you also explain in the commit message the solution  implemented?

lib/rpmsg/rpmsg.c
GUIDINGLI: metal_bitmap_next_clear_bit:
find bit from: start ... max

metal_bitmap_loopnext_clear_bit:
find bit from:
start ... max
**0 ... start**


metal_bitmap_loopnext_clear_bit can search back from the 0, when meet the max.

tnmysh: May be we can improve current function as it is?

Such as if start is not 0, then loopback after max.

lib/rpmsg/rpmsg.c
```diff
@@ -24,12 +24,12 @@
...
  * @return A unique address
  */
-static uint32_t rpmsg_get_address(unsigned long *bitmap, int size)
+static uint32_t rpmsg_get_address(unsigned long *bitmap, unsigned int start, int size)
 {
 	unsigned int addr = RPMSG_ADDR_ANY;
 	unsigned int nextbit;
 
-	nextbit = metal_bitmap_next_clear_bit(bitmap, 0, size);
+	nextbit = metal_bitmap_next_clear_bit(bitmap, start % size, size);
```
xiaoxiang781216, line 32: why need %?

xiaoxiang781216 (approved):
LGTM.

arnopo (approved):

tnmysh (approved):

PR #516, opened 2023-10-13
Docs: style doxygen when built as part of OpenAMP docs

wmamills:
Detect when we are a submodule of OpenAMP-docs.
If so use the top level's Doxyfile overrides on top of our Doxyfile. If not, no change.


wmamills:
This is the same support we already have in libmetal

arnopo (approved):
No regression found when I enabe the  `WITH_DOC` option

tammyleino (approved):

PR #522, opened 2023-10-17
remoteproc: clear bitmap when stop/shutdown

CV-Bowen:
remoteproc: clear bitmap in remoteproc_stop()
    
Fix parse res_table failed when repeat start the remoteproc.
If we don't clear the bitmap it will failed in
      handle_vdev_rsc() -> remoteproc_allocate_id()

### DIFF
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -256,6 +256,7 @@ int remoteproc_stop(struct remoteproc *rproc)
 			if (rproc->ops->stop)
 				ret = rproc->ops->stop(rproc);
 			rproc->state = RPROC_STOPPED;
+			rproc->bitmap = 0;
 		} else {
 			ret = 0;
 		}


lib/remoteproc/remoteproc.c
```diff
@@ -283,6 +284,8 @@ int remoteproc_shutdown(struct remoteproc *rproc)
 					rproc->state = RPROC_OFFLINE;
 				}
 			}
+
+			rproc->bitmap = 0;
```
arnopo, line 288: Any reason to do it again here?  It should already be set to 0 in `remoteproc_stop()`, right?

tnmysh: Agree with @arnopo.

edmooring: Is there a reasonable use case where remoteproc_shutdown() is called without then calling remoteproc_stop()?

CV-Bowen: @arnopo @tnmysh @edmooring Hi, yes, in our case, we only call `remoteproc_shutdown()` because `remoteproc_shutdown()` will call the stop too, and if there is no use case where only call `remoteproc_stop()`, I can delete the bitmap clear in `remoteproc_stop()`.

arnopo: seems that the call stack to update the `rproc->bitmap` is 
```
remoteproc_load
  remoteproc_get_rsc_table
    handle_rsc_table
      handle_vdev_rsc
        remoteproc_allocate_id
```
so the bitmap is set on load and should be reset on stop.

CV-Bowen: Got it, Done.

xiaoxiang781216 (approved):

edmooring (approved):
Good to go now.

arnopo (approved):

PR #523, opened 2023-10-17
rpmsg_virtio: fix rpmsg_service memory leak when stop remote

CV-Bowen:
Rpmsg service may malloc memory in the rpmsg bind callback(): rdev->ns_bind_cb() When stop remote, the remote may not send RPMSG_NS_DESTROY message back (baceause the remote may has crashed), so call the unbind callback in rpmsg_deinit_vdev() to aviod potential memroy leak in rpmsg services.

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -862,6 +862,8 @@ void rpmsg_deinit_vdev(struct rpmsg_virtio_device *rvdev)
 			node = rdev->endpoints.next;
 			ept = metal_container_of(node, struct rpmsg_endpoint, node);
 			rpmsg_destroy_ept(ept);
+			if (ept->ns_unbind_cb)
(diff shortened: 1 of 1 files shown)


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -862,6 +862,8 @@ void rpmsg_deinit_vdev(struct rpmsg_virtio_device *rvdev)
 			node = rdev->endpoints.next;
 			ept = metal_container_of(node, struct rpmsg_endpoint, node);
 			rpmsg_destroy_ept(ept);
+			if (ept->ns_unbind_cb)
```
xiaoxiang781216, line 865: should we move before line 864

CV-Bowen:
This patch still has some problems to fix. @yintao707 will create a new PR to fix this issue and I will close this PR.

arnopo:
ok, so addressed in https://github.com/OpenAMP/open-amp/pull/525

PR #525, opened 2023-10-17
openamp/rpmsg: fix rpmsg_service memory leak when stop remote

yintao707:
When remote cannot send destroy messages, master cannot call ns_unbind_cb, this will cause the master to be unable to release resources

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -862,6 +862,8 @@ void rpmsg_deinit_vdev(struct rpmsg_virtio_device *rvdev)
 			node = rdev->endpoints.next;
 			ept = metal_container_of(node, struct rpmsg_endpoint, node);
 			rpmsg_destroy_ept(ept);
+			if (ept->ns_unbind_cb)
+				ept->ns_unbind_cb(ept);
 		}
 
 		rvdev->rvq = 0;


arnopo:
Seems similar to  https://github.com/OpenAMP/open-amp/pull/508, could you have a look?

tammyleino (approved):

edmooring (approved):
Looks good to go.

yintao707:
> Seems similar to #508, could you have a look?

Thank you for your review,  this has little interaction with #508

arnopo:
> > Seems similar to #508, could you have a look?
> 
> Thank you for your review, this has little interaction with #508

Please could you explain what do you mean by "little interaction" ? 
The `ns_unbind_cb` is used to inform that the other side destroys it endpoint. Using it for the deinit can introduce side effect in the existing code. The ept release introduced in #508 seems more appropriate. Is it an issue to use it?

yintao707:
> > > Seems similar to #508, could you have a look?
> > 
> > 
> > Thank you for your review, this has little interaction with #508
> 
> Please could you explain what do you mean by "little interaction" ? The `ns_unbind_cb` is used to inform that the other side destroys it endpoint. Using it for the deinit can introduce side effect in the existing code. The ept release introduced in #508 seems more appropriate. Is it an issue to use it?

yes, you're right, sorry for my inaccurate expression. But the main reason for this modification is that master cannt received NS_DESTORY message, when remote is powered down or exits abnormally. at that time, master needs to actively call ns_unbind_cb to release resources

xiaoxiang781216:
@arnopo does this change make sense now?

arnopo:
> @arnopo does this change make sense now?

As we go to integrate the `ept->release` ops, look to me that is should also address the need described here. But perhaps I miss something?

arnopo:
could we  close this one as we merge #508?

arnopo:
I close it as no news Please reopen it if needed

PR #524, opened 2023-10-17
openamp: change rx buffer hold flag to count

CV-Bowen:
This commit fix one issue, before this commit, if `rpmsg_hold_rx_buffer()` is called in the endpoint callback to hold current received buffer and call` rpmsg_release_rx_buffer()` to release this buffer immediately.
This buffer will be returned to the virtqueue in `rpmsg_virtio_release_rx_buffer()` and returned to virtqueue again in `rpmsg_virtio_return_buffer()` after the `ept->cb()`.
So same buffer returned to the virtqueue twice, error happened.
Follow shows this process:

```
rpmsg_virtio_rx_callback()
  - get rp_hdr
  - ept->cb(data = RPMSG_LOCATE_DATA(rp_hdr))
    - rpsmg_hold_rx_buffer(data)
    - rpmsg_release_rx_buffer(data) here has returned buffer to virtqueue
  - rpmsg_virtio_return_buffer() return same buffer to virtqueue again
```

The root cause of this issue is that the RPMSG_BUF_HELD has been cleared in `rpmsg_release_rx_buffer()` so `rpmsg_virtio_rx_callback()` will treat this buffer is not HELD and call `rpmsg_virtio_return_buffer()` to return this buffer again.

This commit change the HELD flag in rp_hdr to count to fix this issue and also support hold/release rx buffer recursively.

PR #526, opened 2023-10-18
rpmsg_virtio: add RPMSG_ASSERT to check the virtqueue add error

CV-Bowen:
Add RPMSG_ASSERT() to rpmsg_virtio_return_buffer() to check the possible virtqueue buffer add error.

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -91,6 +91,7 @@ static void rpmsg_virtio_return_buffer(struct rpmsg_virtio_device *rvdev,
 				       uint16_t idx)
 {
 	unsigned int role = rpmsg_virtio_get_role(rvdev);
+	int ret;
 
 	BUFFER_INVALIDATE(buffer, len);
 
@@ -102,14 +103,16 @@ static void rpmsg_virtio_return_buffer(struct rpmsg_virtio_device *rvdev,
 		/* Initialize buffer node */
 		vqbuf.buf = buffer;
 		vqbuf.len = len;
-		virtqueue_add_buffer(rvdev->rvq, &vqbuf, 0, 1, buffer);
+		ret = virtqueue_add_buffer(rvdev->rvq, &vqbuf, 0, 1, buffer);
+		RPMSG_ASSERT(ret == VQUEUE_SUCCESS, "add buffer failed\r\n");
 	}
 #endif /*VIRTIO_DEVICE_ONLY*/
 
 #ifndef VIRTIO_DRIVER_ONLY
 	if (role == RPMSG_REMOTE) {
 		(void)buffer;
-		virtqueue_add_consumed_buffer(rvdev->rvq, idx, len);
+		ret = virtqueue_add_consumed_buffer(rvdev->rvq, idx, len);
+		RPMSG_ASSERT(ret == VQUEUE_SUCCESS, "add consumed buffer failed\r\n");
 	}
 #endif /*VIRTIO_DRIVER_ONLY*/
 }


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -102,14 +103,17 @@ static void rpmsg_virtio_return_buffer(struct rpmsg_virtio_device *rvdev,
...
 	}
 #endif /*VIRTIO_DEVICE_ONLY*/
 
 #ifndef VIRTIO_DRIVER_ONLY
 	if (role == RPMSG_REMOTE) {
 		(void)buffer;
-		virtqueue_add_consumed_buffer(rvdev->rvq, idx, len);
+		ret = virtqueue_add_consumed_buffer(rvdev->rvq, idx, len);
+		RPMSG_ASSERT(ret == VQUEUE_SUCCESS,
+			     "add consumed buffer failed\r\n");
```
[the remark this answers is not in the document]
CV-Bowen: Done

arnopo:
"When enable the VQUEUE_DEBUG and RPMSG_DEBUG, RPMSG_ASSERT can
check the buffer add error."
I can not see how VQUEUE_DEBUG impact RPMSG_ASSERT,trace. I missed something?

CV-Bowen:
@arnopo I mean we can check the `virtqueue_add_buffer()` return error with `RPMSG_ASSERT` when enable `VQUEUE_DEBUG`.
Because `virtqueue_add_buffer()` can return error only when `VQUEUE_DEBUG` is enabled.

arnopo:
> virtqueue_add_buffer

Got it, thanks!

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -102,14 +103,16 @@ static void rpmsg_virtio_return_buffer(struct rpmsg_virtio_device *rvdev,
...
+		RPMSG_ASSERT(ret == VQUEUE_SUCCESS, "add buffer failed\r\n");
 	}
 #endif /*VIRTIO_DEVICE_ONLY*/
 
 #ifndef VIRTIO_DRIVER_ONLY
 	if (role == RPMSG_REMOTE) {
 		(void)buffer;
-		virtqueue_add_consumed_buffer(rvdev->rvq, idx, len);
+		ret = virtqueue_add_consumed_buffer(rvdev->rvq, idx, len);
+		RPMSG_ASSERT(ret == VQUEUE_SUCCESS, "add consumed buffer failed\r\n");
```
tnmysh, line 115: Can we print value of ret right before RPMSG_ASSERT using metal_err() macro? This way we know error code from add_buffer and add_consumed_buffer.

CV-Bowen: Nice catch, how about adding a new marco `RPMSG_ASSERT_WITH_RET()` like this:
```c
#define RPMSG_ASSERT_WITH_RET(_exp, _msg, _ret) do { \
		if (!(_exp)) { \
			metal_log(METAL_LOG_EMERGENCY, \
				  "FATAL: %s, ret=%d - "_msg, __func__, ret); \
			metal_assert(_exp); \
		} \
	} while (0)
```
Because I noticed that other places use `RPMSG_ASSERT` also do not print the return value,  we can replace these `RPMSG_ASSERT` to `RPMSG_ASSERT_WITH_RET`.

arnopo: Do we really need to add extra code on something that should not happen and seems result from a memory corruption?
In my opinion the `RPMSG_ASSERT` should be sufficient

CV-Bowen: @tnmysh @arnopo Both are good for me. 
In addition, even no memory corruption, as I mentioned in https://github.com/OpenAMP/open-amp/pull/524, return a same buffer to the virtqueue twice will lead this error too, so I want to add this assert to check other possible issues.

tnmysh: @arnopo Ok, in that case I am okay with only RPMSG_ASSERT. Yes we don't need extra #define.

xiaoxiang781216: Is it this patch ready for merge? @arnopo .

arnopo (approved):

arnopo:
> > @arnopo I mean we can check the `virtqueue_add_buffer()` return error with `RPMSG_ASSERT` when enable `VQUEUE_DEBUG`. Because `virtqueue_add_buffer()` can return error only when `VQUEUE_DEBUG` is enabled.
> 
> Got it, thanks! Please remove the reference `VQUEUE_DEBUG` in the commit message , at RPMsg level we don't aware of virqueue configuration. we still test the the result ( even if always `VQUEUE_SUCCESS`)

@CV-Bowen, please update the commit message, then ready  to merge

CV-Bowen:
@arnopo Thanks, Done.

edmooring (approved):
Looks good to go.

arnopo (approved):

PR #527, opened 2023-10-18
virtio: Fix delete_virtqueues function

arnopo:
The function should return void. the reason is that vdev->func->delete_virtqueues() is optional and return void.

To be generic we should call virtio_delete_virtqueues in all cases. Then depending on the dispatcher we do nothing or call delete_virtqueues()

Fixes: 7f90610 ("virtio: add create_virtqueues and delete_virtqueues in virtio_dispatch")

### DIFF
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -298,18 +298,13 @@ int virtio_create_virtqueues(struct virtio_device *vdev, unsigned int flags,
  *
  * @param vdev	Pointer to virtio device structure.
  *
- * @return 0 on success, otherwise error code.
  */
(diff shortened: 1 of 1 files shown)


arnopo:
git lint report an false positive error :
`UC4 Line exceeds max length (89>80): "Fixes: 7f90610 ("virtio: add create_virtqueues and delete_virtqueues in virtio_dispatch")"`
To ignore as this line is standard to declare fix

CV-Bowen:
@arnopo LGTM

tnmysh (approved):
LGTM

edmooring (approved):
Looks good to go.

Issue #455, opened 2022-11-30
Build failure - Ubuntu 22.10 / gcc-12


danmilea:
Build open-amp for Linux as described in build_linux in
https://github.com/OpenAMP/open-amp/blob/main/.github/actions/build_ci/entrypoint.sh 

On Ubuntu 22.10 with default compiler gcc-12 the compilation will fail.
[build_failure_gcc12_ubuntu22.10.txt](https://github.com/OpenAMP/open-amp/files/10119563/build_failure_gcc12_ubuntu22.10.txt)

sergeyblagodarnyy:
Try to remove `-Werror` compiling option.

PR #533, opened 2023-11-06
Fix https://github.com/OpenAMP/open-amp/issues/455

danmilea:
The if (&rpc->ept) comparison will always evaluate as 'true', since rpc == &rpc->ept and rpc is previously checked to be valid.

This produces an error when building open-amp with gcc >= 13.


arnopo (changes requested):
LGTM but the commit subject have to be updated

"service: rpmsg_rpc: fix build warning" 

The reference to the github issue should not be part of the commit message, already referenced in the PR. I did the link between the issue and the PR in github

arnopo (approved):

tnmysh (approved):
LGTM.

PR #521, opened 2023-10-16
openamp: add new API rpmsg_virtio_get_rx_buffer_size()

CV-Bowen:
With this API, user can get the rx buffer size, and use this buffer safer in endpoint callback function.
Follow shows an example:
```c
static int rpmsgfs_ept_cb(FAR struct rpmsg_endpoint *ept,
                          FAR void *data, size_t len, uint32_t src,
                          FAR void *priv)
{
    /* No need extra stack variable or malloced buffer, user can fill the response data in the data directly,
     * And to avoid the data overflow, user can call rpmsg_virtio_get_rx_buffer_size() to get the limitation
     * of input data.
     */

    /* Call rpmsg_send(ept, data, data_len) to send the response data */
    rpmsg_send(ept, data, len);
}
```

PR #535, opened 2023-11-07
apps: Add use of rpmsg_virtio_get_tx/rx_buffer_size API

arnopo:
Print the max size of the RX and TX buffer payloads. This allows to test the new API implemented in #521 in the CI.

### DIFF
--- a/apps/examples/echo/rpmsg-echo.c
+++ b/apps/examples/echo/rpmsg-echo.c
@@ -75,6 +75,10 @@ int app(struct rpmsg_device *rdev, void *priv)
 	}
 
 	LPRINTF("Successfully created rpmsg endpoint.\r\n");
+
+	LPRINTF("RPMsg device TX buffer size: %#x\r\n", rpmsg_virtio_get_tx_buffer_size(rdev));
+	LPRINTF("RPMsg device RX buffer size: %#x\r\n", rpmsg_virtio_get_rx_buffer_size(rdev));
+
 	while(1) {
 		platform_poll(priv);
 		/* we got a shutdown request, exit */
--- a/apps/examples/echo/rpmsg-ping.c
+++ b/apps/examples/echo/rpmsg-ping.c
@@ -88,7 +88,6 @@ static void rpmsg_name_service_bind_cb(struct rpmsg_device *rdev,
 				       RPMSG_ADDR_ANY, dest,
 				       rpmsg_endpoint_cb,
 				       rpmsg_service_unbind);
-
 }
 
 /*-----------------------------------------------------------------------------*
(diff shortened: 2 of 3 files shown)


uLipe (approved):

apps/examples/echo/rpmsg-echo.c
```diff
@@ -75,6 +75,10 @@ int app(struct rpmsg_device *rdev, void *priv)
 	}
 
 	LPRINTF("Successfully created rpmsg endpoint.\r\n");
+
+	LPRINTF("RPMsg device RX buffer size: %#x\r\n", rpmsg_virtio_get_tx_buffer_size(rdev));
```
edmooring, line 79: This line should say "TX buffer size".

apps/examples/echo/rpmsg-ping.c
```diff
@@ -131,6 +130,9 @@ int app (struct rpmsg_device *rdev, void *priv)
 		return ret;
 	}
 
+	LPRINTF("RPMsg driver RX buffer size: %#x\r\n", rpmsg_virtio_get_tx_buffer_size(rdev));
```
edmooring, line 133: Again, "TX buffer size".

apps/tests/msg/rpmsg-ping.c
```diff
@@ -128,13 +128,15 @@ int app (struct rpmsg_device *rdev, void *priv)
...
 			       RPMSG_ADDR_ANY, RPMSG_ADDR_ANY,
 			       rpmsg_endpoint_cb, rpmsg_service_unbind);
-
 	if (ret) {
 		LPERROR("Failed to create RPMsg endpoint.\r\n");
 		metal_free_memory(i_payload);
 		return ret;
 	}
 
+	LPRINTF("RPMsg driver RX buffer size: %#x\r\n", rpmsg_virtio_get_tx_buffer_size(rdev));
```
edmooring, line 137: And again, "TX buffer size".

edmooring (changes requested):
Other than the above text fixes, looks good to me.

tnmysh:
@arnopo Can we take care of this change once demos are moved to openamp-system-reference?

apps/examples/echo/rpmsg-echo.c
arnopo: copy past without using my brain... thanks!

arnopo:
> @arnopo Can we take care of this change once demos are moved to openamp-system-reference?

I'm using these applications for few last integration tests before merging PR. So I would prefer to not wait the move

tnmysh (approved):
In that case looks good to me.

PR #540, opened 2023-12-04
remoteproc_virtio: initialize the virtio device id when create vdev

CV-Bowen:
Should init the device ID when create the virtio device.

### DIFF
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -254,6 +254,7 @@ rproc_virtio_create_vdev(unsigned int role, unsigned int notifyid,
 	rpvdev->vdev_rsc_io = rsc_io;
 
 	vdev->notifyid = notifyid;
+	vdev->id.device = vdev_rsc->id;
 	vdev->role = role;
 	vdev->reset_cb = rst_cb;
 	vdev->vrings_num = num_vrings;


lib/remoteproc/remoteproc_virtio.c
```diff
@@ -254,6 +254,7 @@ rproc_virtio_create_vdev(unsigned int role, unsigned int notifyid,
 	rpvdev->vdev_rsc_io = rsc_io;
 
 	vdev->notifyid = notifyid;
+	vdev->id.device = vdev_rsc->id;
```
edmooring, line 257: I assume this is for compliance with the virtio specification, since it doesn't appear that this field is used anywhere in the library.

Also, what about the vendor field that is also in struct virtio_device_id? Should we also set it?

CV-Bowen: Yes, we want more virtio devices to support the remoteproc transport layer not only the rpmsg device. Then we found the virtio device id is not initialized based on the resource table.

I will update the PR to initialize the vendor field.

arnopo: @edmooring, @CV-Bowen 
I don't think that you can get neither the `vendor` nor the `version` of the vdev in  from the `fw_rsc_vdev` struct in the resource table.
Does it mandatory to support some other virtio type ?

CV-Bowen: @arnopo In my opinion, from a software design perspective, a virtio transport layers should support all the virtio devices like the mmio/pci transport layer, the remoteproc transport layer should be the same.

arnopo: > @arnopo In my opinion, from a software design perspective, a virtio transport layers should support all the virtio devices like the mmio/pci transport layer, the remoteproc transport layer should be the same.

Yes 100% agree with that. My point here is that if  the `vendor` and the `version` fields are mandatory, it would require an evolution of the resource table.

edmooring: > > @arnopo In my opinion, from a software design perspective, a virtio transport layers should support all the virtio devices like the mmio/pci transport layer, the remoteproc transport layer should be the same.
> 
> Yes 100% agree with that. My point here is that if the `vendor` and the `version` fields are mandatory, it would require an evolution of the resource table.

Yeah. And there isn't a version field in the virtio device structure. Figuring out a backwards compatible way to do this is going to be a challenge.

edmooring: This is going to have to get to the Linux remoteproc community as well.

CV-Bowen: @arnopo @edmooring  So, let's only initialize the virtio device ID for now?

arnopo (approved):

edmooring (approved):

tnmysh (approved):

PR #541, opened 2023-12-04
virtio.h: add memory operation for virtio device

CV-Bowen:
Buffer management is different for different transport layer:

For MMIO transport layer, the buffer can direclty malloced from
the gust os heap beacase the hypervisor can access all the memmory own
by guest os.

For remoteproc transpor layer, the buffer should be malloced from
the share memory region to make sure the remote core can access this
buffer too.

So add memory ops in virtio device to make different transport/device can
implement their own share memory management.

### DIFF
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -152,6 +152,7 @@ struct virtio_device_id {
 typedef void (*virtio_dev_reset_cb)(struct virtio_device *vdev);
 
 struct virtio_dispatch;
+struct virtio_memory_ops;
 
 /** @brief Device features. */
 struct virtio_feature_desc {
@@ -196,6 +197,9 @@ struct virtio_device {
 
 	/** Virtio dispatch table */
 	const struct virtio_dispatch *func;
+
+	/**< Virtio device memory operations */
+	const struct virtio_memory_ops *mmops;
 
 	/** Private data */
 	void *priv;
@@ -280,6 +284,14 @@ struct virtio_dispatch {
 
 	/** Notify the other side that a virtio vring as been updated. */
 	void (*notify)(struct virtqueue *vq);
+};
+
+struct virtio_memory_ops {
+	/** Allocate memory from the virtio device. */
+	void *(*alloc)(struct virtio_device *dev, size_t size, size_t align);
+
+	/** Free memory allocated from the virtio device. */
+	void (*free)(struct virtio_device *dev, void *buf);
 };
 
 /**
@@ -496,6 +508,54 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
 		return -ENXIO;
 
 	vdev->func->reset_device(vdev);
+	return 0;
+}
+
+/**
+ * @brief Allocate buffer from the virtio device
+ *
+ * @param vdev	Pointer to virtio device structure.
+ * @param buf	Pointer to the allocated buffer (virtual address).
+ * @param size	Allocated buffer size.
+ * @param align	Allocated buffer alignment.
+ *
+ * @return 0 on success, otherwise error code.
+ */
+static inline int virtio_alloc_buf(struct virtio_device *vdev, void **buf,
+				   size_t size, size_t align)
+{
+	if (!vdev || !buf)
+		return -EINVAL;
+
+	if (!vdev->mmops || !vdev->mmops->alloc)
+		return -ENXIO;
+
+	*buf = vdev->mmops->alloc(vdev, size, align);
+	if (!*buf)
+		return -ENOMEM;
+
+	return 0;
+}
+
+/**
+ * @brief Free the buffer allocated by \ref virtio_alloc_buf from the virtio
+ * device.
+ *
+ * @param vdev	Pointer to virtio device structure.
+ * @param buf	Buffer need to be freed.
+ *
+ * @return 0 on success, otherwise error code.
+ */
+static inline int virtio_free_buf(struct virtio_device *vdev, void *buf)
+{
+	if (!vdev)
+		return -EINVAL;
+
+	if (!vdev->mmops || !vdev->mmops->free)
+		return -ENXIO;
+
+	vdev->mmops->free(vdev, buf);
+
 	return 0;
 }
 


lib/include/openamp/remoteproc_virtio.h
```diff
@@ -39,6 +39,11 @@ extern "C" {
 /* define vdev notification function user should implement */
 typedef int (*rpvdev_notify_func)(void *priv, uint32_t id);
 
+/* Define remoteproc virtio memory management function */
+struct remoteproc_virtio;
+typedef void *(*rpvdev_alloc_buf)(struct remoteproc_virtio *rpvdev, size_t size, size_t align);
```
xiaoxiang781216, line 44: should we replace `struct remoteproc_virtio *rpvdev` with `void *priv` like other callback?

CV-Bowen: Yes, I will update.

lib/include/openamp/virtio.h
```diff
@@ -277,6 +283,54 @@ struct virtio_dispatch {
...
+ * @param size		Alloc buffer size.
+ * @param align		Buffer alignment.
+ *
+ * @return Pointer to the return buffer. NULL indicates no enough buffer
+ *         or some errors happened.
+ */
+static inline void *virtio_alloc_buf(struct virtio_device *vdev,
+				     size_t size, size_t align)
+{
+	return vdev->func->alloc_buf(vdev, size, align);
```
uLipe, line 299: Check argumentts:

* vdev -> return -EINVAL if NULL;
* alloc_buf -> function may be not supported, check if it exists first and return -ENXIO if NULL

lib/include/openamp/virtio.h
```diff
@@ -277,6 +283,54 @@ struct virtio_dispatch {
...
+
+/**
+ * @brief Free a buffer alloced from transport layer.
+ *
+ * @param vdev		Pointer to virtio device structure.
+ * @param buf		Pointer to an allocated buffer from transport layer.
+ */
+static inline void virtio_free_buf(struct virtio_device *vdev, void *buf)
+{
+	vdev->func->free_buf(vdev, buf);
```
uLipe, line 331: Check argumentts:

* vdev -> return -EINVAL if NULL;
* free_buf -> function may be not supported, check if it exists first and return -ENXIO if NULL

lib/include/openamp/virtio.h
```diff
@@ -227,6 +227,12 @@ __deprecated void virtio_describe(struct virtio_device *dev, const char *msg,
  */
 
 struct virtio_dispatch {
+	/** Alloc buffer from transport layer heap */
+	void *(*alloc_buf)(struct virtio_device *vdev, size_t size, size_t align);
+
+	/** Free buffer to transport layer heap */
+	void (*free_buf)(struct virtio_device *vdev, void *buf);
+
```
arnopo, line 235: Seems to me that it is not the good place to add these functions
The memory allocation/free is independent from the virtio layer as addressed directly to the transport layer.
The `virtio_dispatch` is used for communication between the virtio layer and the transport layer
What about  a new structure similar to  `virtio_dispatch` ?  `virtio_mem_ops` or a better name...

lib/include/openamp/virtio.h
```diff
@@ -277,6 +283,54 @@ struct virtio_dispatch {
...
+ * @return Pointer to the return buffer. NULL indicates no enough buffer
+ *         or some errors happened.
+ */
+static inline void *virtio_zalloc_buf(struct virtio_device *vdev,
+				      size_t size, size_t align)
+{
+	void *ptr = virtio_alloc_buf(vdev, size, align);
+
+	if (ptr)
+		memset(ptr, 0, size);
```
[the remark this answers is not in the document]
CV-Bowen: Thanks, I will remove this function.

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -16,6 +16,29 @@
...
+static void rproc_virtio_free_buf(struct virtio_device *vdev, void *buf)
+{
+	struct remoteproc_virtio *rpvdev;
+
+	rpvdev = metal_container_of(vdev, struct remoteproc_virtio, vdev);
+	if (!rpvdev->free_buf)
+		return;
+
+	rpvdev->free_buf(rpvdev, buf);
+}
```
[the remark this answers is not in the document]
CV-Bowen: Thanks for your suggestion, I'm a little confused, my motivation is that different platforms may have different memory management strategies for remoteproc transport layer, so I introduce function `rproc_virtio_set_mm_callback()` to let the platform set their own share memory allocate and free functions.

Do you mean we should add a memory operations to the virtio device and then we can make different virtio devices implement their own share memory management strategies?

lib/include/openamp/virtio.h
CV-Bowen: Seems better, I will try this, Thanks.

lib/include/openamp/virtio.h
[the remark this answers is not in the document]
CV-Bowen: Done

lib/include/openamp/virtio.h
CV-Bowen: Done, please review again @arnopo

arnopo:
Hi @CV-Bowen,

Your proposal seems reasonable.

For your information, in the meantime, there is a working group initiated by ARM around a new Virtio transport called [virtio-msg](https://www.youtube.com/watch?v=8VyWp2s1cag).

This group aims to propose a new Virtio transport that should support virtualization and AMP systems. One objective is also to address memory allocation challenges.

My current concern with your PR is that Integrating your work now would imply that we probably have two APIs to support or deprecate this one in the short/middle term.

We have to discuss with the OpenAMP team to determine the strategy we want.
I will come back to you with a decision.

CV-Bowen:
@arnopo Thank you for your reply and waiting for your decisions.

xiaoxiang781216 (approved):

arnopo:
> @arnopo Thank you for your reply and waiting for your decisions.

We discussed the PR in the OpenAMP meeting. The decision is to postpone this PR to the next release to give more time to verify that this approach will also fit the other virtio transports.
It is probably not convenient for you, but we prefer not to introduce a new API before ensuring that it is generic enough.

xiaoxiang781216:
> > @arnopo Thank you for your reply and waiting for your decisions.
> 
> We discussed the PR in the OpenAMP meeting. The decision is to postpone this PR to the next release to give more time to verify that this approach will also fit the other virtio transports. It is probably not convenient for you, but we prefer not to introduce a new API before ensuring that it is generic enough.

when will the next release out? it's always better to merge this patch immediately after the release branch? so the community could try this change as soon as possible.

arnopo:
> > > @arnopo Thank you for your reply and waiting for your decisions.
> > 
> > 
> > We discussed the PR in the OpenAMP meeting. The decision is to postpone this PR to the next release to give more time to verify that this approach will also fit the other virtio transports. It is probably not convenient for you, but we prefer not to introduce a new API before ensuring that it is generic enough.
> 
> when will the next release out? it's always better to merge this patch immediately after the release branch? so the community could try this change as soon as possible.

That seems reasonable  to me , need first reviews from @edmooring  and/or @tnmysh

lib/include/openamp/virtio.h
```diff
@@ -497,6 +509,38 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
...
+ * @param size	Allocated buffer size.
+ * @param align	Allocated buffer alignment.
+ *
+ * @return The allocated buffer address.
+ */
+static inline void *virtio_alloc_buf(struct virtio_device *vdev,
+				     size_t size, size_t align)
+{
+	if (!vdev->mmops || !vdev->mmops->alloc)
+		return NULL;
```
tnmysh, line 525: The change mostly looks good.
Only concern is about return value. How to differentiate if op is not registered or buffer allocation actually failed?

Also is it fatal if buffer allocation is failed ?? Do we need assertion ?

CV-Bowen: Done

CV-Bowen:
@arnopo @tnmysh Could you take a look again? and could this patch be included in the 2024.10 Release Version.

lib/include/openamp/virtio.h
CV-Bowen: Add return value for `virtio_alloc/free_buff()`

lib/include/openamp/virtio.h
CV-Bowen: Done

lib/include/openamp/virtio.h
CV-Bowen: Done

lib/include/openamp/virtio.h
```diff
@@ -499,6 +511,49 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
...
+static inline int virtio_alloc_buf(struct virtio_device *vdev, void **buf,
+				   size_t size, size_t align)
+{
+	if (!vdev || !buf)
+		return -EINVAL;
+
+	if (!vdev->mmops || !vdev->mmops->alloc)
+		return -ENXIO;
+
+	*buf = vdev->mmops->alloc(vdev, size, align);
```
tnmysh, line 533: If *buf == NULL, then this should return -ENOMEM.

CV-Bowen: Thanks, Done

xiaoxiang781216 (approved):

uLipe (approved):

tnmysh (approved):

lib/include/openamp/virtio.h
```diff
@@ -499,6 +511,51 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
...
+static inline int virtio_free_buf(struct virtio_device *vdev, void *buf)
+{
+	if (!vdev)
+		return -EINVAL;
+
+	if (!vdev->mmops || !vdev->mmops->free)
+		return -ENXIO;
+
+	vdev->mmops->free(vdev, buf);
+	return 0;
```
arnopo, line 556: add a blank line before

lib/include/openamp/virtio.h
```diff
@@ -499,6 +511,51 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
...
+	if (!vdev || !buf)
+		return -EINVAL;
+
+	if (!vdev->mmops || !vdev->mmops->alloc)
+		return -ENXIO;
+
+	*buf = vdev->mmops->alloc(vdev, size, align);
+	if (!*buf)
+		return -ENOMEM;
+	return 0;
```
arnopo, line 536: nit:  Add a blank line before

lib/include/openamp/virtio.h
```diff
@@ -499,6 +511,51 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
 	return 0;
 }
 
+/**
+ * @brief Allocate buffer from the virtio device
+ *
+ * @param vdev	Pointer to virtio device structure.
+ * @param buf	Pointer to the allocated buffer.
```
arnopo, line 518: Today we are working at virtio device level with virtual address.  could be nice to indicate it in comment
`Pointer to the allocated buffer (virtual address)`

lib/include/openamp/virtio.h
```diff
@@ -499,6 +511,51 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
...
+		return -ENXIO;
+
+	*buf = vdev->mmops->alloc(vdev, size, align);
+	if (!*buf)
+		return -ENOMEM;
+	return 0;
+}
+
+/**
+ * @brief Free buffer to the virtio device
```
arnopo, line 540: Perhaps it is obvious for us, but as there libmetal also provides metal_allocate_memoryI would add the information
 `Free the buffer allocated by \ref virtio_alloc_buf from the virtio device`

arnopo (commented):
LGTM, few minor comments

lib/include/openamp/virtio.h
CV-Bowen: Done

lib/include/openamp/virtio.h
CV-Bowen: Done

lib/include/openamp/virtio.h
CV-Bowen: Done

lib/include/openamp/virtio.h
CV-Bowen: Done

edmooring (approved):
Looks good to go.

arnopo:
I have updated the commit message to fix some typos.  commit  sha1 is now 4ace3543aea62d9356dec5f044111187a8691190

PR #542, opened 2023-12-04
remoteproc_virtio: add shm_io for remoteproc virtio

CV-Bowen:
Now the rpmsg device need pass a share memory io region to rpmsg_init_vdev().
I think every virtio device need a share memory io to access the share memory and this region should be provided by the transport layer. Like the MMIO transport layer did in OpenAMP.

So I add shm_io in the remoteproc virtio device and also provide an API to set it.

### DIFF
--- a/lib/include/openamp/remoteproc_virtio.h
+++ b/lib/include/openamp/remoteproc_virtio.h
@@ -50,6 +50,9 @@ struct remoteproc_virtio {
 	/** Metal I/O region of vdev_info, can be NULL */
 	struct metal_io_region *vdev_rsc_io;
 
+	/** Metal I/O region of vdev share memory */
+	struct metal_io_region *shm_io;
+
 	/** Notification function */
 	rpvdev_notify_func notify;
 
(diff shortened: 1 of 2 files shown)


tnmysh:
Any usage of this API anywhere ? Either in libmetal or OpenAMP ?
If not, may be not right time to have this change.
Also we don't alwasy assume shm_io will be always used. dmabuf based memory allocation is also possible. I think such use case was posted by TI in another PR.

@arnopo, @edmooring any thoughts?

arnopo:
Quite agree with @tnmysh.  This PR introduces an unused field. If possible I would prefer it to be integrated in a PR that provides the full picture of the need.

@tnmysh : I remember we discussed the possibility of using dmabuf-based memory allocation in a meeting, but I don't see any associated PR or issue. Could you please provide a reference if you have it?

edmooring (commented):
Are there any consumers for this new API? I can't find any in either open-amp or openamp-system-reference.

CV-Bowen:
@arnopo @tnmysh @edmooring Thanks, I use this API in NuttX (but code not up streamed). We want make the remoteproc transport layer supports more virtio devices.
I will continue this PR when I up streamed the NuttX part code.

CV-Bowen:
> @tnmysh : I remember we discussed the possibility of using dmabuf-based memory allocation in a meeting, but I don't see any associated PR or issue. Could you please provide a reference if you have it?

@tnmysh I want to continue this PR and I'm interested in the dma method. That will be very thankfully if you can proviud a reference if you have it.

tnmysh:
@arnopo , @CV-Bowen here is PR that mentions using dma-buf: 

https://github.com/OpenAMP/libmetal/pull/82

I remember TI folks were working to update that PR. Not sure about latest status.

xiaoxiang781216:
don't need anymore? @CV-Bowen

PR #545, opened 2023-12-26
lib: Add include directory to open_amp-static target

zaporozhets:
Hi,

The following PR adds include directory in the open_amp-static target, enabling building and linking static library using CMake subdirectory. This allows to have open_amp library as git submodule and build it with the parent project as cmake dependency.

For example:

```
add_subdirectory(open-amp)


target_link_libraries(main_app
    PUBLIC
    open_amp-static
)
```

Let me know if you have any suggestions on the target name or better ideas on implementation.


tnmysh (approved):
Looks Good to me.

edmooring (approved):
Looks good to me as well.

arnopo (approved):

PR #551, opened 2024-02-06
Fixes #550: OpenAMP elf loader loads ELF sections to their load addresses

### LINKED ISSUE #550: OpenAMP ELF loader loads ELF sections to their load addresses.
Unlike the Linux rproc elf loader, the OpenAMP elf loader loads the remote elf program segments to their load addresses instead of their link addresses. This results in memory corruption when the remote firmware attempts to relocate segments from their link addresses to load addresses at runtime such as the data section.

UmairKhanUET:
Unlike the Linux rproc elf loader, the OpenAMP elf loader loads the remote elf program segments to their load addresses instead of their link addresses. This results in memory corruption when the remote firmware attempts to relocate segments from their link addresses to load addresses at runtime such as the data section.

Signed-off-by: Umair Khan <umair_khan@mentor.com>

### DIFF
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -368,7 +368,7 @@ static const void *elf_next_load_segment(void *elf_info, int *nseg,
 		if (!phdr)
 			return NULL;
 		elf_parse_segment(elf_info, phdr, &p_type, noffset,
-				  da, NULL, nfsize, nmsize);
+				  NULL, da, nfsize, nmsize);
 		*nseg = *nseg + 1;
 	}
 	return phdr;


arnopo (approved):
LGTM,
@tnmysh , @edmooring could you verify that this does not introduce regression on yopur platforms?

tnmysh:
Yes, I will get back on this. I was wondering how its working without this patch.

@bentheredonethat FYI.

arnopo:
> Yes, I will get back on this. I was wondering how its working without this patch.
> 
> @bentheredonethat FYI.

Probably because you have same memory mapping regarding the main and remote processor.

UmairKhanUET:
> Yes, I will get back on this. I was wondering how its working without this patch.
> 
> @bentheredonethat FYI.

May be because generally when firmware runs out of RAM, the ELF sections are placed at same link and load addresses. But it doesn't stop one from having different addresses either. May be this is the first time we are seeing such a case.

tnmysh:
@arnopo I have been trying to look for history on when this code was used on Xilinx's platforms. I can't find it being used currently or anytime in near past. It's very hard to prepare setup for this atleast with 2023 petalinux.

The code looks okay for me, and we don't have any use case currently that is practicing this code. I am okay with code, and sure no regression will happen on xlnx platforms as it's not being used.

edmooring (approved):
This change looks correct to me. 
With the default CMake options, this code can only be built if the target is zynqmp_r5. It also has to be specifically enabled with WITH_LOAD_FW=on. The default Petalinux/Yocto build for the R5 forces this option to "off". 
To my knowledge, this code has not been exercised, much less tested, in either the official OpenAMP repository or in the Xilinx/AMD downstream in at least 5 years.

arnopo:
Thanks  @edmooring and @tnmysh for the feedback.

We have to find a way to better test it.
Look to me that the load code could be optimized to make this more simple with less footprint. 
let's add this in agenda of next open-amp meeting call.

PR #554, opened 2024-02-08
remoteproc_virtio: fix vrings_info structure init

arnopo:
Only the first index of the vring_info is set to 0. Apply the memset to the whole memory allocated.


edmooring (approved):
Looks good to go.

Issue #501, opened 2023-07-14
Move examples to openamp-system-reference repo.


tnmysh:
Openamp library isn't the right place to have demo and examples.

Move examples to openamp-system-reference repository without breaking backward compatibility.

tnmysh:
@arnopo , @wmamills 

Created this PR in openamp-system-reference repo to address this issue: https://github.com/OpenAMP/openamp-system-reference/pull/44

Once that is merged I will create another to remove demos from this repo.

arnopo:
Applications are now available in the OpenAMP system reference. This ticket will be closed when the OpenAMP library is cleaned up (target release: 2025.4).

arnopo:
Clean-up done in  https://github.com/OpenAMP/open-amp/pull/643

PR #564, opened 2024-03-15
apps: Move common function declaration to common header

glneo:
Several platform_*() functions are common across the example machines. They actually have to be as they are consumed by example apps that build across these machines. Move these common declarations to common a header.


tnmysh:
Hi @glneo,

We are not taking new changes in apps Xilinx platform specific directories.
This code should be moved to openamp-system-reference.

So, this refactoring can be avoided.

glneo:
Yes it should be moved, I was the one who originally suggested that (I'm Andrew with TI in case my username hid that).

But this fix has been bugging me and I got tired of waiting until these examples are moved. Looks like the current plan is to move them over by [October 2024](https://github.com/OpenAMP/libmetal/issues/246) so only 7 more months from now.. Would like to just fix it here first.

tnmysh:
> Yes it should be moved, I was the one who originally suggested that (I'm Andrew with TI in case my username hid that).

Yeah I got that.

> 
> But this fix has been bugging me and I got tired of waiting until these examples are moved.

In such case, right PR would be to move demos to openamp-system-reference and removing it from here. Few PRs in apps area have been rejected before for the same reason. I believe correct PR would be to remove it from this repo first. We can have it as part of #501.

> Looks like the current plan is to move them over by [October 2024](https://github.com/OpenAMP/libmetal/issues/246) so only 7 more months from now.. Would like to just fix it here first.

glneo:
> In such case, right PR would be to move demos to openamp-system-reference and removing it from here. Few PRs in apps area have been rejected before for the same reason. I believe correct PR would be to remove it from this repo first. We can have it as part of #501.

Would you like me to do that then? I've asked if I could make that change here before but was told some Xilinx internal testing still depended on it and so y'all would move it. That was 8 months ago..

tnmysh:
> > In such case, right PR would be to move demos to openamp-system-reference and removing it from here. Few PRs in apps area have been rejected before for the same reason. I believe correct PR would be to remove it from this repo first. We can have it as part of #501.
> 
> Would you like me to do that then? I've asked if I could make that change here before but was told some Xilinx internal testing still depended on it and so y'all would move it. That was 8 months ago..

Xilinx's internal testing is moved to downstream repo. Also, there was discussion started to create platform agnostic baremetal demos in openamp-system-reference. I prefer not to add/remove any apps code in Xilinx's platform until new demos are available in openamp-system-reference. However, I leave decision to others.

tammyleino (approved):

arnopo (approved):
A discussed in system-reference meeting waiting migration of the samples and tests in system-reference repo we can merge this one ( after @tnmysh approve)

tnmysh (approved):

PR #566, opened 2024-03-15
apps: linux_rpc_demo: Do not access members of rpmsg_endpoint

glneo:
The contents of this struct are internal to the RPMsg layer. Access should only happen using accessor functions.

### DIFF
--- a/apps/examples/linux_rpc_demo/linux_rpc_demod.c
+++ b/apps/examples/linux_rpc_demo/linux_rpc_demod.c
@@ -82,10 +82,9 @@ int rpmsg_handle_open(void *data, struct rpmsg_rpc_svr *rpcs)
 	char *buf;
 	struct rpmsg_rpc_resp_open rpc_open_resp;
 	int payload_size = sizeof(rpc_open_resp);
-	struct rpmsg_endpoint *ept = &rpcs->ept;
 	int fd, ret;
 
-	if (!rpc_open_req || !ept)
+	if (!rpc_open_req)
 		return -EINVAL;
 	buf = rpc_open_req->filename;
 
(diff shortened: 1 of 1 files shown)


arnopo (approved):
No strong rule that indicates that ept is an internal structure.
Nevertheless  you are right better to consider it when possible.

apps/examples/linux_rpc_demo/linux_rpc_demod.c
```diff
@@ -223,8 +223,11 @@ int rpmsg_handle_term(void *data, struct rpmsg_rpc_svr *rpcs)
 	struct rpmsg_rpc_req_term *rpc_term_req = req_ptr;
 	struct rpmsg_endpoint *ept = &rpcs->ept;
```
tnmysh, line 224: Looks like ept variable is redundant and can be removed.

glneo: It is not needed in most of the functions in this demo, it is still checked in all of them anyway. The check doesn't make sense in the first place since this var is a pointer to a variable in `rpcs` which will never be 0 unless `rpsc` is 0, so we could just check for that. Like all the issues I've found in these demos, I'm simply ignoring all the ones that do not affect my refactoring work.

tnmysh: If ept->name is  removed then "ept" and "rpcs" both are unused variable from this function. Can we remove them ? I agree with the change to remove ept->name.

glneo: Yes, we can and should remove these from several spots in this example app. But that should be done in a follow up patch, this patch simply follows the standard that all other functions use (check for !ept even if not used).

tnmysh: @arnopo , @edmooring are you okay with this?
Ideally if we should take care of unused variables in the same patch for, but if there is already follow up PR available, I am okay to ack this.

arnopo: @tnmysh 
Good ctach and your point seems to me valid. there's no need to reproduce the same bad code here. to fix it later
It's better to directly merge the good version.

glneo: Fair enough, adding a patch to this series to fix these.

xiaoxiang781216 (approved):

edmooring (approved):
Looks good to go.

apps/examples/linux_rpc_demo/linux_rpc_demod.c
arnopo: > Fair enough, adding a patch to this series to fix these.

Thanks!
@tnmysh  ok for you with this?

PR #567, opened 2024-03-15
Use priv data for name-service callback

glneo:


arnopo (approved):

xiaoxiang781216 (approved):

lib/rpmsg/rpmsg.c
```diff
@@ -273,14 +273,16 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
 			     const char *name,
 			     uint32_t src, uint32_t dest,
 			     rpmsg_ept_cb cb,
-			     rpmsg_ns_unbind_cb ns_unbind_cb)
+			     rpmsg_ns_unbind_cb ns_unbind_cb,
+			     void *priv)
```
edmooring, line 277: Why the newline?

edmooring (approved):
Otherwise, looks good to go.

lib/rpmsg/rpmsg.c
arnopo: I will merge and fix it

PR #568, opened 2024-03-15
virtio: Use virtqueue_get_buffer() helpers

glneo:
Length was being read without first invalidating. There are functions that handle invalidating and reading from vq_ring members. Use those here.


arnopo (approved):

edmooring (approved):
Looks good to go,

PR #569, opened 2024-03-15
virtio: Inline vring_init() and vring_need_event() functions

glneo:
Both of these functions should be internal to virtqueue. They should have had no usefulness outside of this file and so moving them should not cause any API issue. This simplifies the virtio ring setup code for later improvements

### DIFF
--- a/lib/include/openamp/virtio_ring.h
+++ b/lib/include/openamp/virtio_ring.h
@@ -186,31 +186,6 @@ static inline int vring_size(unsigned int num, unsigned long align)
 	return size;
 }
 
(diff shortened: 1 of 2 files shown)


arnopo (changes requested):
`vring_init()`  is used at least by Zephyr, so should be kept in the API. https://elixir.bootlin.com/zephyr/latest/source/drivers/ethernet/eth_ivshmem_queue.c#L51
 
 I don't think that vring_need_event()  can be used externally.  So Ok to me to not apply deprecation process

PR #570, opened 2024-03-15
virtqueue: Fix comment on shm_io and fix type

glneo:
This should hold a pointer to an metal_io_region, make that the type.

Also fix the comment, this region holds the address of the message buffers, not the vring's descriptor table nor available/used rings. It is only used for virt-to-phys/phys-to-vert on the buffers pointed to by these descriptors.

This comment seems to have cause an issue in virtio_mmio_drv where this region was used to translate the address of the vring descriptors. This may have worked if the vring descriptors where part of the same IO space as the buffers they point to, but this is not guaranteed to always be the case. Fix that here.

### DIFF
--- a/lib/virtio_mmio/virtio_mmio_drv.c
+++ b/lib/virtio_mmio/virtio_mmio_drv.c
@@ -325,7 +325,7 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 	virtio_mmio_write32(vdev, VIRTIO_MMIO_QUEUE_NUM, vq->vq_nentries);
 	virtio_mmio_write32(vdev, VIRTIO_MMIO_QUEUE_ALIGN, 4096);
 	virtio_mmio_write32(vdev, VIRTIO_MMIO_QUEUE_PFN,
-			    ((uintptr_t)metal_io_virt_to_phys(vq->shm_io,
+			    ((uintptr_t)metal_io_virt_to_phys(vmdev->shm_io,
 			    (char *)vq->vq_ring.desc)) / 4096);
 
 	vdev->vrings_info[vdev->vrings_num].vq = vq;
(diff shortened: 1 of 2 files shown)


arnopo (commented):
Please split in 2 commits, else LGTM

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -325,7 +325,7 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 	virtio_mmio_write32(vdev, VIRTIO_MMIO_QUEUE_NUM, vq->vq_nentries);
 	virtio_mmio_write32(vdev, VIRTIO_MMIO_QUEUE_ALIGN, 4096);
 	virtio_mmio_write32(vdev, VIRTIO_MMIO_QUEUE_PFN,
-			    ((uintptr_t)metal_io_virt_to_phys(vq->shm_io,
+			    ((uintptr_t)metal_io_virt_to_phys(vmdev->shm_io,
```
danmilea, line 328: The two source snippets are logically identical because vq->shm_io is set to vmdev->shm_io in the call to virtqueue_set_shmem_io a few line above.

glneo: Yes they are, this is not meant to be a functional change. The point is to not need to poke into the `vq` struct as part of a larger restructuring. Also the value `shm_io` inside `vq` is meant for the buffers, and here they use it for the vring pointers themselves, which only works in some cases (when they are both part of one large IO region), it is more correct to not assume this will always be true.

edmooring (approved):
Looks good to me.

lib/virtio_mmio/virtio_mmio_drv.c
arnopo: @danmilea 
any concern or it is ok for you?

danmilea: This commit doesn't change functionality in any way. It's OK to merge from my point of view.

arnopo (approved):

PR #571, opened 2024-03-16
remoteproc: Fix the NULL vs RPROC_IS_ERR() bug for remoteproc_get_rsc_table()

mengtanhzc:
* Since remoteproc_get_rsc_table() returns RPROC_ERR_PTR and never NULL, use RPROC_IS_ERR() to check the return value.

### DIFF
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -112,28 +112,27 @@ static void *remoteproc_get_rsc_table(struct remoteproc *rproc,
 	 * the caller should be responsible to release the memory
 	 */
 	rsc_table = metal_allocate_memory(len);
-	if (!rsc_table) {
-		return RPROC_ERR_PTR(-RPROC_ENOMEM);
-	}
+	if (!rsc_table)
+		return NULL;
+
 	ret = store_ops->load(store, offset, len, &img_data, RPROC_LOAD_ANYADDR,
 			      NULL, 1);
 	if (ret < 0 || ret < (int)len || !img_data) {
 		metal_log(METAL_LOG_ERROR,
 			  "get rsc failed: 0x%llx, 0x%llx\r\n", offset, len);
-		ret = -RPROC_EINVAL;
 		goto error;
 	}
 	memcpy(rsc_table, img_data, len);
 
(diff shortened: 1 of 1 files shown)


glneo:
`rsc_table` is checked against NULL again down below in the `error3:` label. It might be cleaner to just update `remoteproc_get_rsc_table()` to return NULL on failure as all the current users of that function expect that.

mengtanhzc:
> `rsc_table` is checked against NULL again down below in the `error3:` label. It might be cleaner to just update `remoteproc_get_rsc_table()` to return NULL on failure as all the current users of that function expect that.

I'll  update remoteproc_get_rsc_table() later. Thanks for the review!

mengtanhzc:
> `rsc_table` is checked against NULL again down below in the `error3:` label. It might be cleaner to just update `remoteproc_get_rsc_table()` to return NULL on failure as all the current users of that function expect that.

I have updated the commit, please review it again. Thanks a lot. 
And.. RPROC_ERR_PTR() is not in use, should I delete it?

glneo:
> And.. RPROC_ERR_PTR() is not in use, should I delete it?

I'd say yes. All these custom error codes are only used internally by the remoteproc framework, not sure why they exist at all. Everywhere else we just use the standard `errno.h` codes, and most of the `RPROC_*` are just copies of the same. We should start to look into just using the standard error codes and remove these custom error codes.

arnopo:
> > And.. RPROC_ERR_PTR() is not in use, should I delete it?
> 
> I'd say yes. All these custom error codes are only used internally by the remoteproc framework, not sure why they exist at all. Everywhere else we just use the standard `errno.h` codes, and most of the `RPROC_*` are just copies of the same. We should start to look into just using the standard error codes and remove these custom error codes.


Agree.  `RPROC_PTR_ERR`, `RPROC_IS_ERR` and `RPROC_ERR_PTR `can be removed as unused.
the  Remoteproc error codes could also be remapped to metal `errno.h` code.  In term we should not have specific error code in the Lib but used standard erno codes

arnopo (approved):

danmilea (approved):
Good to merge.

PR #572, opened 2024-03-18
remoteproc: Un-inline remoteproc_init_mem() and remoteproc_add_mem()

glneo:
Having these functions as inline header functions forces the contents of struct remoteproc_mem to remain externally exposed. It also does not save space in most cases as every call-site gets the full contents of the function instead of just a call instruction. Make these normal functions like all others in remoteproc.h.


arnopo (approved):

edmooring (approved):
Looks good to go.

PR #578, opened 2024-03-27
Safer use of strncpy patch

tarmasr:
@arnopo : as requested

This patch adds 0-termination of the destination buffer after each call of strncpy() in the open-amp project. In one place optional {...} braces were added to an else branch because its if branch now needs braces.

### DIFF
--- a/lib/rpmsg/rpmsg.c
+++ b/lib/rpmsg/rpmsg.c
@@ -142,6 +142,7 @@ int rpmsg_send_ns_message(struct rpmsg_endpoint *ept, unsigned long flags)
 	ns_msg.flags = flags;
 	ns_msg.addr = ept->addr;
 	strncpy(ns_msg.name, ept->name, sizeof(ns_msg.name));
+	ns_msg.name[sizeof(ns_msg.name) - 1] = '\0';
 	ret = rpmsg_send_offchannel_raw(ept, ept->addr,
 					RPMSG_NS_EPT_ADDR,
 					&ns_msg, sizeof(ns_msg), true);
@@ -276,6 +277,7 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
 			     rpmsg_ns_unbind_cb ns_unbind_cb)
 {
 	strncpy(ept->name, name ? name : "", sizeof(ept->name));
+	ept->name[sizeof(ept->name) - 1] = '\0';
 	ept->refcnt = 1;
 	ept->addr = src;
 	ept->dest_addr = dest;
--- a/apps/system/linux/machine/generic/platform_info.c
+++ b/apps/system/linux/machine/generic/platform_info.c
@@ -145,6 +145,7 @@ static int sk_unix_client(const char *descr)
 	addr.sun_family = AF_UNIX;
 	strncpy(addr.sun_path, descr + strlen(UNIX_PREFIX),
 		sizeof addr.sun_path);
+	addr.sun_path[sizeof(addr.sun_path) - 1] = '\0';
 	if (connect(fd, (struct sockaddr *)&addr, sizeof(addr)) >= 0) {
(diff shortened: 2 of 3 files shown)


lib/rpmsg/rpmsg.c
```diff
@@ -142,6 +142,7 @@ int rpmsg_send_ns_message(struct rpmsg_endpoint *ept, unsigned long flags)
 	ns_msg.flags = flags;
 	ns_msg.addr = ept->addr;
 	strncpy(ns_msg.name, ept->name, sizeof(ns_msg.name));
+	ns_msg.name[sizeof(ns_msg.name) - 1] = '\0';
```
arnopo, line 145: this one seems useless as the '\0' has already been addes in `rpmsg_register_endpoint`

arnopo (changes requested):
please 
- split into 2 commits ( one for the apps, one for the lib updates) 
- fix the CI error adding a commit subject (you can have a look as  examples  to patches that have already been merged. 
- address the minor comment

tnmysh:
LGTM after addressing Arnaud's concern.

lib/rpmsg/rpmsg.c
tarmasr: @arnopo : No -- that would be unwise, because it relies on ept->name only being set by rpmsg_register_endpoint. It's just a micro-optimization to remove the '\0' addition here, because at the next code review people might wonder why this strncpy wasn't updated to a safer version. Best to leave it.

arnopo: A comment should also address your concern without adding useless code. "Moreover, this '\0' addition can give the impression that the endpoint name could be truncated."
For instance:
`/*The strncpy is safe at this point. The '\0' has been added in rpmsg_register_endpoint(). */`

tarmasr: Thanks, but I don't want to submit a compromised patch. I have withdrawn the PR and suggest that you create your own patch if you thinks it's important enough.

tarmasr: @arnopo: Additional:

(1) If you use strncpy(), always follow it with explicit \0 termination. No exceptions, no special pleading.

(2) If you known that source and destination buffers are the same size and the the source is already 0-terminated, then use memcpy() instead of strncpy()

I suggest that you make those changes yourself. I'm not your little helper.

arnopo: > @arnopo: Additional:
> 
> (1) If you use strncpy(), always follow it with explicit \0 termination. No exceptions, no special pleading.
> 
> (2) If you known that source and destination buffers are the same size and the the source is already 0-terminated, then use memcpy() instead of strncpy()
> 
> I suggest that you make those changes yourself. I'm not your little helper.


Please keepa positive mindset. This is an open-source project based on contributions. Contributors are not 'little helpers' for the maintainers, and maintainers are not subcontractors assigned to the project. If a user finds an issue, it seems fair for them to propose a fix as a contributor to help improve the library. We review and comment/challenge when something seems to need improvement. If contributors do not agree with a comment, we discuss and try to find a solution. That's just the normal process.

That said, your proposal for the memcpy() seems to me a good compromise.

tnmysh: > Please keepa positive mindset. This is an open-source project based on contributions. Contributors are not 'little helpers' for the maintainers, and maintainers are not subcontractors assigned to the project. If a user finds an issue, it seems fair for them to propose a fix as a contributor to help improve the library. We review and comment/challenge when something seems to need improvement. If contributors do not agree with a comment, we discuss and try to find a solution. That's just the normal process.

Thanks @arnopo for clearing this point.

I was also looking another option to use `strlcpy` as safer replacement of strncpy.
https://linux.die.net/man/3/strlcpy

Thanks.

PR #585, opened 2024-04-25
rpmsg_virtio: Move get RX buffer into receive loop and returned buffer kick to common location 

glneo:


tnmysh (commented):
>  This should have
> no functional change but prepares for a later bug fix.

As commit message says this is needed for bug fix, then this change can be squashed in same commit where actual bug is fixed.

glneo:
Okay, the bug fix I have in mind has a few more dependencies so for now I'll remove that part of the commit message. The patch should stand on its own just the same.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -569,14 +569,16 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 	uint16_t idx;
 	int status;
 
-	metal_mutex_acquire(&rdev->lock);
-
-	/* Process the received data from remote node */
-	rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
+	while (true) {
```
arnopo, line 572: nitpicking, to be aligned with the other infinite loop in the lib 
```suggestion
	while (1) {
```

arnopo (commented):
Minor remark else LGTM

lib/rpmsg/rpmsg_virtio.c
glneo: Sure, changed.

arnopo (approved):

tnmysh (approved):
LGTM.

PR #589, opened 2024-05-06
apps: remove remaining master terminology

arnopo:
Some potentially offensive terms are still present in the code, Replace them.

### DIFF
--- a/apps/machine/zynq7/platform_info_remoteproc_master.c
+++ b/apps/machine/zynq7/platform_info_remoteproc_master.c
@@ -31,7 +31,7 @@ extern struct hil_platform_ops zynq_a9_proc_ops;
 #define SHM_SIZE                          0x00200000
 #define VRING0_IPI_VECT                   6
 #define VRING1_IPI_VECT                   3
-#define MASTER_CPU_ID                     0
+#define HOST_CPU_ID                       0
 #define REMOTE_CPU_ID                     1
 
 /**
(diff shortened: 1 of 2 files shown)


apps/machine/zynq7/platform_info_remoteproc_master.c
```diff
@@ -31,7 +31,7 @@ extern struct hil_platform_ops zynq_a9_proc_ops;
 #define SHM_SIZE                          0x00200000
 #define VRING0_IPI_VECT                   6
 #define VRING1_IPI_VECT                   3
-#define MASTER_CPU_ID                     0
+#define MAIN_CPU_ID                       0
```
glneo, line 34: `HOST` seems to be used more often when paired with `REMOTE`, such as in `RPMSG_HOST`.

Is this not used anywhere? If not, better to just drop these two defines.

arnopo: the `REMOTE_CPU_ID` is used, so I have kept it for readability

glneo: Okay, then still would suggest `HOST_CPU_ID` if these defines need to remain.

arnopo:
I also fix few typos in a separate commit

edmooring (approved):
Looks good to go.

iuliana-prodan (approved):

PR #598, opened 2024-06-18
virtio_mmio: Remove unneeded use of libmetal device

glneo:
A virtual metal_device is created, then the needed IO regions are created and added to this device. Immediately we extract these same regions back out and make use of them. There is no reason to do this, instead simply use the created IO regions.

This also removes the need to have struct metal_device defined to have more than one IO region (METAL_MAX_DEVICE_REGIONS), which is not the default and can change per-platform. If the libmetal library was built with the default value for METAL_MAX_DEVICE_REGIONS, then this would have led to runtime failures.

### DIFF
--- a/lib/virtio_mmio/virtio_mmio_drv.c
+++ b/lib/virtio_mmio/virtio_mmio_drv.c
@@ -141,46 +141,20 @@ const struct virtio_dispatch virtio_mmio_dispatch = {
 static int virtio_mmio_get_metal_io(struct virtio_device *vdev, uintptr_t virt_mem_ptr,
 				    uintptr_t cfg_mem_ptr)
 {
-	struct metal_device *device;
-	int32_t err;
 	struct virtio_mmio_device *vmdev = metal_container_of(vdev,
 							      struct virtio_mmio_device, vdev);
 
-	/* Setup shared memory device */
-	vmdev->shm_device.regions[0].physmap = (metal_phys_addr_t *)&vmdev->shm_mem.base;
-	vmdev->shm_device.regions[0].virt = (void *)virt_mem_ptr;
-	vmdev->shm_device.regions[0].size = vmdev->shm_mem.size;
+	/* Setup shared memory region */
+	vmdev->shm_io = metal_allocate_memory(sizeof(*vmdev->shm_io));
+	metal_io_init(vmdev->shm_io, (void *)virt_mem_ptr,
+		      (metal_phys_addr_t *)&vmdev->shm_mem.base,
+		      vmdev->shm_mem.size, -1, 0, NULL);
 
-	VIRTIO_ASSERT((METAL_MAX_DEVICE_REGIONS > 1),
-		      "METAL_MAX_DEVICE_REGIONS must be greater that 1");
-
-	vmdev->shm_device.regions[1].physmap = (metal_phys_addr_t *)&vmdev->cfg_mem.base;
-	vmdev->shm_device.regions[1].virt = (void *)cfg_mem_ptr;
-	vmdev->shm_device.regions[1].size = vmdev->cfg_mem.size;
-
-	err = metal_register_generic_device(&vmdev->shm_device);
-	if (err) {
-		metal_log(METAL_LOG_ERROR, "Couldn't register shared memory device: %d\n", err);
-		return err;
-	}
-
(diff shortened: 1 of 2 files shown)


xiaoxiang781216 (approved):

danmilea (approved):
Looks good to merge.

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -141,46 +141,20 @@ const struct virtio_dispatch virtio_mmio_dispatch = {
...
+	vmdev->shm_io = metal_allocate_memory(sizeof(*vmdev->shm_io));
+	metal_io_init(vmdev->shm_io, (void *)virt_mem_ptr,
+		      (metal_phys_addr_t *)&vmdev->shm_mem.base,
+		      vmdev->shm_mem.size, -1, 0, NULL);
+
+	/* Setup configuration region */
+	vmdev->cfg_io = metal_allocate_memory(sizeof(*vmdev->cfg_io));
+	metal_io_init(vmdev->cfg_io, (void *)cfg_mem_ptr,
+		      (metal_phys_addr_t *)&vmdev->cfg_mem.base,
+		      vmdev->cfg_mem.size, -1, 0, NULL);
```
arnopo, line 157: I'm quite puzzled by this. One of the key principles of the OpenAMP library is to support static allocation s wherever possible. If we want to extend the virtio MMIO for AMP systems, we should also support "safe" allocation.

I can see two strategies:

- Implement dynamic allocation with metal_allocate_memory() to have a robust memory allocator.
- Declare allocations statically where possible.

The OpenAMP library uses static allocation for the virtio and rpmsg layers. For the virtio transport layer (remoteproc virtio), we have dynamic allocation.

I would be in favor here of implementing static allocation as proposed by @danmilea in [danmilea/open-amp@ba16118](https://github.com/danmilea/open-amp/commit/ba161189f8346b07c3bcb9db5ebc294084fcb32d).

I notice that there is another allocation in the virtio_mmio_register_device() function.

glneo: Sure, seems reasonable, adding a follow up patch to switch to static allocations here.

glneo: And for the more general issue of dynamic allocations, I've been looking at the same as it would be a blocker for "safety".

I'd like to drop out those last couple dynamic allocations (move them to the application side, not hide them in the library as done today) but it would take a slight modification to the API. Specifically the allocations in rproc_virtio_create_vdev(). So to do any work on that we would want to make it an internal-only API which it should have been from the start.

Unfortunately as we don't separate internal from external APIs yet, someone used it in a Zephyr project sample. I've got a patch to fix that[0]. After that is in, we should mark that API internal-only so we can clean it up without breaking anyone.

https://github.com/glneo/zephyr/commit/4acc5fced3913e4ecbe35b04774b360a5293708c

arnopo: > And for the more general issue of dynamic allocations, I've been looking at the same as it would be a blocker for "safety".
> 
> I'd like to drop out those last couple dynamic allocations (move them to the application side, not hide them in the library as done today) but it would take a slight modification to the API. Specifically the allocations in rproc_virtio_create_vdev(). So to do any work on that we would want to make it an internal-only API which it should have been from the start.
> 
> Unfortunately as we don't separate internal from external APIs yet, someone used it in a Zephyr project sample. I've got a patch to fix that[0]. After that is in, we should mark that API internal-only so we can clean it up without breaking anyone.
> 
> [glneo/zephyr@4acc5fc](https://github.com/glneo/zephyr/commit/4acc5fced3913e4ecbe35b04774b360a5293708c)


`rproc_virtio_create_vdev() ` has been used to avoid to embed the remoteproc part of the library.  
The use of the remoteproc add around 2kB of code and 200 bytes of data.

Another objective  was to separate the remotproc layers from the remoteproc_virtio layer.

That said yes this part need to be reworked to support the resource table  in safety context with static allocation.
What about a new API that would replace this one?

arnopo (approved):
LGTM

edmooring (approved):
Looks good to go.

arnopo:
@danmilea : ok for you with the static allocation?

PR #600, opened 2024-06-28
remoteproc: Do not update the rproc state if rproc->ops fails

### LINKED ISSUE #599: Question about the state of rproc
I've noticed that the `rproc->state` changes even if the `rproc->ops` fails, this seems unnecessary and different from the linux remoteproc framework. Should the state not be changed if the ops fails?

https://github.com/OpenAMP/open-amp/blob/main/lib/remoteproc/remoteproc.c#L242
```
int remoteproc_start(struct remoteproc *rproc)
{
	int ret = -RPROC_ENODEV;

	if (rproc) {
		metal_mutex_acquire(&rproc->lock);
		if (rproc->state == RPROC_READY) {
			ret = rproc->ops->start(rproc);
>>>			rproc->state = RPROC_RUNNING;
		} else {
			ret = -RPROC_EINVAL;
		}
		metal_mutex_release(&rproc->lock);
	}
	return ret;
}
```

Linux remoteproc framwork:
https://github.com/torvalds/linux/blob/master/drivers/remoteproc/remoteproc_core.c#L1302
```
static int rproc_start(struct rproc *rproc, const struct firmware *fw)
{
	... ....

	/* power up the remote processor */
	ret = rproc->ops->start(rproc);
	if (ret) {
		dev_err(dev, "can't start rproc %s: %d\n", rproc->name, ret);
		goto unprepare_subdevices;
	}

	... ...

	rproc->state = RPROC_RUNNING;

	dev_info(dev, "remote processor %s is now up\n", rproc->name);

	return 0;

stop_rproc:
	rproc->ops->stop(rproc);
unprepare_subdevices:
	rproc_unprepare_subdevices(rproc);
reset_table_ptr:
	rproc->table_ptr = rproc->cached_table;

	return ret;
}
```

mengtanhzc:
* When rproc->ops fails, the state of rproc should not be updated.
Fix #599

### DIFF
--- a/lib/remoteproc/remoteproc.c
+++ b/lib/remoteproc/remoteproc.c
@@ -222,7 +222,9 @@ int remoteproc_config(struct remoteproc *rproc, void *data)
 				ret = rproc->ops->config(rproc, data);
 			else
 				ret = 0;
-			rproc->state = RPROC_READY;
+
+			if (!ret)
+				rproc->state = RPROC_READY;
 		} else {
 			ret = -RPROC_EINVAL;
 		}
@@ -238,8 +240,10 @@ int remoteproc_start(struct remoteproc *rproc)
 	if (rproc) {
 		metal_mutex_acquire(&rproc->lock);
 		if (rproc->state == RPROC_READY) {
-			ret = rproc->ops->start(rproc);
-			rproc->state = RPROC_RUNNING;
+			if (rproc->ops->start)
+				ret = rproc->ops->start(rproc);
+			if (!ret)
+				rproc->state = RPROC_RUNNING;
 		} else {
 			ret = -RPROC_EINVAL;
 		}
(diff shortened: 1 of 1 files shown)


lib/remoteproc/remoteproc.c
```diff
@@ -258,8 +261,13 @@ int remoteproc_stop(struct remoteproc *rproc)
 		    rproc->state != RPROC_OFFLINE) {
 			if (rproc->ops->stop)
 				ret = rproc->ops->stop(rproc);
-			rproc->state = RPROC_STOPPED;
-			rproc->bitmap = 0;
+			else
+				ret = 0;
```
arnopo, line 265: Any reason to  no more consider rproc->ops->stop == NULL as an error?
if no reason, pleas remove the else case

arnopo (commented):
a minor point to address then LGTM

lib/remoteproc/remoteproc.c
mengtanhzc: Yes, I made a mistake here.  Please review again, thanks!

arnopo (approved):

tnmysh (approved):

PR #602, opened 2024-07-09
openamp: replace all strncpy to strlcpy

wyr-7:


lib/rpmsg/rpmsg.c
```diff
@@ -305,7 +305,7 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
 			     rpmsg_ept_cb cb,
 			     rpmsg_ns_unbind_cb ns_unbind_cb, void *priv)
 {
-	strncpy(ept->name, name ? name : "", sizeof(ept->name));
+	strlcpy(ept->name, name ? name : "", sizeof(ept->name));
```
arnopo, line 308: I wonder if we should create our own strncpy.
we could create a `libutil/strncpy.c` that contains a `safe_strncpy` function.

```c
char *safe_strncpy(char *dest, char *src, size_t size) {
	size_t len;

	if (size == 0) {
		dest[0] = '\0';
	} else {	
		len = strnlen(src, size - 1);
		memcpy(dest, src, len);
		dest[len] = '\0';
	}
	return dest;
}

```
@edmooring, @tnmysh : any opinon/suggestion?

tnmysh: I believe that makes sense. In kernel as well, strlcpy was deprecated, and strscpy was introduced. https://www.kernel.org/doc/html/latest/core-api/kernel-api.html?highlight=declare_bitmap#c.strscpy

tnmysh: I am okay, to introduce such function.

edmooring (commented):
I think we need to look closely at how Linux RPMSG handles these cases. My impression was that the Linux kernel implementation could leave some strings not null-terminated.

lib/include/openamp/rpmsg.h
```diff
@@ -55,6 +55,7 @@ typedef void (*rpmsg_ept_release_cb)(struct rpmsg_endpoint *ept);
 typedef void (*rpmsg_ns_unbind_cb)(struct rpmsg_endpoint *ept);
 typedef void (*rpmsg_ns_bind_cb)(struct rpmsg_device *rdev,
 				 const char *name, uint32_t dest);
+typedef int (*rpmsg_notify_wait_cb)(struct rpmsg_device *rdev, uint32_t id);
```
arnopo, line 58: Please introduce this commit in a new PR, you also need to provide more information to explain why you need this update.

xiaoxiang781216:
@wyr-7 let's close this pr.

PR #605, opened 2024-08-06
rpmsg_virtio: rpmsg_deinit_vdev: Add check for empty endpoint list

bentheredonethat:
There can exist scenario where the endpoint list in a rpmsg vdev can be empty of endpoint-created nodes but the original remains, check if that is the case to ensure the while loop can properly terminate.

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -957,6 +957,15 @@ void rpmsg_deinit_vdev(struct rpmsg_virtio_device *rvdev)
 			node = rdev->endpoints.next;
 			ept = metal_container_of(node, struct rpmsg_endpoint, node);
 			rpmsg_destroy_ept(ept);
+
+			/*
+			 * The first node will still be present because the
+			 * first node is created before endpoints. So
+			 * exit the loop if only this original node remains.
+			 */
+			if (!node->next) {
+				break;
+			}
 		}
 
 		rvdev->rvq = 0;


bentheredonethat:
CC @tnmysh

wmamills:
@bentheredonethat can you describe the scenario that where this arises?

bentheredonethat:
I see this as bug in rpmsg_deinit_vdev because the initial list head is created before rpsmg-create endpoint and this adds check for that initial list head node.

bentheredonethat:
I took a further look. Because there can exist independtly the RPMsg vdev's ns_ept I believe that the alternative solution of moving the RPMsg vdev's list of endpoints to create endpoint is not valid.

Given that, I think keeping the logic as is (the existing patch) is valid.

CC @wmamills

tnmysh (approved):
LGTM.

arnopo:
The point is not clear to me I cannot figure out how we can have   "node->next == NULL" regarding `metal_list_XXX` macro
after `metal_list_init(&rdev->endpoints)`
```c
    rdev->endpoints.prev = rdev->endpoints;
    rdev->endpoints.next = rdev->endpoints;

```
After `rpmsg_register_endpoint(rdev, &rdev->ns_ept,, ...)`
```c
    rdev->endpoints.prev =  rdev->ns_ept->node;
    rdev->endpoints.next =  rdev->ns_ept->node;

    rdev->ns_ept->node.prev =  rdev->endpoints;
    rdev->ns_ept->node.next =  rdev->endpoints;

```
After ` metal_list_del(&ept->node)`
```c
    rdev->endpoints.prev =   rdev->endpoints;
    rdev->endpoints.next =   rdev->endpoints;

   rdev->ns_ept->node =  rdev->ns_ept->node;
   rdev->ns_ept->node =  rdev->ns_ept->node;
```

Please could you detail the sequence?

bentheredonethat:
we can skip this after further inspection. will close

PR #609, opened 2024-09-14
ns: acknowledge the received creation message

wyr-7:
the two phase handsake make the client could initiate the transfer immediately without the server side send any dummy message first.


arnopo:
Hello @wyr-7 
Any reason to duplicate https://github.com/OpenAMP/open-amp/pull/160?

PR #610, opened 2024-09-18
openamp/virtio.h: update vdev->features and make final_features optional

wyr-7:
add helpful judge for final_features is NULL and update vdev->features in virtio_negotiate_features

### DIFF
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -468,14 +468,15 @@ static inline int virtio_negotiate_features(struct virtio_device *vdev,
 					    uint32_t features,
 					    uint32_t *final_features)
 {
-	if (!vdev || !final_features)
+	if (!vdev)
 		return -EINVAL;
 
 	if (!vdev->func || !vdev->func->negotiate_features)
 		return -ENXIO;
 
-	*final_features = vdev->func->negotiate_features(vdev, features);
-	vdev->features = *final_features;
+	vdev->features = vdev->func->negotiate_features(vdev, features);
+	if (final_features)
+		*final_features = vdev->features;
 	return 0;
 }
 
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -475,6 +475,7 @@ static inline int virtio_negotiate_features(struct virtio_device *vdev,
 		return -ENXIO;
 
 	*final_features = vdev->func->negotiate_features(vdev, features);
+	vdev->features = *final_features;
 	return 0;
 }
 
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -204,11 +204,11 @@ static void rproc_virtio_set_features(struct virtio_device *vdev,
 static uint32_t rproc_virtio_negotiate_features(struct virtio_device *vdev,
 						uint32_t features)
 {
-	uint32_t dfeatures = rproc_virtio_get_dfeatures(vdev);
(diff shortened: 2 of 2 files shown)


lib/include/openamp/virtio.h
```diff
@@ -468,13 +468,15 @@ static inline int virtio_negotiate_features(struct virtio_device *vdev,
...
+	if (!vdev)
 		return -EINVAL;
 
 	if (!vdev->func || !vdev->func->negotiate_features)
 		return -ENXIO;
 
-	*final_features = vdev->func->negotiate_features(vdev, features);
+	vdev->features = vdev->func->negotiate_features(vdev, features);
+	if (final_features)
+		*final_features = vdev->features;
```
arnopo, line 479: Here you try 2 address 2 things 
- make `final_features `optional
- update `vdev->features`

It look like this PR should contain 4 patches

- patch 1:  update `virtio_dispatch` struct  comments to specify that `negotiate_features` return negociated feature in case of success , negative value in case of error

- patch 2:  fix rproc_virtio_negotiate_features that retun 0 instead of dfeatures & features

- patch 3: set `vdev->features` in `virtio_negotiate_features `
- patch 4: make `final_features` param optional ( code + parameter comment)

arnopo (approved):

arnopo (changes requested):
not approved but request changes...

lib/include/openamp/virtio.h
wyr-7: > Here you try 2 address 2 things
> 
> * make `final_features `optional
> * update `vdev->features`
> 
> It look like this PR should contain 4 patches
> 
> * patch 1:  update `virtio_dispatch` struct  comments to specify that `negotiate_features` return negociated feature in case of success , negative value in case of error
> * patch 2:  fix rproc_virtio_negotiate_features that retun 0 instead of dfeatures & features
> * patch 3: set `vdev->features` in `virtio_negotiate_features `
> * patch 4: make `final_features` param optional ( code + parameter comment)

Ok,  For the 4 patches there are some issues：
patch 1: negotiate_features in virtio_dispatch return type is uint32_t, so can't negative value in case of error.
patch 3 and patch 4: can we  merge to one directly like the commit?

edmooring (changes requested):
This is a good change, but it should be in its own pull request. It also should have a better explanation in the commit message, and a comment explaining what the function actually returns.

Suggestion for commit message:

Make rproc_virtio_negotiate_features() return the mask of features successfully negotiated. 

The comment for rproc_virtio_negotiate_features () could be worded similarly.

lib/include/openamp/virtio.h
wyr-7: Ok, done.

wyr-7:
> This is a good change, but it should be in its own pull request. It also should have a better explanation in the commit message, and a comment explaining what the function actually returns.
> 
> Suggestion for commit message:
> 
> Make rproc_virtio_negotiate_features() return the mask of features successfully negotiated.
> 
> The comment for rproc_virtio_negotiate_features () could be worded similarly.

wyr-7:
> This is a good change, but it should be in its own pull request. It also should have a better explanation in the commit message, and a comment explaining what the function actually returns.
> 
> Suggestion for commit message:
> 
> Make rproc_virtio_negotiate_features() return the mask of features successfully negotiated.
> 
> The comment for rproc_virtio_negotiate_features () could be worded similarly.

Ok, done.

xiaoxiang781216 (approved):

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #612, opened 2024-09-18
openamp: add VIRTIO_RING_F_MUST_NOTIFY event

wyr-7:
add VIRTIO_RING_F_MUST_NOTIFY support to can't suppress interrupt


arnopo:
I can not find this feature bit in the VIRTIO specification.  Without this prerequisite, it  seems to me not possible to integrate it in the library.
Could you describe the use case you are trying to solve with this bit?

CV-Bowen:
@arnopo This feature bit is used to prevet local core to access the share memory (avail.flag and used.flag) managered by remote core to solve the lower power issue.
And after the discussion, we think this patch is not a common modification and will close this PR. @wyr-7

PR #613, opened 2024-09-18
rpmsg_virtio.c: replace metal_cpu_yield to metal_yield

wyr-7:
The CPU yield is not supported by all OSes/processors. If not supported, this causes a busy loop that monopolizes the CPU. Replace it with the new metal_yield, it would be managed at the OS level and dispatched to metal_cpu_yield, metal_sleep_usec, or others.

### DIFF
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -10,6 +10,7 @@
 
 #include <metal/alloc.h>
 #include <metal/sleep.h>
+#include <metal/sys.h>
 #include <metal/utilities.h>
 #include <openamp/rpmsg_virtio.h>
 #include <openamp/virtqueue.h>
@@ -271,8 +272,7 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
 		} else if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK) {
 			return 0;
 		}
-		/* TODO: clarify metal_cpu_yield usage*/
-		metal_cpu_yield();
+		metal_yield();
 	}
 }
 
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -12,7 +12,7 @@
 #include <openamp/remoteproc.h>
 #include <openamp/remoteproc_virtio.h>
 #include <openamp/virtqueue.h>
-#include <metal/cpu.h>
+#include <metal/sys.h>
 #include <metal/utilities.h>
 #include <metal/alloc.h>
 
@@ -412,6 +412,6 @@ void rproc_virtio_wait_remote_ready(struct virtio_device *vdev)
 		status = rproc_virtio_get_status(vdev);
 		if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK)
 			return;
-		metal_cpu_yield();
+		metal_yield();
 	}
 }


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -271,8 +271,8 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
 		} else if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK) {
 			return 0;
 		}
-		/* TODO: clarify metal_cpu_yield usage*/
-		metal_cpu_yield();
+		/* TODO: clarify metal_sleep_usec usage*/
```
arnopo, line 274: You can remove the comment

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -271,8 +271,8 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
...
+		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
```
arnopo, line 275: Use explicit value `1000`,  RPMSG_TICKS_PER_INTERVAL is quite obscur.

arnopo (commented):
This PR seems valid to me. We should not have the CPU yield notion in OpenAMP. Waiting 1 ms is more generic, and seems enough.

Please rework your commit message to something more generic.

Commit proposal:

```
The CPU yield is not supported by all OSes/processors. If not supported, this causes
a busy loop that monopolizes the CPU. Replace it with metal_sleep_usec, which is
OS-dependent and allows the OS scheduler to switch to another thread.
```

lib/rpmsg/rpmsg_virtio.c
wyr-7: > You can remove the comment

Ok, done

lib/rpmsg/rpmsg_virtio.c
wyr-7: > Use explicit value `1000`, RPMSG_TICKS_PER_INTERVAL is quite obscur.

Ok, done

arnopo (approved):

xiaoxiang781216 (approved):

edmooring (changes requested):
I disagree with this change. It overrides decisions made in 
* lib/processor/cpu/hosted.h
* lib/processor/cpu/x86.h
* lib/processor/cpu/aarch64.h

which could surprise current users of the library.
I think changing the implementation of metal_cpu_yield() in lib/processor/generic/cpu.h to use metal_sleep_usec() would solve the same problem, with fewer surprises to our current users.

arnopo:
We discussed this with @edmooring. He is right; it is more flexible to manage this in libmetal. As `metal_cpu_yield` is processor-dependent, the proposal is to create a new libmetal API `metal_yield`.

`metal_yield` would be managed at the OS level and dispatched to metal_cpu_yield, metal_sleep_usec, or others.

wyr-7:
> We discussed this with @edmooring. He is right; it is more flexible to manage this in libmetal. As `metal_cpu_yield` is processor-dependent, the proposal is to create a new libmetal API `metal_yield`.
> 
> `metal_yield` would be managed at the OS level and dispatched to metal_cpu_yield, metal_sleep_usec, or others.



> We discussed this with @edmooring. He is right; it is more flexible to manage this in libmetal. As `metal_cpu_yield` is processor-dependent, the proposal is to create a new libmetal API `metal_yield`.
> 
> `metal_yield` would be managed at the OS level and dispatched to metal_cpu_yield, metal_sleep_usec, or others.

Ok, add metal_yield() in pr:https://github.com/OpenAMP/libmetal/pull/311, this pr is denpends on it.

edmooring (changes requested):
The change itself looks good. The commit message needs to be changed to match the new implementation. It could be as simple as "Replace it with the new metal_yield(),".

wyr-7:
> The change itself looks good. The commit message needs to be changed to match the new implementation. It could be as simple as "Replace it with the new metal_yield(),".

Ok, thanks, done.

arnopo (approved):
LGTM to rebase after https://github.com/OpenAMP/libmetal/pull/311 integration to pass the CI

xiaoxiang781216 (approved):

edmooring (approved):
Looks good to go.

tnmysh (approved):

PR #614, opened 2024-09-19
fix rpmsg_virtio_get_tx_payload_buffer possible errors not considered

wyr-7:
1. don't need check status when get_tx_payload
2. driver/rptun_notify_wait and driver/rpmsg_virtio_notify_wait may return -EAGAIN,
    so if only judge status == RPMSG_EOPNOTSUPP, it will cause busy loop in
    rpmsg_virtio_get_tx_payload_buffer;
3. add assert when get tx buffer failed


arnopo:
Please provide more details on issues you address

> 1. don't need check status when get_tx_payload
The status it checked because it can have been updated ( for instance host crash). so prefer to keep it

> 2. driver/rptun_notify_wait and driver/rpmsg_virtio_notify_wait may return -EAGAIN,
>    so if only judge status == RPMSG_EOPNOTSUPP, it will cause busy loop in
>    rpmsg_virtio_get_tx_payload_buffer;
this one seems to fix a bug: if `rpmsg_virtio_notify_wait` return RPMSG_SUCCESS, we don't call `rpmsg_virtio_get_tx_buffer`

> 3. add assert when get tx buffer failed
Pleas detail, why do you consider a logic problem? it can occurs if the host side is busy...

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -390,6 +390,8 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		}
 	}
 
+	metal_assert(!wait || rp_hdr);
```
edmooring, line 393: This effectively changes the API for the function, since the previous behavior was just to return null when wait == 0 and getting the buffer failed. With this change, instead of an error return, the system aborts. This could break client code in a very bad way. I would prefer that this commit be reverted.

edmooring (changes requested):
To summarize, I would prefer that commits 1 and 3 were removed from this pull request, and that the commit message for 2 be updated.

CV-Bowen:
@arnopo @edmooring About your questions:

> don't need check status when get_tx_payload

When the remote core crashed, local core should crashed or reboot to recovery the whole AMP system?
So how about add a new ops `panic`in remoteproc instead slove this problem in rpmsg transport layer?

> add assert when get tx buffer failed

We think when local core can't get the tx buffer for 15 seconds, there are already some issues so let's direclty assert to found this issue.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -384,11 +384,9 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		 * use metal_sleep_usec() method by default.
 		 */
 		status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
-		if (status == RPMSG_EOPNOTSUPP) {
+		if (status != RPMSG_SUCCESS) {
```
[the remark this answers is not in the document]
wyr-7: > This looks like a valid bug fix. The only change I would request is to remove the line in the commit message about redundant continue logic, since it doesn't seem to be relevant any more.

You are right, the second line commit message has been removed.

arnopo:
> @arnopo @edmooring About your questions:
> 
> > don't need check status when get_tx_payload
> 
> When the remote core crashed, local core should crashed or reboot to recovery the whole AMP system? So how about add a new ops `panic`in remoteproc instead slove this problem in rpmsg transport layer?
> 
> > add assert when get tx buffer failed
> 
> We think when local core can't get the tx buffer for 15 seconds, there are already some issues so let's direclty assert to found this issue.

I'm not in favor of deciding the behavior in the library. This should be a system decision. For instance, you can have a processor that manages some safety or critical resources. In such a case, we would not crash if the remote processor crashes.
It is better to avoid panics and asserts as much as possible, especially if the result depends on the remote processor.

CV-Bowen:
@arnopo OK, the scenes you described do indeed require these features. @wyr-7 Let's only keep the commit 2.

wyr-7:
> @arnopo OK, the scenes you described do indeed require these features. @wyr-7 Let's only keep the commit 2.

Ok, @arnopo @edmooring @CV-Bowen  commits 1 and 3 were removed and only keep the commit 2.

arnopo (commented):
Please fix commit message  ( uper case , '.', subject, ...)
```
rpmsg: fix rpmsg_virtio_get_tx_payload_buffer error

If rpmsg_virtio_notify_wait returns RPMSG_SUCCESS, we don't call
rpmsg_virtio_get_tx_buffer.
```
else look go to go

wyr-7:
> Please fix commit message ( uper case , '.', subject, ...)
> 
> ```
> rpmsg: fix rpmsg_virtio_get_tx_payload_buffer error
> 
> If rpmsg_virtio_notify_wait returns RPMSG_SUCCESS, we don't call
> rpmsg_virtio_get_tx_buffer.
> ```
> 
> else look go to go.

OK, I'll pay attention to it later. Done.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -359,19 +359,13 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
...
 	struct rpmsg_hdr *rp_hdr;
-	uint8_t virtio_status;
 	uint16_t idx;
 	int tick_count;
 	int status;
 
 	/* Get the associated remote device for channel. */
 	rvdev = metal_container_of(rdev, struct rpmsg_virtio_device, rdev);
 
-	/* Validate device state */
```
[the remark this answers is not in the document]
wyr-7: Ok,  it reserved the status check now.

lib/rpmsg/rpmsg_virtio.c
wyr-7: You are right, I have reverted this commit.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -390,11 +390,9 @@ static void *rpmsg_virtio_get_tx_payload_buffer(struct rpmsg_device *rdev,
 		 * use metal_sleep_usec() method by default.
 		 */
 		status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
-		if (status == RPMSG_EOPNOTSUPP) {
+		if (status != RPMSG_SUCCESS) {
```
arnopo, line 393: Sorry, I missed this before...

If rpmsg_virtio_notify_wait() returns an error that is not RPMSG_EOPNOTSUPP, we call metal_sleep_usec()
but here I suppose that it is not the expected behavior...Indeed rpmsg_virtio_notify_wait() has been implemented as an alternative of the the metal_sleep_usec() .

what about just removing 
```
		} else if (status == RPMSG_SUCCESS) {
			break;
```
the loop would be:
```c
while (1) {
	/* Lock the device to enable exclusive access to virtqueues */
	rp_hdr = rpmsg_virtio_get_tx_buffer(rvdev, len, &idx);
	if (rp_hdr || !tick_count)
		break;
	/*
	 * Try to use wait loop implemented in the virtio dispatcher and
	 * use metal_sleep_usec() method by default.
	 */	status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
	if (status == RPMSG_EOPNOTSUPP) {
		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
		tick_count--;
	}
}	
```

arnopo:
I would like to merge this one in the release as it fixes an issue  

@wyr-7  could you address my comment before tomorrow code  freeze?
@edmooring @tnmysh,  could you review it?

Thanks in Advance

tnmysh (approved):

tnmysh:
LGTM.

edmooring (approved):
Looks good to go.

lib/rpmsg/rpmsg_virtio.c
wyr-7: @arnopo But when rpmsg_virtio_notify_wait return rvdev->notify_wait_cb, it does not limit the return value only to RPMSG_EOPNOTSUPP, this may have other error conditions, should we continue trying to get the buffer?

arnopo: It is a good question.
In the current version of the patch we continue trying to get buffer but after calling `metal_sleep_usec()`.
this do not seem to work as expected. `rpmsg_virtio_notify_wait()` is used as an alternative of the basic `metal_sleep_usec()`. `metal_sleep_usec()` should be called only if `rpmsg_virtio_notify_wait` ops is not supported.

I would be in favor of returning NULL in case of error (and I suspect that it was the initial expectation)

```
while (1) {
	/* Lock the device to enable exclusive access to virtqueues */
	rp_hdr = rpmsg_virtio_get_tx_buffer(rvdev, len, &idx);
	if (rp_hdr || !tick_count)
		break;
	/*
	 * Try to use wait loop implemented in the virtio dispatcher and
	 * use metal_sleep_usec() method by default.
	 */	status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
	if (status == RPMSG_EOPNOTSUPP) {
		metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
		tick_count--;
		} else if (status != RPMSG_SUCCESS) {
			break;
		}
}
```

wyr-7: @arnopo Yes, but I think we should return back to the `metal_sleep_usec()` when `rpmsg_virtio_notify_wait()` return other error number (not RPMSG_EOPNOTSUPP) to make the rpmsg user can get the tx buffer as mush as possible.

arnopo: I can not understand why we should call `metal_sleep_usec()`. `rpmsg_virtio_notify_wait `is an alternative of `metal_sleep_usec`. either user customizes the wait or they use legacy one based on metal_sleep_usec . I can not see  a reason to use one and then the other. as this will just introduce extra delay on rpmsg_virtio_notify_wait  failure.
 
for me `rpmsg_virtio_notify_wait  `should return either a success on notification received or an error on timeout. 
- in case of success, loop to get a buffer
- in case of failure we leave the function returning NULL value,

Do you see a reason to loop again on failure?

CV-Bowen: @arnopo The wait is interrupted by a signal in `rpmsg_virtio_notify_wait()` is a case. I mean the error is not necessarily a timeout error, it depends on the specific implementation.

arnopo: > @arnopo The wait is interrupted by a signal in `rpmsg_virtio_notify_wait()` is a case. I mean the error is not necessarily a timeout error, it depends on the specific implementation.

Please, list the error cases you are facing that justify that we stay in the loop to recover it. because I cannot see an error that justify it.

CV-Bowen: @arnopo
OK, in our case, we want to continue to process the remain rx buffers in rx virtqueue when can not get the tx buffer in the `rpmsg_virtio_rx_callback()` thread.
This is the example code:
```c
static int rptun_notify_wait(FAR struct rpmsg_device *rdev, uint32_t id)
{
  FAR struct rptun_priv_s *priv = (FAR struct rptun_priv_s *)
    metal_container_of(rdev, struct rpmsg_s, rdev);

  if (current thread PID != rpsmg virtio rx callback thread PID)
    {
      /* Do not allow to continue to process the rx buffers recursively if
       * get tx payload buffer thread is not the rpmsg virtio rx callback thread.
      */

      return -EAGAIN;
    }

  /* Wait new tx buffer */

  nxsem_tickwait(&priv->semtx, MSEC2TICK(RPTUN_TIMEOUT_MS));

  /* Process the remain rx buffers in rx virtquues recursively  */

  remoteproc_get_notification(xxx);

  return 0;
}
```
So we may return `-EAGAIN`, but maybe we can change this value to `RPMSG_EOPNOTSUPP` to work around this issue. But I don't think it's the best way.

arnopo: > @arnopo OK, in our case, we want to continue to process the remain rx buffers in rx virtqueue when can not get the tx buffer in the `rpmsg_virtio_rx_callback()` thread. This is the example code:
> 
> ```c
> static int rptun_notify_wait(FAR struct rpmsg_device *rdev, uint32_t id)
> {
>   FAR struct rptun_priv_s *priv = (FAR struct rptun_priv_s *)
>     metal_container_of(rdev, struct rpmsg_s, rdev);
> 
>   if (current thread PID != rpsmg virtio rx callback thread PID)
>     {
>       /* Do not allow to continue to process the rx buffers recursively if
>        * get tx payload buffer thread is not the rpmsg virtio rx callback thread.
>       */
> 
>       return -EAGAIN;
>     }
> 
>   /* Wait new tx buffer */
> 
>   nxsem_tickwait(&priv->semtx, MSEC2TICK(RPTUN_TIMEOUT_MS));
> 
>   /* Process the remain rx buffers in rx virtquues recursively  */
> 
>   remoteproc_get_notification(xxx);
> 
>   return 0;
> }
> ```
> 
> So we may return `-EAGAIN`, but maybe we can change this value to `RPMSG_EOPNOTSUPP` to work around this issue. But I don't think it's the best way.

Thanks for the details.

If I well understood your issue is that you can have reentrance as you would like to treat new RX message while you are waiting for a TX buffer for a previous one.

Something that I still not understand is why do you need to enter in the loop that call metal_sleep_usec() for that? 
for me this look like a workaround to fix something else.

If your thread  need to sleep t, what about calling nxsig_usleep and returning 0 instead of returning -EAGAIN ?

CV-Bowen: @arnopo If the thread called `rpmsg_get_tx_payload_buffer(wait = true)` is not the rx thread (that call rpmsg_virtio_rx_callback()), and there is no tx buffer.
`rptun_notify_wait()` will return `-EAGAIN` and `rpmsg_get_tx_payload_buffer(wait = true)` will return NULL immediately with the old code, but we actually want to wait until the tx buffer returned by remote or timeout (15s).

> If your thread need to sleep t, what about calling nxsig_usleep and returning 0 instead of returning -EAGAIN ?

Yes, we can return `RPMSG_EOPNOTSUPP` instead `-EAGAIN` to work around this issue, but I think `RPMSG_EOPNOTSUPP` means the `notify_wait()` callback is not implemented but we actually have implemented it. This is why I think return `RPMSG_EOPNOTSUPP` in `rptun_notify_wait()` is not the best way to solve this issue.

arnopo: That means we have something not well designed here. I prefer that we take the time to develop a proper fix rather than merge something that addresses your use case but is not clean.


From my perspective we have to fix the `rpmsg_virtio_notify_wait` Apis. 

we should pass a timeout value as parameter
```c
	while (1) {
		/* Lock the device to enable exclusive access to virtqueues */
		rp_hdr = rpmsg_virtio_get_tx_buffer(rvdev, len, &idx);
		if (rp_hdr || !tick_count)
			break;

		/*
		 * Try to use wait loop implemented in the virtio dispatcher and
		 * use metal_sleep_usec() method by default.
		 */
-		status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
+		status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq, **RPMSG_TICKS_PER_INTERVAL**);
		if (status == RPMSG_EOPNOTSUPP) {
			metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
			tick_count--;
		} else if (status != RPMSG_SUCCESS) {
			break;
		}
	}

````
I think it is too late to fix the APIs for this release. If you are okay with returning `-RPMSG_EOPNOTSUPP `instead of `-EAGAIN `as a workaround for this release, I propose we go with this workaround and integrate the following fix for this release:

```
	while (1) {
		/* Lock the device to enable exclusive access to virtqueues */
		rp_hdr = rpmsg_virtio_get_tx_buffer(rvdev, len, &idx);
		if (rp_hdr || !tick_count)
			break;

		/*
		 * Try to use wait loop implemented in the virtio dispatcher and
		 * use metal_sleep_usec() method by default.
		 */
		status = rpmsg_virtio_notify_wait(rvdev, rvdev->rvq);
		if (status == RPMSG_EOPNOTSUPP) {
			metal_sleep_usec(RPMSG_TICKS_PER_INTERVAL);
			tick_count--;
-		} else if (status == RPMSG_SUCCESS) {
+		} else if (status != RPMSG_SUCCESS) {
			break;
		}
	}

````

@edmooring , @tnmysh any opinion on that?

CV-Bowen: @arnopo OK, it's good for me, but could you help to update this PR because it's already very late in China and @wyr-7 has already rested. Or I can create a new PR to do this thing?
What do you think?

CV-Bowen: https://github.com/OpenAMP/open-amp/pull/624 @arnopo I have create a new PR and it's up to you to decide whether to use the new PR or update the current PR, Thansks.

wyr-7:
@arnopo @CV-Bowen Thanks,  I have updated the latest discussion results to this PR.

wyr-7:
@arnopo  There are some errors when running CI, can you help to check it out?
`WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
Collecting cmake
  Downloading cmake-3.30.5-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (6.4 kB)
Downloading cmake-3.30.5-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (26.9 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 26.9/26.9 MB 278.6 MB/s eta 0:00:00
Installing collected packages: cmake
Successfully installed cmake-3.30.5
SDK URL=
error: Blank SDK download URL`

wyr-7:
> Yes something goes wrong with this test, just ignore it for the moment as it is relying to the Zephyr main branch that can not be stage.

Ok, thanks.

arnopo:
I mergerd PR#624.

@CV-Bowen @wyr-7  you can update this one to update rpmsg_virtio_notify_wait and other API to better meet you needs

arnopo (changes requested):
need to be udated according to https://github.com/OpenAMP/open-amp/pull/614#discussion_r1806763997

PR #615, opened 2024-09-19
rpmsg: Allow to send virtqueue_kick only when RX queue is empty

wyr-7:
Add VQ_RX_EMPTY_NOTIFY config to define the behavior. If VQ_RX_EMPTY_NOTIFY is disabled, notify the other side each time a buffer is released. If VQ_RX_EMPTY_NOTIFY is enabled, only send one notification when the RX queue is empty to improve performance.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -80,6 +80,14 @@ if (WITH_VIRTIO_MMIO_DRV)
   add_definitions(-DWITH_VIRTIO_MMIO_DRV)
 endif (WITH_VIRTIO_MMIO_DRV)
 
+option (WITH_VQ_RX_EMPTY_NOTIFY "Build with virtqueue rx empty notify enabled" OFF)
+
+if (NOT WITH_VQ_RX_EMPTY_NOTIFY)
+  add_definitions(-DVQ_RX_EMPTY_NOTIFY=0)
+else (NOT WITH_VQ_RX_EMPTY_NOTIFY)
+  add_definitions(-DVQ_RX_EMPTY_NOTIFY=1)
+endif (NOT WITH_VQ_RX_EMPTY_NOTIFY)
+
 option (WITH_DCACHE "Build with all cache operations enabled" OFF)
 
 if (WITH_DCACHE)
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -321,8 +321,6 @@ static bool rpmsg_virtio_release_rx_buffer_nolock(struct rpmsg_virtio_device *rv
 	/* Return buffer on virtqueue. */
 	len = virtqueue_get_buffer_length(rvdev->rvq, idx);
 	rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
-	/* Tell peer we returned an rx buffer */
-	virtqueue_kick(rvdev->rvq);
 
 	return true;
 }
@@ -337,8 +335,11 @@ static void rpmsg_virtio_release_rx_buffer(struct rpmsg_device *rdev,
 	rp_hdr = RPMSG_LOCATE_HDR(rxbuf);
 
 	metal_mutex_acquire(&rdev->lock);
-	if (rpmsg_virtio_buf_held_dec_test(rp_hdr))
+	if (rpmsg_virtio_buf_held_dec_test(rp_hdr)) {
 		rpmsg_virtio_release_rx_buffer_nolock(rvdev, rp_hdr);
+		/* Tell peer we returned an rx buffer */
+		virtqueue_kick(rvdev->rvq);
+	}
 	metal_mutex_release(&rdev->lock);
 }
 
@@ -560,6 +561,7 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 	struct rpmsg_device *rdev = &rvdev->rdev;
 	struct rpmsg_endpoint *ept;
 	struct rpmsg_hdr *rp_hdr;
+	bool release = false;
 	uint32_t len;
 	uint16_t idx;
 	int status;
@@ -568,16 +570,19 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		/* Process the received data from remote node */
 		metal_mutex_acquire(&rdev->lock);
 		rp_hdr = rpmsg_virtio_get_rx_buffer(rvdev, &len, &idx);
-		metal_mutex_release(&rdev->lock);
 
 		/* No more filled rx buffers */
-		if (!rp_hdr)
+		if (!rp_hdr) {
+			if (VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY) && release)
+				/* Tell peer we returned some rx buffer */
+				virtqueue_kick(rvdev->rvq);
+			metal_mutex_release(&rdev->lock);
 			break;
+		}
 
 		rp_hdr->reserved = idx;
 
 		/* Get the channel node from the remote device channels list. */
-		metal_mutex_acquire(&rdev->lock);
 		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
 		rpmsg_ept_incref(ept);
 		RPMSG_BUF_HELD_INC(rp_hdr);
@@ -600,8 +605,15 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 
 		metal_mutex_acquire(&rdev->lock);
 		rpmsg_ept_decref(ept);
-		if (rpmsg_virtio_buf_held_dec_test(rp_hdr))
+		if (rpmsg_virtio_buf_held_dec_test(rp_hdr)) {
 			rpmsg_virtio_release_rx_buffer_nolock(rvdev, rp_hdr);
+			if (VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY))
+				/* Kick will be sent only when last buffer is released */
+				release = true;
+			else
+				/* Tell peer we returned an rx buffer */
+				virtqueue_kick(rvdev->rvq);
+		}
 		metal_mutex_release(&rdev->lock);
 	}
 }
--- a/README.md
+++ b/README.md
@@ -86,6 +86,9 @@ library for it project:
   This option can be set to OFF if the only the remote mode is implemented.
 * **WITH_VIRTIO_DEVICE** (default ON): Build with virtio device enabled.
   This option can be set to OFF if the only the driver mode is implemented.
+* **WITH_VQ_RX_EMPTY_NOTIFY** (default OFF): Choose notify mode. When set to
+  ON, only notify when there are no more Message in the RX queue. When set to
+  OFF, notify for each RX buffer released.
 * **WITH_STATIC_LIB** (default ON): Build with a static library.
 * **WITH_SHARED_LIB** (default ON): Build with a shared library.
 * **WITH_ZEPHYR** (default OFF): Build open-amp as a zephyr library. This option


arnopo (commented):
I'm not convinced that this is an improvement in terms of performance. For some use cases, yes; for others, no. 

* In the current implementation, each time we free a buffer, we notify the other side.

  Pros:
  - If the other side is waiting for a free buffer, it does not have to wait for the local side to process all the buffers.

  Cons:
  - One notification per buffer.

* With your PR, there is only one notification when the RX queue is empty.

  Pros:
  - Decreases the number of IRQs on the other side.

  Cons:
  - If the remote side is waiting for a free buffer, it has to wait until all buffers are processed. This is not compatible with some real-time use cases, such as audio streaming.
  -  **my main concern:** It seems that your proposal is not compliant with the VIRTIO protocol, so it cannot be the default behavior. 

Have you tried using VIRTIO_F_EVENT_IDX?

CV-Bowen:
@arnopo Hi, about your concerns:

>   * If the remote side is waiting for a free buffer, it has to wait until all buffers are processed. This is not compatible with some real-time use cases, such as audio streaming.

The rpmsg virtio default implementation in OpenAMP use polling method to wait the tx buffer, so in this case, the real-time performance is not affected and the number of interrupts are reduced.

But if we implemented the `notify_wait_cb` in `struct rpmsg_virtio_device` and make the tx buffer wait method from polling mode to event mode, things would be different and the real-time performance is indeed affected (Only when the buffers is not enough), and also the interrupts number are increased.

Is the default behavior more important?

>   * **my main concern:** It seems that your proposal is not compliant with the VIRTIO protocol, so it cannot be the default behavior.

As you said, this behavior indeed does not conform to the VIRTIO standard, the virtio standard says:
- VirtIO Device --> VirtIO Driver: the device side must send a notification if writes a descriptor index into the used ring (see 2.7.7.2 Device Requirements: Used Buffer Notification Suppression)
- VirtIO Driver --> VirtIO Device: the driver side must send a notification if writes a descriptor index into the available ring (see 2.7.10.1 Driver Requirements: Available Buffer Notification Suppression).

But actually, Linux does the same thing like this PR, please see: https://github.com/torvalds/linux/blob/master/drivers/rpmsg/virtio_rpmsg_bus.c#L778
```c
static void rpmsg_recv_done(struct virtqueue *rvq)
{
	struct virtproc_info *vrp = rvq->vdev->priv;
	struct device *dev = &rvq->vdev->dev;
	struct rpmsg_hdr *msg;
	unsigned int len, msgs_received = 0;
	int err;

	msg = virtqueue_get_buf(rvq, &len);
	if (!msg) {
		dev_err(dev, "uhm, incoming signal, but no used buffer ?\n");
		return;
	}

	while (msg) {
		err = rpmsg_recv_single(vrp, dev, msg, len);
		if (err)
			break;

		msgs_received++;

		msg = virtqueue_get_buf(rvq, &len);
	}

	dev_dbg(dev, "Received %u messages\n", msgs_received);

	/* tell the remote processor we added another available rx buffer */
	if (msgs_received)
		virtqueue_kick(vrp->rvq);
}
```

arnopo:
> @arnopo Hi, about your concerns:
> 
> > * If the remote side is waiting for a free buffer, it has to wait until all buffers are processed. This is not compatible with some real-time use cases, such as audio streaming.
> 
> The rpmsg virtio default implementation in OpenAMP use polling method to wait the tx buffer, so in this case, the real-time performance is not affected and the number of interrupts are reduced.
> 
> But if we implemented the `notify_wait_cb` in `struct rpmsg_virtio_device` and make the tx buffer wait method from polling mode to event mode, things would be different and the real-time performance is indeed affected (Only when the buffers is not enough), and also the interrupts number are increased.
> 
> Is the default behavior more important?

Yes, it is, as it can affect existing projects. If we want to maintain the stability of the library, we have to take legacy into account.

> 
> > * **my main concern:** It seems that your proposal is not compliant with the VIRTIO protocol, so it cannot be the default behavior.
> 
> As you said, this behavior indeed does not conform to the VIRTIO standard, the virtio standard says:
> 
> * VirtIO Device --> VirtIO Driver: the device side must send a notification if writes a descriptor index into the used ring (see 2.7.7.2 Device Requirements: Used Buffer Notification Suppression)
> * VirtIO Driver --> VirtIO Device: the driver side must send a notification if writes a descriptor index into the available ring (see 2.7.10.1 Driver Requirements: Available Buffer Notification Suppression).
> 
> But actually, Linux does the same thing like this PR, please see: https://github.com/torvalds/linux/blob/master/drivers/rpmsg/virtio_rpmsg_bus.c#L778
> 
> ```c
> static void rpmsg_recv_done(struct virtqueue *rvq)
> {
> 	struct virtproc_info *vrp = rvq->vdev->priv;
> 	struct device *dev = &rvq->vdev->dev;
> 	struct rpmsg_hdr *msg;
> 	unsigned int len, msgs_received = 0;
> 	int err;
> 
> 	msg = virtqueue_get_buf(rvq, &len);
> 	if (!msg) {
> 		dev_err(dev, "uhm, incoming signal, but no used buffer ?\n");
> 		return;
> 	}
> 
> 	while (msg) {
> 		err = rpmsg_recv_single(vrp, dev, msg, len);
> 		if (err)
> 			break;
> 
> 		msgs_received++;
> 
> 		msg = virtqueue_get_buf(rvq, &len);
> 	}
> 
> 	dev_dbg(dev, "Received %u messages\n", msgs_received);
> 
> 	/* tell the remote processor we added another available rx buffer */
> 	if (msgs_received)
> 		virtqueue_kick(vrp->rvq);
> }
> ```
rproc virtio/rpmsg virtio in Linux is probably not the best example in terms of virtio specification conformance. :smiley:

That said, your proposal is still valid in terms of optimization. However, I would be in favor of supporting both, even if it would add a new build configuration to test.

@edmooring @tnmysh, any opinions on that?

CV-Bowen:
> rproc virtio/rpmsg virtio in Linux is probably not the best example in terms of virtio specification conformance.

Yes, kick only one time when we need add serveral buffers to the virtqueue is a very easy way to optimize the performance and I found more virtio drivers' implementations in linux do this thing:

virtio_net: https://github.com/torvalds/linux/blob/master/drivers/net/virtio_net.c#L3053
vsock_virtio: https://github.com/torvalds/linux/blob/master/net/vmw_vsock/virtio_transport.c#L155

@arnopo So what's your plan about this? If it's OK, you could give us some advices so that we can continue this PR.

CC @edmooring @tnmysh

edmooring (commented):
I agree with @arnopo at least in part. If we do support this, it should be in addition to the current approach, and it should not be the default. Current users of the library will be surprised if the default behavior changes. 

This might be an optimization for performance in some applications, but it might also be a pessimization (the opposite of optimization) for others because it could affect latency.

arnopo:
> > rproc virtio/rpmsg virtio in Linux is probably not the best example in terms of virtio specification conformance.
> 
> Yes, kick only one time when we need add serveral buffers to the virtqueue is a very easy way to optimize the performance and I found more virtio drivers' implementations in linux do this thing:
> 
> virtio_net: https://github.com/torvalds/linux/blob/master/drivers/net/virtio_net.c#L3053 vsock_virtio: https://github.com/torvalds/linux/blob/master/net/vmw_vsock/virtio_transport.c#L155
> 
> @arnopo So what's your plan about this? If it's OK, you could give us some advices so that we can continue this PR.
> 
> CC @edmooring @tnmysh

Concerning the implementation, one way to do it could be to define an RPMSG_VIRTIO_ENABLED macro similar to VIRTIO_ENABLED and a VQ_RX_EMPTY_NOTIFY build config. Then, use if (RPMSG_VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY)) to define the behavior.

```c
if (RPMSG_VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY))
   /* only notify when no more Message in queue */
else
  /*  notify for each buffer released */
```

CV-Bowen:
@arnopo @edmooring Thanks for the reply and we will update this PR based on your suggestions @arnopo.

wyr-7:
> Concerning the implementation, one way to do it could be to define an RPMSG_VIRTIO_ENABLED macro similar to VIRTIO_ENABLED and a VQ_RX_EMPTY_NOTIFY build config. Then, use if (RPMSG_VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY)) to define the behavior.
> 
> ```c
> if (RPMSG_VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY))
>    /* only notify when no more Message in queue */
> else
>   /*  notify for each buffer released */
> ```

rpmsg_virtio.c already have use VIRTIO_ENABLED，so can we use VIRTIO_ENABLED directly?

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -571,8 +575,15 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		metal_mutex_release(&rdev->lock);
 
 		/* No more filled rx buffers */
-		if (!rp_hdr)
+		if (!rp_hdr) {
+#if VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY)
+			/* Tell peer we returned some rx buffer */
+			if (release) {
+				virtqueue_kick(rvdev->rvq);
```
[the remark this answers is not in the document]
wyr-7: Ok, done.

cmake/options.cmake
```diff
@@ -80,6 +80,14 @@ if (WITH_VIRTIO_MMIO_DRV)
   add_definitions(-DWITH_VIRTIO_MMIO_DRV)
 endif (WITH_VIRTIO_MMIO_DRV)
 
+option (WITH_VQ_RX_EMPTY_NOTIFY "Build with virtqueue rx empty notify enabled" OFF)
+
+if (NOT WITH_VQ_RX_EMPTY_NOTIFY)
```
arnopo, line 85: pease document `WITH_VQ_RX_EMPTY_NOTIFY` in README.md

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -600,8 +610,15 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 
 		metal_mutex_acquire(&rdev->lock);
 		rpmsg_ept_decref(ept);
-		if (rpmsg_virtio_buf_held_dec_test(rp_hdr))
+		if (rpmsg_virtio_buf_held_dec_test(rp_hdr)) {
 			rpmsg_virtio_release_rx_buffer_nolock(rvdev, rp_hdr);
+#if VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY)
+			release = true;
```
arnopo, line 616: add a  comment  such as:
/* kick will be sent only when last buffer is released */

arnopo (changes requested):

arnopo (changes requested):
implementation seems LGTM,  few typos to fix. 

Please also fix commit message, here is a proposal:

**rpmsg: Allow to send virtqueue_kick only when RX queue is empty** 

**A**dd VQ_RX_EMPTY_NOTIFY config to define the behavior. If
VQ_RX_EMPTY_NOTIFY is disabled, notify the other side each
time a buffer is released. **If VQ_RX_EMPTY_NOTIFY is enabled,
only send one notification** when the RX queue is empty to
improve performanc**e.**

wyr-7:
> impementation seemms LGTM, few typos to fix.
> 
> Please also fix commit message, here is a proposal:
> 
> **rpmsg: Allow to send virtqueue_kick only when RX queue is empty**
> 
> **A**dd VQ_RX_EMPTY_NOTIFY config to define the behavior. If VQ_RX_EMPTY_NOTIFY is disabled, notify the other side each time a buffer is released. **If VQ_RX_EMPTY_NOTIFY is enabled, only send one notification** when the RX queue is empty to improve performanc**e.**

Thanks for pointing out the grammatical error, done.

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #616, opened 2024-09-19
virtio: Features change to 64 bit in all virtio_dispatch

wyr-7:
change all 32 bit features to 64 bit for adapt more features

### DIFF
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -177,7 +177,7 @@ static inline uint8_t rpmsg_virtio_get_status(struct rpmsg_virtio_device *rvdev)
  * @return The features supported by both the rpmsg driver and rpmsg device.
  */
 __deprecated
-static inline uint32_t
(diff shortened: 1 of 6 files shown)


arnopo (commented):
regarding virtio spec 64 bits could be not enough. If we update this I would be in favour of having a "features word selection index" as done in virtio mmio transport would be more reliable.
This imply to create new APIs and deprecate exiting ones... 

For instance
```
static inline uint32_t
rpmsg_virtio_get_features_by_idx(struct rpmsg_virtio_device *rvdev, unint32 word_idx)
```

CV-Bowen:
@arnopo Hi, from the virtio spec v1.3, the 64 bits is enough for now and linux use `uint64_t` to represent the feature bit too. Does Linux plan to update to "features word selection index" APIs in the future?

CV-Bowen:
@arnopo Thanks, as you said, it would be better to use "features word selection index" method to support feature bits larger than 64 bits and we will update this PR.

PR #619, opened 2024-09-23
Update CI

arnopo:
Update CI to build the https://github.com/OpenAMP/openamp-system-reference/tree/main/examples/legacy_apps applications instead of the https://github.com/OpenAMP/open-amp/tree/main/apps applications, which will be deprecated in the next release.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -31,6 +31,11 @@ jobs:
       with:
         repository: OpenAMP/libmetal
         path: libmetal
+    - name: Checkout openamp-system-reference
+      uses: actions/checkout@v4
+      with:
+        repository: OpenAMP/openamp-system-reference
+        path: openamp-system-reference
     - name: build for Linux
       id: build_linux
       uses: ./open-amp/.github/actions/build_ci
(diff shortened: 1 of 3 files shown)


tnmysh (approved):
Looks good to me.

.github/workflows/continuous-integration.yml
```diff
@@ -23,12 +23,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout open-amp
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
```
wmamills, line 26: "path: open-amp" HERE ???
with adjustments in script?

I think libmetal and openamp-system-reference are sub directories of open-amp.  That works but is a little strange.  Is this easy to fix as described above??  If not we will deal with it later.

wmamills (changes requested):
In general looks good, some nits you should fix.
And then decide to tackle the dir structure or not

.github/workflows/continuous-integration.yml
wmamills: I take this comment back.  Every build would need to be updated not just the linux build.
We should do this in another PR

wmamills (approved):
I reviewed this and tested it as well.
I think it is ready.

PR #622, opened 2024-10-08
virtio.h: add some user-friendly apis

CV-Bowen:
-    virtio.h: add new api virtio_has_feature()
```c
virtio_has_feature() can be easily used to heck if the virtio device
support a specific feature.

And assgin feature to vdev->feature for virtio device role when get
features, so the virtio device side can use virtio_has_featrue() to
check weather the virtio device support a feature.
```

-    virtio.h: add new feature bit VIRTIO_F_ANY_LAYOUT
```c
Follow the virtio spec, this feature bit indicates that the device
accepts arbitrary descriptor layouts.
```

### DIFF
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -449,6 +449,9 @@ static inline int virtio_get_features(struct virtio_device *vdev,
 		return -ENXIO;
 
 	*features = vdev->func->get_features(vdev);
+	if (VIRTIO_ROLE_IS_DEVICE(vdev))
+		vdev->features = *features;
+
 	return 0;
 }
 
@@ -563,6 +566,28 @@ static inline int virtio_free_buf(struct virtio_device *vdev, void *buf)
 	vdev->mmops->free(vdev, buf);
 
 	return 0;
+}
+
+/**
+ * @brief Check if the virtio device support a specific feature.
+ *
+ * @param vdev		Pointer to device structure.
+ * @param feature_bit	Feature bit to check.
+ *
+ * @return true if the feature is supported, otherwise false.
+ */
+static inline bool virtio_has_feature(struct virtio_device *vdev,
+				      unsigned int feature_bit)
+{
+	uint32_t features;
+
+	if (!vdev && feature_bit >= sizeof(features) * 8)
+		return false;
+
+	if (!vdev->features)
+		virtio_get_features(vdev, &features);
+
+	return (vdev->features & (1UL << feature_bit)) != 0;
 }
 
 #if defined __cplusplus
--- a/lib/include/openamp/virtio.h
+++ b/lib/include/openamp/virtio.h
@@ -119,6 +119,12 @@ struct virtio_device_id {
 #define VIRTIO_F_NOTIFY_ON_EMPTY (1 << 24)
 
 /*
+ * This feature indicates that the device accepts arbitrary
+ * descriptor layouts.
+ */
+#define VIRTIO_F_ANY_LAYOUT (1 << 27)
+
+/*
  * The guest should never negotiate this feature; it
  * is used to detect faulty drivers.
  */


xiaoxiang781216 (approved):

CV-Bowen:
@arnopo Hi, I meet a CI issue:
```c
#define virtio_read_config_member(vdev, structname, member, _ptr) \
	virtio_read_config((vdev), metal_offset_of(structname, member), \
			   _ptr, sizeof(*(_ptr)))
```
This code reports `MACRO_ARG_REUSE: Macro argument reuse '_ptr' - possible side-effects?`
So can we use `typeof` to avoid this issue?

lib/include/openamp/virtio.h
```diff
@@ -503,6 +503,23 @@ static inline int virtio_reset_device(struct virtio_device *vdev)
...
+ * @return true if the feature is supported, otherwise false.
+ */
+static inline bool virtio_has_feature(struct virtio_device *vdev,
+				      unsigned int feature_bit)
+{
+	if (!vdev)
+		return false;
+
+	return (vdev->features & (1UL << feature_bit)) != 0;
+}
```
arnopo, line 535: This function return  valid value only if virtio_negotiate_features has been called first . That also means that  it does not work for  virtio device role.

lib/include/openamp/virtio.h
```diff
@@ -138,6 +138,20 @@ struct virtio_device_id {
...
+	virtio_read_config((vdev), metal_offset_of(structname, member), \
+			   _ptr, sizeof(*(_ptr)))
+
+/*
+ * Write the virtio device configuration member.
+ */
+#define virtio_write_config_member(vdev, structname, member, _ptr) \
+	virtio_write_config((vdev), metal_offset_of(structname, member), \
+			    _ptr, sizeof(*(_ptr)))
+
```
arnopo, line 154: Declaring a local variable in the macro should solve the build issue

```
#define virtio_write_config_member(vdev, structname, member, _ptr)  do { \
	int ptr = (_ptr); \
	virtio_write_config((vdev), metal_offset_of(structname, member), \
			    ptr, sizeof(*(ptr)));
	} while (0)
```

lib/include/openamp/virtio.h
```diff
@@ -132,6 +138,20 @@ struct virtio_device_id {
 #define VIRTIO_TRANSPORT_F_START      28
 #define VIRTIO_TRANSPORT_F_END        32
 
+/*
+ * Read the virtio device configuration member.
+ */
+#define virtio_read_config_member(vdev, structname, member, _ptr)
```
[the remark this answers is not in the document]
CV-Bowen: NuttX is a RTOS that we used, and we have implemented many virtio drivers in it, virtio-blk is one for them and use this API:
https://github.com/apache/nuttx/blob/master/drivers/virtio/virtio-blk.c#L584
Now these APIs are placed in NuttX but we think them should be moved to OpenAMP.

lib/include/openamp/virtio.h
CV-Bowen: Because I need use the sizeof(member), i find a new way, please review again.

lib/include/openamp/virtio.h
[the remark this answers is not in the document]
glneo: Thanks for the info. I'm trying to make sure we do not needlessly grow the OpenAMP API.

`virtio_has_feature()` seems like a reasonable addition to me as it abstracts access to the contents of the vdev struct (opaque structures with accessors is good design).

'virtio_{read,write}_config_member()' on the other hand looks like a thin wrapper around `metal_offset_of()`. Accessing structures custom to the transport layer through the virtio layer API feels like a layering issue to begin with, so I want to make sure we *really* want to keep building more API's on top of that..

lib/include/openamp/virtio.h
arnopo: > Because I need use the sizeof(member), i find a new way, but have same issue :(.

Perhaps this could work:

```
#define virtio_write_config_member(vdev, structname, member, ptr) do { \
    __typeof__(structname) __tmp_structname = (structname); \
    __typeof__(member) __tmp_member = (member); \
    virtio_write_config(vdev, metal_offset_of(__tmp_structname, __tmp_member), \
                        ptr, sizeof(((__tmp_structname *)0)->__tmp_member)); \
} while (0)
```
but is it work for every compiler...?

@glneo  concerns seems valid. This macro introduces an additional API, but it may not offer significant benefits to the library.

CV-Bowen:
@glneo @arnopo Please review again.

lib/include/openamp/virtio.h
CV-Bowen: Done, @arnopo Could you review again?

arnopo: > Yes, I consider assign back the final features to `vdev->features` when call `virtio_get_feature()` for VIRTIO_DEVICE side, so VIRTIO_DEVICE can use this `virtio_has_feature()` after called `virtio_get_feature()`. What do you think?

This does not prevent the use case where `virtio_has_feature` is called before `virtio_get_features`.
to be safe `virtio_has_feature` should call `virtio_get_features` if `vdev->features == NULL`

arnopo (changes requested):

CV-Bowen:
@arnopo Done, please review again.

lib/include/openamp/virtio.h
```diff
@@ -565,6 +568,28 @@ static inline int virtio_free_buf(struct virtio_device *vdev, void *buf)
...
+ * @param feature_bit	Feature bit to check.
+ *
+ * @return true if the feature is supported, otherwise false.
+ */
+static inline bool virtio_has_feature(struct virtio_device *vdev,
+				      unsigned int feature_bit)
+{
+	uint32_t features;
+
+	if (!vdev)
```
arnopo, line 584: We should also check the feature_bit value. For time being we support only up to 32 features bit
so we should check that ` feature_bit <  sizeof(features)`

arnopo (changes requested):
My apologies for this late review. Don't hesitate to ping us if you don't see activity after 2 or 3 weeks.

In addition of my comment could you please rework your commit title 
replacing `virtio.h` by `lib: virtio`

lib/include/openamp/virtio.h
CV-Bowen: Done

CV-Bowen:
@arnopo Thanks for your suggestions. PR has been updated and please review again.

arnopo:
@edmooring @tnmysh: please could you review this one

edmooring (approved):
Looks good to go.

Issue #625, opened 2024-10-21
remoteproc/rpmsg: Wrong argument to safe_strcpy()


jonny-svaerd-arm:
In commit c3132d0 and 27bec14 the `strncpy` function is replaced by a new `safe_strcpy` function, however the last argument to some of the calls are made with `sizeof(name)` where `name` is a `const char *` (resulting in pointer size).

I'm guessing these most likely were intended to be `strlen(name)`?

Thanks!

arnopo:
Good catch!  what i cannot understand  is why the CI does not detect it... 
i will investigate and send a fix

arnopo:
@jonny-svaerd-arm: Please,  could you have a look to the fix?

jonny-svaerd-arm:
Looks good to me @arnopo. Thanks for fixing it so quickly, much appreciated.

arnopo:
> Looks good to me @arnopo. Thanks for fixing it so quickly, much appreciated.

Thank you for reporting the issue! t was good timing as we are starting tests for the upcoming release.

PR #626, opened 2024-10-21
Fix safe_strcpy calls

arnopo:
Fix issue #625 reported by @jonny-svaerd-arm 

Fix `safe_strcpy()`  argument that provides the size of the source string. Indeed, if the source string is of type char*, it is not valid to use `sizeof` to determine its length. Use `strnlen` to get the size of the string while ensuring that we have a maximum size.

On the contrary, it is preferable to use sizeof on `ept->name`, that is an array of char

### DIFF
--- a/lib/include/openamp/remoteproc.h
+++ b/lib/include/openamp/remoteproc.h
@@ -576,7 +576,7 @@ int remoteproc_remove(struct remoteproc *rproc);
  * @brief Initialize remoteproc memory
  *
  * @param mem	Pointer to remoteproc memory
- * @param name	Memory name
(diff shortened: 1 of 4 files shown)


tnmysh (approved):

arnopo:
The `strnlen` function is not supported by all compilers.  I have replaced, for instance,  `strnlen(name, RPROC_MAX_NAME_LEN))` by `RPROC_MAX_NAME_LEN` and added the  (already existing) limitation in the API documentation.

tnmysh (approved):

edmooring (approved):
Looks good to go.

PR #627, opened 2024-10-21
options: make WITH_APPS cmake option deprecated

arnopo:
The applications are now in the openamp-system-reference repository. The "apps" folder is now deprecated and should be removed in coming releases.


tnmysh (approved):

edmooring (approved):
Looks good to go.

PR #630, opened 2024-11-04
rpmsg_virtio: fix the rpmsg_get_tx/rx_buffer() api

CV-Bowen:
user tx/rx buffer size = orignal tx/rx buffer_size - sizeof(rpmsg_hdr)

And use buffer size in config to calculate the user buffer size instead the buffer size from virtqueue, because there may be no buffer in virtqueue.


arnopo:
> And use buffer size in config to calculate the user buffer size instead the buffer size from virtqueue, because there may be no buffer in virtqueue.

Please clarify your need because you are breaking the expected behavior: returning 0 if no buffer is available.

PR #631, opened 2024-11-04
virtqueue: add virtqueue_get_next_avail_buffer() API

CV-Bowen:
In virtio device side, we always need to get the next avaiable buffer based on current buffer index. So add this API for convinience use.

For example, virtio blk driver origanize the buffer: +----------+
| Reqeust  | (Flags: Read | Next)
+----------+
| Buffer   | (Flags: Read/Write | Next)
+----------+
| Response | (Flags: Write)
+----------+

For the virtio blk device size, we need get the Buffer and Response buffer based on the Request buffer index.


CV-Bowen:
@arnopo Yes, this is what we want to do and I have updated this PR to add a new API: `virtqueue_get_available_buffers()` to get all the buffers. Please take a look again.

lib/virtio/virtqueue.c
```diff
@@ -231,6 +231,63 @@ void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
 	return buffer;
 }
 
+void *virtqueue_get_next_avail_buffer(struct virtqueue *vq, uint16_t idx,
```
arnopo, line 234: do you still need this one? virtqueue_get_available_buffers should replace this function

lib/virtio/virtqueue.c
```diff
@@ -231,6 +231,63 @@ void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
...
+	VRING_INVALIDATE(vq->vq_ring.desc[next], sizeof(struct vring_desc));
+	buffer = virtqueue_phys_to_virt(vq, vq->vq_ring.desc[next].addr);
+	if (next_len)
+		*next_len = vq->vq_ring.desc[next].len;
+
+	return buffer;
+}
+
+int virtqueue_get_available_buffers(struct virtqueue *vq,
+				    struct virtqueue_buf *vb, int vbsize,
```
arnopo, line 257: as mentioned in my previous comment you should return  a struct vq_buf that contain the address, the size and the index of the needed buffers instead of just returning the buffer

arnopo (changes requested):

lib/virtio/virtqueue.c
CV-Bowen: Done

lib/virtio/virtqueue.c
CV-Bowen: Done

lib/virtio/virtqueue.c
```diff
@@ -697,3 +733,25 @@ static int virtqueue_navail(struct virtqueue *vq)
...
+	if (!(vq->vq_ring.desc[idx].flags & VRING_DESC_F_NEXT))
+		return NULL;
+
+	next = vq->vq_ring.desc[idx].next;
+	if (next_idx)
+		*next_idx = next;
+
+	VRING_INVALIDATE(vq->vq_ring.desc[next], sizeof(struct vring_desc));
+	buffer = virtqueue_phys_to_virt(vq, vq->vq_ring.desc[next].addr);
+	if (next_len)
```
arnopo, line 753: Instead, I would add a test on next_len to return an error if it is null. Having a buffer without its length seems dangerous to me.

arnopo (changes requested):

lib/virtio/virtqueue.c
CV-Bowen: I add a VQ_ASSERT to this condition

CV-Bowen:
@arnopo Sorry, It's been a long time, I have been busy with internal projects during this period. I have update thie PR according to your comments. Please review again.

lib/include/openamp/virtqueue.h
```diff
@@ -289,8 +289,23 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx);
...
+ * @internal
+ *
+ * @brief Returns the next buffer available for use in the VirtIO queue
+ *
+ * @param vq		Pointer to VirtIO queue control block
+ * @param idx		Index used in vring desc table
+ * @param next_idx	Pointer to the index of the next buffer
+ * @param next_len	Pointer to the length of the next buffer
+ *
+ * @return Pointer to next available buffer of the desc[idx] described
```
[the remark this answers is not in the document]
CV-Bowen: > * we return NULL in case of error
Done
> * s/Pointer/pointer/
Why need change `Pointer` to `pointer`, I see other comments also use `Pointer` ?

lib/include/openamp/virtqueue.h
```diff
@@ -289,8 +289,23 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx);
...
  */
-void *virtqueue_get_available_buffer(struct virtqueue *vq, uint16_t *avail_idx,
-				     uint32_t *len);
+void *virtqueue_get_first_avail_buffer(struct virtqueue *vq, uint16_t *avail_idx,
+				       uint32_t *len);
+
+/**
+ * @internal
+ *
+ * @brief Returns the next buffer available for use in the VirtIO queue
```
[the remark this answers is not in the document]
CV-Bowen: Done

lib/include/openamp/virtqueue.h
```diff
@@ -289,8 +289,23 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx);
...
+ *
+ * @param vq		Pointer to VirtIO queue control block
+ * @param idx		Index used in vring desc table
+ * @param next_idx	Pointer to the index of the next buffer
```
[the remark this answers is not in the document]
CV-Bowen: Done

lib/include/openamp/virtqueue.h
```diff
@@ -289,8 +289,33 @@ void *virtqueue_get_buffer(struct virtqueue *vq, uint32_t *len, uint16_t *idx);
...
+/**
+ * @internal
+ *
+ * @brief Returns the next buffer available for use in the VirtIO queue
+ *
+ * This API retrieves the next available buffer pointer, descriptor table
+ * index, and buffer length from the available buffer chain. The user must
+ * provide the current descriptor table index of the active available
+ * buffer. If the next_idx parameter is not provided, the function returns
+ * NULL and the next_len parameter is optional.
```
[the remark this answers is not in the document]
CV-Bowen: Done

edmooring (approved):
Looks good to go.

PR #632, opened 2024-12-20
rpmsg/rpmsg_virtio: support set trace callback to trace buffer status

luckyyaojin:
support set trace callback to trace buffer status

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -139,5 +139,10 @@ if (DEFINED RPMSG_BUFFER_SIZE)
 endif (DEFINED RPMSG_BUFFER_SIZE)
 
 option (WITH_DOC "Build with documentation" OFF)
+option (WITH_RPMSG_TRACE "Enable Rpmsg Device Trace" OFF)
+
+if (WITH_RPMSG_TRACE)
+  add_definitions(-DRPMSG_TRACE)
+endif (WITH_RPMSG_TRACE)
 
 message ("-- C_FLAGS : ${CMAKE_C_FLAGS}")
--- a/lib/include/openamp/rpmsg.h
+++ b/lib/include/openamp/rpmsg.h
@@ -129,6 +129,19 @@ struct rpmsg_device_ops {
 	int (*get_tx_buffer_size)(struct rpmsg_device *rdev);
 };
 
+/**
+ * struct rpmsg_device_trace - trace operation for debugging
+ *
+ * This structure is used by the rpmsg device to trace the various events.
+ *
+ * @get_tx_buffer: trace callback, called when get a tx buffer
+ * @release_tx_buffer: trace callback, called when release a tx buffer
+ */
+struct rpmsg_device_trace {
+	void (*get_tx_buffer)(struct rpmsg_device *rvdev, void *hdr);
+	void (*release_tx_buffer)(struct rpmsg_device *rvdev, void *hdr);
+};
+
 /** @brief Representation of a RPMsg device */
 struct rpmsg_device {
 	/** List of endpoints */
@@ -152,6 +165,8 @@ struct rpmsg_device {
 
 	/** RPMsg device operations */
 	struct rpmsg_device_ops ops;
+
+	struct rpmsg_device_trace trace;
 
 	/** Create/destroy namespace message */
 	bool support_ns;
--- a/lib/rpmsg/rpmsg_internal.h
+++ b/lib/rpmsg/rpmsg_internal.h
@@ -133,6 +133,47 @@ void rpmsg_ept_incref(struct rpmsg_endpoint *ept);
  */
 void rpmsg_ept_decref(struct rpmsg_endpoint *ept);
 
+#ifdef RPMSG_TRACE
+/**
+ * @internal
+ *
+ * @brief Trace the get tx buffer process
+ *
+ * This function is used to trace the get tx buffer process
+ *
+ * @param rdev	pointer to rpmsg endpoint
+ * @param hdr	pointer to rpmsg header
+ *
+ */
+static inline void
+rpmsg_device_trace_get_tx_buffer(struct rpmsg_device *rdev, void *hdr)
+{
+	if (rdev->trace.get_tx_buffer && hdr)
+		rdev->trace.get_tx_buffer(rdev, hdr);
+}
+
+/**
+ * @internal
+ *
+ * @brief Trace the release tx buffer process
+ *
+ * This function is used to trace the release tx buffer process
+ *
+ * @param rdev	pointer to rpmsg endpoint
+ * @param hdr	pointer to rpmsg header
+ *
+ */
+static inline void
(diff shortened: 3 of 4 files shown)


cmake/options.cmake
```diff
@@ -139,5 +139,10 @@ if (DEFINED RPMSG_BUFFER_SIZE)
 endif (DEFINED RPMSG_BUFFER_SIZE)
 
 option (WITH_DOC "Build with documentation" OFF)
+option (WITH_RPMSG_TRACE "Enable Rpmsg Device Trace" OFF)
+
+if (WITH_RPMSG_TRACE)
```
tnmysh, line 144: Please explain commit mesasge in detail what this option is for.

lib/include/openamp/rpmsg.h
```diff
@@ -129,6 +129,19 @@ struct rpmsg_device_ops {
...
+/**
+ * struct rpmsg_device_trace - trace operation for debugging
+ *
+ * This structure is used by the rpmsg device to trace the various events.
+ *
+ * @get_tx_buffer: trace callback, called when get a tx buffer
+ * @release_tx_buffer: trace callback, called when release a tx buffer
+ */
+struct rpmsg_device_trace {
+	void (*get_tx_buffer)(struct rpmsg_device *rvdev, void *hdr);
```
tnmysh, line 141: What these callbacks are expected to trace ?

cmake/options.cmake
luckyyaojin: Adding this is to track the application and release of rpmsg buffer

lib/include/openamp/rpmsg.h
luckyyaojin: rpmsg_get_tx_payload_buffer and rpmsg_release_tx_buffer

luckyyaojin:
@arnopo hi, please help me how to solve this  ci problem: open-amp Heath Check / Zephyr build from latest on main (pull_request) 

CMake Error at /github/workspace/zephyrproject/zephyr/cmake/modules/boards.cmake:228 (message):
  Invalid BOARD; see above.
Call Stack (most recent call first):
  cmake/modules/sysbuild_default.cmake:15 (include)
  /github/workspace/zephyrproject/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:75 (include)
  /github/workspace/zephyrproject/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  /github/workspace/zephyrproject/zephyr/share/sysbuild-package/cmake/SysbuildConfig.cmake:8 (include)
  template/CMakeLists.txt:10 (find_package)


-- Configuring incomplete, errors occurred!
FATAL ERROR: command exited with status 1: /github/workspace/.venv/bin/cmake -DWEST_PYTHON=/github/workspace/.venv/bin/python3 -B/github/workspace/zephyrproject/zephyr/build -GNinja -DBOARD=lpcxpresso54114_m4 -S/github/workspace/zephyrproject/zephyr/share/sysbuild -DAPP_DIR:PATH=/github/workspace/zephyrproject/zephyr/samples/subsys/ipc/openamp

cmake/options.cmake
tnmysh: Have you checked virtqueue_dump functionality? Is it not enough for above purpose?

luckyyaojin: virtqueue_dump is not enough for record historical information of buffer application and release

arnopo:
> @arnopo hi, please help me to solve this ci problem, think you very much: open-amp Heath Check / Zephyr build from latest on main (pull_request)
> 
> CMake Error at /github/workspace/zephyrproject/zephyr/cmake/modules/boards.cmake:228 (message): Invalid BOARD; see above. Call Stack (most recent call first): cmake/modules/sysbuild_default.cmake:15 (include) /github/workspace/zephyrproject/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:75 (include) /github/workspace/zephyrproject/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate) /github/workspace/zephyrproject/zephyr/share/sysbuild-package/cmake/SysbuildConfig.cmake:8 (include) template/CMakeLists.txt:10 (find_package)
> 
> -- Configuring incomplete, errors occurred! FATAL ERROR: command exited with status 1: /github/workspace/.venv/bin/cmake -DWEST_PYTHON=/github/workspace/.venv/bin/python3 -B/github/workspace/zephyrproject/zephyr/build -GNinja -DBOARD=lpcxpresso54114_m4 -S/github/workspace/zephyrproject/zephyr/share/sysbuild -DAPP_DIR:PATH=/github/workspace/zephyrproject/zephyr/samples/subsys/ipc/openamp

I will have a look , you can ignore it for the moment , only the required checks are mandatory

CV-Bowen:
@arnopo @tnmysh After some thinking, about the trace functions for the rpmsg virtio, I wonder if we can add a new api `metal_trace` in Libmetal, so this api can be used for not only the rpmsg virtio but also the virtio drivers.

CV-Bowen:
@arnopo I mean add a new API named `metal_trace()` or something, because the log system is still different with the trace system in many OS.

CV-Bowen:
> > @arnopo I mean add a new API named `metal_trace()` or something, because the log system is still different with the trace system in many OS. Or your suggestion is adding a new log level to represent the trace message?
> 
> I better understand your request. You would have two types of information: logs and traces.My concern is that these concepts could differ depending on the OS. Please first detail which kind of information you expect to have in traces versus logs. We should have a clear definition of Open-AMP traces and logs as the first step, to be able to discuss how to implement them.

Firstly, in many embedded projects, the log is always output the UART for the convenience of debugging, and the trace information is always output the ram and can be dumped manually or triggered by some special system event (e.g. PANIC). So the log is a always a slow process but trace has small impact on performance.

So I think the log is used to print some infrequent information such as the error information and initialization information.
Instead, trace can be used to profile the performance issue and debug the bug that requires historical sending or receiving information. So I would like to add a trace log every time a message is sent or received. And the trace log should also can be used to record the send and receive data to debug more complicated bug.

tnmysh:
> Firstly, in many embedded projects, the log is always output the UART for the convenience of debugging, and the trace information is always output the ram and can be dumped manually or triggered by some special system event (e.g. PANIC). So the log is a always a slow process but trace has small impact on performance.

open-amp can have a trace buffer type of resource where you can copy all the logs in RAM when you use metal_log API. In Heterogeneous system it's recommended to not throw logs on UART usually because number of UARTs are mostly less than number of cores and possibility of corruption of logs. The recommended way is to use trace buffer and from linux you can get those logs via debugfs or memory dump etc.

> trace can be used to profile the performance

We do have debug level or many other levels of log for this. Besides it's not good idea to keep any logs enabled to measure performance of actual software for long time. We use logs only for notification purpose if performance goes down or varies a lot.

And I agree it varies from project to project, and that is why should not be introduced in a library, but probably some higher level framework.

PR #633, opened 2025-01-02
CI: fix zephyr test on Zephyr for lpcxpresso54114 board

arnopo:
The name of the board has been updated since the Zephyr commit:
  d8cfa6fb2916 ("boards: nxp: convert lpcxpresso54114 to hwmv2")

Fix the build command according to the new name.
The update is compatible with the Zephyr v3.7.


edmooring (approved):
Looks good to go.

PR #636, opened 2025-02-15
docs: update the readthedocs integration for PRs in this repo

wmamills:
Upstream Read The Docs has tightened up the schema checking and this broke what we were doing for PRs.

A good number of things were tried but the best compromise is to:
* supply a dummy sphinx config file in this repo
* override the default build command to use the openamp-docs dir

Also:
* add more documentation about what is going on
* lock to html only format.  PRs only do html anyway but make sure

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,5 +1,16 @@
+# The main readthedocs build is in openamp-docs and includes this repo as a
+# sub-module.
+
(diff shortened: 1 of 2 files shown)


wmamills:
@arnopo this is basically a copy of the libmetal one with only the dir name changed.
It should be OK

arnopo (approved):

PR #637, opened 2025-02-17
cmake: update cmake minimum support to 3.16

arnopo:
We currently support an old version of CMake (3.0.2) that was released in 2018.

This requires adding specific policies to support the evolution to the latest versions of Zephyr.

Update the minimum version to 3.16, aligned with the libmetal.

By updating to CMake 3.16 as the minimum version, we can also clean up the following policy add-ons:
- CMP0048 added in CMake v3.0
- CMP0077 added in CMake v3.13

Associated PR in libmetal library: https://github.com/OpenAMP/libmetal/pull/327

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,4 @@
-cmake_minimum_required (VERSION 3.0.2)
-if (POLICY CMP0048)
-  cmake_policy(SET CMP0048 NEW)
-endif()
-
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

anchao (approved):

edmooring (approved):
Works for me on ARM Linux.

PR #643, opened 2025-04-04
Remove applications and tests that are now available  in openamp-system-reference

arnopo:
As the apps are now available in https://github.com/OpenAMP/openamp-system-reference/tree/main/examples

Remove the apps folder as it is no longer maintained. Clean up associated build and documentation.

We need to decide if we integrate this PR into the upcoming v2025.04 release or if we integrate it just after the release to allow more time to update some external Yocto layers or CIs.


bentheredonethat:
LGTM

tnmysh (approved):

edmooring (approved):
Looks good to go.

edmooring:
@arnopo I would prefer to integrate this change after the 2025.04 release, and put a warning that it is going to happen in the release notes.

arnopo:
> @arnopo I would prefer to integrate this change after the 2025.04 release, and put a warning that it is going to happen in the release notes.

I propose to put this discuss on the agenda for next OpenAMP meeting

wmamills:
This was discusses at the OpenAMP rp call on April 17 and group consensus was to do this now.

PR #647, opened 2025-04-28
rpmsg_virtio.c: wait_remote_ready use metal_sleep_usec to avoid starvation

PetervdPerk-NXP:
`metal_cpu_yield` can either result in an unimplemented function (thus not being executed) or a yield function of the corresponding os..

The yield function however only yields execution to another thread of the same or higher priority. Thus starving any lower priority threads.

`wait_remote_ready` can be blocked for a long time i.e. seconds waiting on the remote core to start communication.

Instead use metal_sleep_usec, allowing lower priority tasks to be executed. While waiting on the remote core.

### DIFF
--- a/lib/remoteproc/remoteproc_virtio.c
+++ b/lib/remoteproc/remoteproc_virtio.c
@@ -15,6 +15,7 @@
 #include <metal/sys.h>
 #include <metal/utilities.h>
 #include <metal/alloc.h>
+#include <metal/sleep.h>
 
 static void rproc_virtio_delete_virtqueues(struct virtio_device *vdev)
 {
@@ -412,6 +413,6 @@ void rproc_virtio_wait_remote_ready(struct virtio_device *vdev)
 		status = rproc_virtio_get_status(vdev);
 		if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK)
 			return;
-		metal_yield();
+		metal_sleep_usec(1000);
 	}
 }
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -272,7 +272,7 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
 		} else if (status & VIRTIO_CONFIG_STATUS_DRIVER_OK) {
 			return 0;
 		}
-		metal_yield();
+		metal_sleep_usec(1000);
 	}
 }
 


arnopo:
Hello @PetervdPerk-NXP 

Could you please provide more details about your issue, such as the operating system and processor you are using? It would also be helpful to reference the `yield `instruction you have tested.

As a preliminary review, it seems that the yield behavior may depend on the platform and operating systems. In such cases, the issue should be addressed in Libmetal.

PetervdPerk-NXP:
Hi @arnopo ,

So for my usecase I'm using Zephyr RTOS and by default `metal_yield` -> `metal_cpu_yield` -> `#define metal_cpu_yield()` thus wouldn't be executed and thus `wait_remote_ready` would simply busy wait.
https://github.com/OpenAMP/libmetal/blob/1c45553b68c747f531cd045c5648d01be165805b/lib/system/zephyr/sys.h#L27
https://github.com/OpenAMP/libmetal/blob/1c45553b68c747f531cd045c5648d01be165805b/lib/processor/generic/cpu.h#L15

I tried to modify [zephyr/sys.h](https://github.com/OpenAMP/libmetal/blob/1c45553b68c747f531cd045c5648d01be165805b/lib/system/zephyr/sys.h#L27) by defining `metal_yield` -> [k_yield](https://docs.zephyrproject.org/apidoc/latest/group__thread__apis.html#ga08a3484c33444ecedc2d71d78495a295). This resolves busy waiting but then the kernel only yields to same or higher priority tasks. An example where this is problematic would be the Zephyr shell/logging task which is the lowest priority thus that wouldn't get any cpu time anymore, but typically any other low priority workqueue wouldn't work either.

So I was thinking since `wait_remote_ready` is a wait function for a remote core why not sleep instead at certain interval `metal_sleep_usec(1000)` which pretty much samples the wait remote core 1 kHz.

Another solution would be [nuttx/sys.h](https://github.com/OpenAMP/libmetal/blob/1c45553b68c747f531cd045c5648d01be165805b/lib/system/nuttx/sys.h#L25) route where they define `metal_yield` -> `metal_sleep_usec(1000)`, but then al occurrence of `metal_yield` would sleep which isn't preferred.

edmooring (approved):
The looks reasonable. The commit message really helps to understand the reasoning behind the change.

arnopo:
Thanks @PetervdPerk-NXP for the clarification

Your proposal seems going in the right direction. It should to work for supported OSes (FreeRTOS, Linux, NuttX, and Zephyr). I just wonder if this could introduce regressions in bare-metal implementations.

I will approve it but i expect more feedback before merge it. 
If merged, I will integrate it when merging the v2025.04 release in Zephyr

arnopo (approved):

iuliana-prodan (approved):
LGTM

xiaoxiang781216 (approved):

xiaoxiang781216:
We made a similar change locally too.

tnmysh (approved):
LGTM.

arnopo:
I merge it, we will revert it for next release if a regression is found.

PR #664, opened 2025-10-14
CI: fix No space left on device message

arnopo:
Add jlumbroso/free-disk-space action to clean-up the workspace before running action.


wmamills (approved):
I agree this is the best option for now

PR #670, opened 2025-10-23
lib: remoteproc: Fix remoteproc_remove_virtio

arnopo:
The code contains redundant checks with both metal_assert(vdev) and if (!vdev).
Moreover, if the assert is disabled, it may lead to dereferencing a null pointer.
We should not rely on asserts for API validation. Instead, replace the assert with an error message.


tnmysh (approved):
LGTM.

PR #682, opened 2026-04-20
Remove unused vars from CMake platform files

glneo:
Remove unused vars from CMake platform files and rename to be more generic as needed.


edmooring (commented):
Looks good to go.

tnmysh (approved):
Thanks for this cleanup.

arnopo:
I need to run a few non-regression tests before merging it, but LGTM.

arnopo (approved):

arnopo:
No blocking point but just a related question before I merge this PR.

I wonder if it still makes sense to keep `cmake/platforms` folder?  There is no more application and  cmake options should inherit from the main project

Does someone is still using this .cmake file in open-amp?

glneo:
@arnopo I was thinking the same thing but didn't want to go that far in this PR. If no one speaks up claiming to use them I'd be happy to just delete them in a follow up PR.

tnmysh:
> No blocking point but just a related question before I merge this PR.
> 
> I wonder if it still makes sense to keep `cmake/platforms` folder? There is no more application and cmake options should inherit from the main project
> 
> Does someone is still using this .cmake file in open-amp?

@arnopo I use these files for a day-to-day native cmake build.

arnopo:
> > No blocking point but just a related question before I merge this PR.
> > I wonder if it still makes sense to keep `cmake/platforms` folder? There is no more application and cmake options should inherit from the main project
> > Does someone is still using this .cmake file in open-amp?
> 
> @arnopo I use these files for a day-to-day native cmake build.

Ok let keep them for the moment.
@tnmysh : For my understanding, is it out of habit that you use it this way, or is there a reason behind it?

tnmysh:
> > > No blocking point but just a related question before I merge this PR.
> > > I wonder if it still makes sense to keep `cmake/platforms` folder? There is no more application and cmake options should inherit from the main project
> > > Does someone is still using this .cmake file in open-amp?
> > 
> > 
> > @arnopo I use these files for a day-to-day native cmake build.
> 
> Ok let keep them for the moment. @tnmysh : For my understanding, is it out of habit that you use it this way, or is there a reason behind it?

Yes, it is legacy build system that I have been using it.
However, we should discuss if we can provide new build system in openamp-system-reference. Ideally, I would like to maintain cmake build files in openamp-system-reference, and re-use them for the libraries.

PR #684, opened 2026-04-29
Enhance rpmsg buf config

tnmysh:
Corresponding Linux kernel patch series: https://lore.kernel.org/all/20251114184640.3020427-1-tanmay.shah@amd.com/

### DIFF
--- a/lib/include/openamp/rpmsg_virtio.h
+++ b/lib/include/openamp/rpmsg_virtio.h
@@ -15,6 +15,7 @@
 #include <metal/io.h>
 #include <metal/mutex.h>
 #include <metal/cache.h>
+#include <metal/compiler.h>
 #include <openamp/rpmsg.h>
 #include <openamp/virtio.h>
 
@@ -29,6 +30,7 @@ extern "C" {
 
 /* The feature bitmap for virtio rpmsg */
 #define VIRTIO_RPMSG_F_NS	0 /* RP supports name service notifications */
+#define VIRTIO_RPMSG_F_BUFSZ	1 /* fw provides tx and rx single buf size */
 
 #if defined(VIRTIO_USE_DCACHE)
 #define BUFFER_FLUSH(x, s)		metal_cache_flush(x, s)
@@ -59,7 +61,14 @@ struct rpmsg_virtio_shm_pool {
  * This structure is used by the RPMsg virtio host to configure the virtiio
  * layer.
  */
+METAL_PACKED_BEGIN
 struct rpmsg_virtio_config {
+	/** version of this struct */
+	uint8_t version;
+
+	/** size of the config space */
+	uint16_t size;
+
 	/** The size of the buffer used to send data from host to remote */
 	uint32_t h2r_buf_size;
 
@@ -68,7 +77,14 @@ struct rpmsg_virtio_config {
 
 	/** The flag for splitting shared memory pool to TX and RX */
 	bool split_shpool;
-};
+} METAL_PACKED_END;
+
+/*
+ * split_shpool is a local OpenAMP option and is not part of the virtio
+ * device configuration space ABI.
+ */
+#define RPMSG_VIRTIO_CONFIG_SIZE \
+	((uint16_t)(sizeof(struct rpmsg_virtio_config) - sizeof(bool)))
 
 /** @brief Representation of a RPMsg device based on virtio */
 struct rpmsg_virtio_device {
--- a/lib/rpmsg/rpmsg_virtio.c
+++ b/lib/rpmsg/rpmsg_virtio.c
@@ -62,6 +62,8 @@ struct vbuff_reclaimer_t {
 #if VIRTIO_ENABLED(VIRTIO_DRIVER_SUPPORT)
 #define RPMSG_VIRTIO_DEFAULT_CONFIG                \
 	(&(const struct rpmsg_virtio_config) {     \
+		.version = 1,			   \
+		.size = RPMSG_VIRTIO_CONFIG_SIZE, \
 		.h2r_buf_size = RPMSG_BUFFER_SIZE, \
 		.r2h_buf_size = RPMSG_BUFFER_SIZE, \
 		.split_shpool = false,             \
@@ -274,6 +276,40 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
 		}
 		metal_sleep_usec(1000);
 	}
+}
+
+static int rpmsg_virtio_read_config(struct rpmsg_virtio_device *rvdev)
+{
+	struct rpmsg_virtio_config config = {0};
+	int ret;
+
+	ret = virtio_read_config(rvdev->vdev, 0, &config,
+				 RPMSG_VIRTIO_CONFIG_SIZE);
+	if (ret)
+		return RPMSG_ERR_DEV_STATE;
+
+	if (config.size < RPMSG_VIRTIO_CONFIG_SIZE ||
+	    !config.h2r_buf_size || !config.r2h_buf_size)
+		return RPMSG_ERR_BUFF_SIZE;
+
+	rvdev->config = config;
+	return RPMSG_SUCCESS;
+}
+
+static int rpmsg_virtio_write_config(struct rpmsg_virtio_device *rvdev)
+{
+	int ret;
+
+	if (rvdev->config.size != RPMSG_VIRTIO_CONFIG_SIZE ||
+	    !rvdev->config.h2r_buf_size || !rvdev->config.r2h_buf_size)
+		return RPMSG_ERR_BUFF_SIZE;
+
+	ret = virtio_write_config(rvdev->vdev, 0, &rvdev->config,
+				  rvdev->config.size);
+	if (ret)
+		return RPMSG_ERR_DEV_STATE;
+
+	return RPMSG_SUCCESS;
 }
 
 /**
@@ -706,7 +742,9 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 int rpmsg_virtio_get_tx_buffer_size(struct rpmsg_device *rdev)
 {
 	struct rpmsg_virtio_device *rvdev;
+	uint32_t features = 0;
 	int size = 0;
+	int ret;
 
 	if (!rdev)
 		return RPMSG_ERR_PARAM;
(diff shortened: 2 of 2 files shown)


tnmysh:
@arnopo 

we can review this one after the release.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -830,6 +846,13 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 	}
 
 	if (VIRTIO_ROLE_IS_DEVICE(vdev)) {
+		status = virtio_get_features(vdev, &features);
+		if (status)
+			return status;
+
+		if (features & (1 << VIRTIO_RPMSG_F_BUFSZ))
+			rvdev->config = *config;
```
arnopo: Here you should use `virtio_read_config` to retrieve the config from the transport layer
Did you have a look to @xiaoxiang781216  work in https://github.com/OpenAMP/open-amp/pull/155 ?
Seems to me the good approach

arnopo (changes requested):

lib/rpmsg/rpmsg_virtio.c
tnmysh: @arnopo done in latest push.

edmooring (approved):
Looks good to go.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -830,6 +850,14 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 	}
 
 	if (VIRTIO_ROLE_IS_DEVICE(vdev)) {
+		status = virtio_get_features(vdev, &features);
+		if (status)
+			return status;
+
+		if (features & (1 << VIRTIO_RPMSG_F_BUFSZ))
+			virtio_write_config(rvdev->vdev, 0, (void *)config,
+					    sizeof(*config));
```
arnopo: The config could be read by the virtio driver before this code is executed.  For instance the Linux probes the rpmsg in parallel of the start of the remoteproc firmware based on the resource table that should contain the config.
For my here we should only read the config if `VIRTIO_RPMSG_F_BUFSZ` feature bit is set else we should rely on *config.  
One question is: should we also mange the config write or should we consider that it is part of the resource table. 

I propose to discuss this in next OpenAMP meeting

nathalie-ckc: 7/1 OpenAMP System Reference call: 

- Arnaud to confirm it is a standard pattern w/ MMIO
- Tanmay to confirm would work to have read instead of write for current use case & update accordingly.  

Allow write only for use case when it is compiled as driver.

arnopo: Extracted from virtio spec 1 .4:

```
2.5.2 Device Requirements: Device Configuration Space
The device MUST allow reading of any device-specific configuration field before FEATURES_OK is set by the driver. This includes fields which are conditional on feature bits, as long as those feature bits are offered by the device.
```
```
3.2 Device Operation
When operating the device, each field in the device configuration space can be changed by either the driver or the device.

Whenever such a configuration change is triggered by the device, driver is notified. This makes it possible for drivers to cache device configuration, avoiding expensive configuration reads unless notified.

```
So it seems that having a static configuration as  first step is quite simple and reliable

1) the device provides a static configuration space (in the resource table).
2) the driver can update it until FEATURES_OK is set
3) the device read the configuration space when DRIVER_OK is set

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -59,7 +61,20 @@ struct rpmsg_virtio_shm_pool {
  * This structure is used by the RPMsg virtio host to configure the virtiio
  * layer.
  */
+METAL_PACKED_BEGIN
 struct rpmsg_virtio_config {
+	/** version of this struct */
```
iuliana-prodan, line 66: This needs an update, right?
Because it doesn't match the Linux v5 header which dropped the reserved fields, so every field after `version` is now at a different offset.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -762,9 +767,24 @@ int rpmsg_virtio_get_rx_buffer_size(struct rpmsg_device *rdev)
 		/*
 		 * If other core is host then buffers are provided by it,
 		 * so get the buffer size from the virtqueue.
+		 * If virtio device has BUFSZ feature, then the rx buffer is
+		 * provided by the vdev config space in the resource table.
 		 */
-		size = (int)virtqueue_get_desc_size(rvdev->rvq) -
-		       sizeof(struct rpmsg_hdr);
+		features = 0;
```
iuliana-prodan: Redundant, is already zero-initialized at declaration.

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -762,9 +767,24 @@ int rpmsg_virtio_get_rx_buffer_size(struct rpmsg_device *rdev)
...
+		ret = virtio_get_features(rvdev->vdev, &features);
+		if (ret) {
+			metal_mutex_release(&rdev->lock);
+			return RPMSG_ERR_DEV_STATE;
+		}
+
+		if (features & (1 << VIRTIO_RPMSG_F_BUFSZ)) {
```
iuliana-prodan: Nitpick: `1U <<` for feature masks (here and in rpmsg_init_vdev_with_config).

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -830,6 +850,14 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 	}
 
 	if (VIRTIO_ROLE_IS_DEVICE(vdev)) {
+		status = virtio_get_features(vdev, &features);
+		if (status)
+			return status;
+
+		if (features & (1 << VIRTIO_RPMSG_F_BUFSZ))
+			virtio_write_config(rvdev->vdev, 0, (void *)config,
```
iuliana-prodan: Return value ignored, the same happens for `virtio_read_config` above.

iuliana-prodan: `sizeof(*config)` writes the full struct including `split_shpool`, inconsistent with the advertised `size` field.

nathalie-ckc:
2026-07-15 System Reference Call:
Tanmay to address Iulia's comments, but will make update to PR once the Linux kernel side is finalized.  Tanmay has sent latest revision in Linux kernel.

lib/include/openamp/rpmsg_virtio.h
tnmysh: Yes. same as mentioned in: https://github.com/OpenAMP/openamp-system-reference/pull/106

lib/rpmsg/rpmsg_virtio.c
tnmysh: Ack.

lib/rpmsg/rpmsg_virtio.c
tnmysh: Ack.

lib/rpmsg/rpmsg_virtio.c
tnmysh: Ack, I will fix in next rev.

tnmysh: Ack, I will fix in next rev.

nathalie-ckc:
8/26/26 System Reference call:
Tanmay updated based on V7 on Linux side. A few small items to address in Mathieu's most recent feedback.  open-amp reviewers can wait until Linux side is complete before reviewing.

nathalie-ckc:
9/8/26: Mathieu merged the Linux part. Tanmay will rebase, then it will be ready for OpenAMP review

tnmysh:
This PR is ready for review. @arnopo

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -276,6 +278,40 @@ static int rpmsg_virtio_wait_remote_ready(struct rpmsg_virtio_device *rvdev)
...
+{
+	struct rpmsg_virtio_config config = {0};
+	int ret;
+
+	ret = virtio_read_config(rvdev->vdev, 0, &config,
+				 RPMSG_VIRTIO_CONFIG_SIZE);
+	if (ret)
+		return RPMSG_ERR_DEV_STATE;
+
+	if (config.size < RPMSG_VIRTIO_CONFIG_SIZE ||
```
arnopo, line 291: Check on the config version seems missing here and in `rpmsg_virtio_write_config`

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -837,9 +901,15 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 	status = virtio_get_features(vdev, &features);
 	if (status)
 		return status;
-	rdev->support_ns = !!(features & (1 << VIRTIO_RPMSG_F_NS));
+	rdev->support_ns = !!(features & (1U << VIRTIO_RPMSG_F_NS));
 
 	if (VIRTIO_ROLE_IS_DRIVER(vdev)) {
+		if (features & (1U << VIRTIO_RPMSG_F_BUFSZ)) {
+			status = rpmsg_virtio_write_config(rvdev);
```
arnopo, line 908: This part seems to me not valid. The driver should only read the config. It must not update it.

it seems that something is also wrong in `rproc_virtio_write_config`

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -837,9 +901,15 @@ int rpmsg_init_vdev_with_config(struct rpmsg_virtio_device *rvdev,
 	status = virtio_get_features(vdev, &features);
 	if (status)
 		return status;
-	rdev->support_ns = !!(features & (1 << VIRTIO_RPMSG_F_NS));
+	rdev->support_ns = !!(features & (1U << VIRTIO_RPMSG_F_NS));
```
arnopo, line 904: Please this in a separate Commit

arnopo (changes requested):

lib/include/openamp/rpmsg_virtio.h
```diff
@@ -59,7 +61,14 @@ struct rpmsg_virtio_shm_pool {
...
+	uint8_t version;
+
+	/** size of the config space */
+	uint16_t size;
+
 	/** The size of the buffer used to send data from host to remote */
 	uint32_t h2r_buf_size;
```
iuliana-prodan: The size fields are in the opposite order from the merged Linux abi https://git.kernel.org/pub/scm/linux/kernel/git/remoteproc/linux.git/tree/include/linux/rpmsg/virtio_rpmsg.h?h=for-next

lib/rpmsg/rpmsg_virtio.c
iuliana-prodan: agree - check `version == 1` like linux does and reject sizes `<= sizeof(struct rpmsg_hdr)`

lib/include/openamp/rpmsg_virtio.h
tnmysh: This is expected, as it's parsed accordingly in the driver: https://git.kernel.org/pub/scm/linux/kernel/git/remoteproc/linux.git/tree/drivers/rpmsg/virtio_rpmsg_bus.c?h=for-next#n846

arnopo: I wonder if it would be more obvious to talk about a driver and a device. What about `drv2dev_buf_size`?

lib/rpmsg/rpmsg_virtio.c
tnmysh: Yeah, this part is redundant. I will remove it.

lib/rpmsg/rpmsg_virtio.c
tnmysh: Ack.

lib/rpmsg/rpmsg_virtio.c
tnmysh: Ack.

lib/include/openamp/rpmsg_virtio.h
tnmysh: @arnopo If we want to rename the structures field, can we do it in a different commit?
This name exists before this change, and irrelevant to the functionality being introduced in the PR.

PR #687, opened 2026-06-01
ELF loader should validate before load

### LINKED ISSUE #688: CVE-2026-37540 report:  integer overflow vulnerability on ELF Loader
- Reported on version: V2025.10.0

-  File  impacted:
 
    - lib/remoteproc/elf_loader.c 

- Vulnerability Analysis from CVE:

    OpenAMP's ELF loader parses firmware images destined for a remote processor. During parsing, the loader reads two 16-bit fields from the ELF header (typically program/section header count and entry size) and multiplies them to compute a buffer size. The product is stored in a 32-bit integer without prior validation.

    When both operands approach their 16-bit maximum, the resulting 32-bit value can wrap. The loader then allocates an undersized buffer while subsequent parsing logic uses the original, unbounded counts to iterate. This mismatch produces an out-of-bounds write during firmware image processing.

    Because OpenAMP runs on Asymmetric Multi-Processing (AMP) systems coordinating between a Cortex-A host and a Cortex-M coprocessor, exploitation can corrupt remoteproc state controlling the secondary core.

See  https://www.sentinelone.com/vulnerability-database/cve-2026-37540/ for details.

bentheredonethat:
OpenAMP’s ELF loader should reject malformed firmware images before using ELF header metadata to size allocations, copy header tables, or look up section names.

  OpenAMP’s remoteproc ELF loader trusts several ELF header and section-header fields while parsing firmware images. These fields control program-header and section-header table sizes, table
  offsets, segment copy ranges, and section string-table lookups.

  The main issue is in the header-table loading path. The loader computes program and section table sizes from firmware-controlled values such as e_phnum * e_phentsize and e_shnum * e_shentsize.
  Those computed sizes are then used for allocation and memcpy(), while later code walks the copied buffers as native Elf32_Phdr, Elf64_Phdr, Elf32_Shdr, or Elf64_Shdr arrays. If the count, entry
  size, or offset metadata is malformed, the allocation/copy size can diverge from the way the loader later indexes the table.

  The range checks around these copies also rely on offset-plus-length arithmetic. With malformed offsets or sizes, unchecked addition can wrap and make an invalid range appear valid. Similar range
  assumptions apply when loading the section string table and when the loader asks the backing image store for the next chunk of firmware data.

  Section-name lookup has a separate parser-side read issue. The loader searches for sections such as .resource_table by using each section header’s sh_name as an offset into the loaded section
  string table. Because sh_name is firmware-controlled, the loader must not form or compare name_table + sh_name unless the offset is within the loaded string table and the referenced string is NUL-
  terminated before the end of that table.

  The fix makes the ELF loader fail closed on malformed metadata:

  - add portable checked arithmetic helpers for size_t multiplication, addition, and image-chunk range validation;
  - validate ELF header shape before loading tables, including ELF class, ELF header size, program-header entry size, section-header entry size, and section string-table index;
  - use checked multiplication for program-header and section-header table sizes;
  - replace wrapping offset-plus-length checks with overflow-safe range validation;
  - validate the section string-table range before copying it;
  - store the loaded section string-table size in the ELF image state;
  - bound section-name lookup by checking sh_name and comparing only within the loaded string table;
  - avoid wrapping arithmetic in the remoteproc caller’s non-seekable offset comparison.

  With these changes, valid firmware images continue to load normally, while malformed images with inconsistent table metadata, overflowing ranges, invalid entry sizes, or out-of-bounds section
  names are rejected before allocation, copy, or table lookup proceeds.

### DIFF
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -124,6 +124,50 @@ static int elf_shstrndx(const void *elf_info)
 	}
 }
 
+/**
+ * @brief Validate that an ELF header matches the supported header layout.
+ *
+ * @param elf_info	ELF information containing a copied ELF header.
+ *
+ * @return 0 on success, otherwise -RPROC_EINVAL.
+ */
+static int elf_validate_header(const void *elf_info)
+{
+	if (elf_is_64(elf_info) != 0) {
+		const Elf64_Ehdr *ehdr = elf_info;
+
+		if (ehdr->e_ident[EI_CLASS] != ELFCLASS64)
+			return -RPROC_EINVAL;
+		if (ehdr->e_ehsize != sizeof(Elf64_Ehdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_phnum != 0 &&
+		    ehdr->e_phentsize != sizeof(Elf64_Phdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_shnum != 0 &&
+		    ehdr->e_shentsize != sizeof(Elf64_Shdr))
+			return -RPROC_EINVAL;
+	} else {
+		const Elf32_Ehdr *ehdr = elf_info;
+
+		if (ehdr->e_ident[EI_CLASS] != ELFCLASS32)
+			return -RPROC_EINVAL;
+		if (ehdr->e_ehsize != sizeof(Elf32_Ehdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_phnum != 0 &&
+		    ehdr->e_phentsize != sizeof(Elf32_Phdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_shnum != 0 &&
+		    ehdr->e_shentsize != sizeof(Elf32_Shdr))
+			return -RPROC_EINVAL;
+	}
+
+	if (elf_shnum(elf_info) != 0 &&
+	    elf_shstrndx(elf_info) >= elf_shnum(elf_info))
+		return -RPROC_EINVAL;
+
+	return 0;
+}
+
 static void **elf_phtable_ptr(void *elf_info)
 {
 	if (elf_is_64(elf_info) == 0) {
@@ -414,14 +458,25 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 			return ELF_STATE_INIT;
 		} else {
 			size_t infosize = elf_info_size(img_data);
+			bool info_allocated = false;
+			int ret;
 
 			if (!*img_info) {
 				*img_info = metal_allocate_memory(infosize);
 				if (!*img_info)
 					return -RPROC_ENOMEM;
 				memset(*img_info, 0, infosize);
+				info_allocated = true;
 			}
 			memcpy(*img_info, img_data, tmpsize);
+			ret = elf_validate_header(*img_info);
+			if (ret < 0) {
+				if (info_allocated) {
+					metal_free_memory(*img_info);
+					*img_info = NULL;
+				}
+				return ret;
+			}
 			load_state = elf_load_state(*img_info);
 			*load_state = ELF_STATE_WAIT_FOR_PHDRS;
 			last_load_state = ELF_STATE_WAIT_FOR_PHDRS;
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -207,6 +207,21 @@ static void **elf_shstrtab_ptr(void *elf_info)
 	}
 }
 
+/**
+ * @brief Store the loaded ELF section string table size.
+ *
+ * @param elf_info	ELF information.
+ * @param size		Loaded section string table size.
+ */
+static void elf_set_shstrtab_size(void *elf_info, size_t size)
+{
+	if (elf_is_64(elf_info) == 0) {
+		((struct elf32_info *)elf_info)->shstrtab_size = size;
+	} else {
+		((struct elf64_info *)elf_info)->shstrtab_size = size;
+	}
+}
+
 static int *elf_load_state(void *elf_info)
 {
 	if (elf_is_64(elf_info) == 0) {
@@ -598,6 +613,7 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 		*shstrtab = metal_allocate_memory(shstrtab_size);
 		if (!*shstrtab)
 			return -RPROC_ENOMEM;
+		elf_set_shstrtab_size(*img_info, shstrtab_size);
 		memcpy(*shstrtab,
 		       (const char *)img_data + shstrtab_offset,
 		       shstrtab_size);
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -5,7 +5,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+#include <stdbool.h>
 #include <string.h>
+#include <internal/utilities.h>
 #include <metal/alloc.h>
 #include <metal/log.h>
 #include <openamp/elf_loader.h>
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -235,6 +235,35 @@ static int *elf_load_state(void *elf_info)
 	}
 }
 
+/**
+ * @brief Compare a section name with a bounded section string table entry.
+ *
+ * @param name		Section name to match.
+ * @param name_table	Loaded section string table.
+ * @param name_table_size	Size of the loaded section string table.
+ * @param sh_name	Offset of the candidate name in the string table.
+ *
+ * @return true if the name matches, otherwise false.
+ */
+static bool elf_section_name_matches(const char *name, const char *name_table,
+				     size_t name_table_size, size_t sh_name)
+{
+	size_t name_len;
+	size_t remaining;
+	const char *candidate;
+
+	if (sh_name >= name_table_size)
+		return false;
+
+	remaining = name_table_size - sh_name;
+	name_len = strlen(name);
+	if (name_len >= remaining)
+		return false;
+
+	candidate = name_table + sh_name;
+	return memcmp(name, candidate, name_len + 1) == 0;
+}
+
 static void elf_parse_segment(void *elf_info, const void *elf_phdr,
 			      unsigned int *p_type, size_t *p_offset,
 			      metal_phys_addr_t *p_vaddr,
@@ -303,6 +332,7 @@ static void *elf_get_section_from_name(void *elf_info, const char *name)
 {
 	unsigned int i;
 	const char *name_table;
+	size_t name_table_size;
 
 	if (elf_is_64(elf_info) == 0) {
 		struct elf32_info *einfo = elf_info;
@@ -310,10 +340,13 @@ static void *elf_get_section_from_name(void *elf_info, const char *name)
 		Elf32_Shdr *shdr = einfo->shdrs;
 
 		name_table = einfo->shstrtab;
+		name_table_size = einfo->shstrtab_size;
 		if (!shdr || !name_table)
 			return NULL;
 		for (i = 0; i < ehdr->e_shnum; i++, shdr++) {
-			if (strcmp(name, name_table + shdr->sh_name))
+			if (!elf_section_name_matches(name, name_table,
+						      name_table_size,
+						      shdr->sh_name))
 				continue;
 			else
 				return shdr;
@@ -324,10 +357,13 @@ static void *elf_get_section_from_name(void *elf_info, const char *name)
 		Elf64_Shdr *shdr = einfo->shdrs;
 
 		name_table = einfo->shstrtab;
+		name_table_size = einfo->shstrtab_size;
 		if (!shdr || !name_table)
 			return NULL;
 		for (i = 0; i < ehdr->e_shnum; i++, shdr++) {
-			if (strcmp(name, name_table + shdr->sh_name))
+			if (!elf_section_name_matches(name, name_table,
+						      name_table_size,
+						      shdr->sh_name))
 				continue;
 			else
 				return shdr;
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -571,6 +571,8 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 		int shstrndx;
 		void *shdr;
 		void **shstrtab;
+		bool range_contained;
+		int ret;
 
 		metal_log(METAL_LOG_DEBUG, "Loading ELF shstrtab.\r\n");
 		shstrndx = elf_shstrndx(*img_info);
@@ -581,8 +583,11 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 				  NULL, &shstrtab_offset,
 				  &shstrtab_size, NULL, NULL,
 				  NULL, NULL);
-		if (offset > shstrtab_offset ||
-		    offset + len < shstrtab_offset + shstrtab_size) {
+		ret = OPENAMP_RANGE_CONTAINS(offset, len, shstrtab_offset,
+					     shstrtab_size, &range_contained);
+		if (ret)
+			return -RPROC_EINVAL;
+		if (!range_contained) {
 			*noffset = shstrtab_offset;
 			*nlen = shstrtab_size;
 			return *load_state;
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -490,14 +490,22 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 	if (*load_state == ELF_STATE_WAIT_FOR_PHDRS) {
 		size_t phdrs_size;
 		size_t phdrs_offset;
+		bool range_contained;
+		int ret;
 		void **phdrs;
 		const void *img_phdrs;
 
 		metal_log(METAL_LOG_DEBUG, "Loading ELF program header.\r\n");
 		phdrs_offset = elf_phoff(*img_info);
-		phdrs_size = elf_phnum(*img_info) * elf_phentsize(*img_info);
-		if (offset > phdrs_offset ||
-		    offset + len < phdrs_offset + phdrs_size) {
+		ret = OPENAMP_MUL_SAFE(elf_phnum(*img_info),
+				       elf_phentsize(*img_info), &phdrs_size);
+		if (ret)
+			return -RPROC_EINVAL;
+		ret = OPENAMP_RANGE_CONTAINS(offset, len, phdrs_offset,
+					     phdrs_size, &range_contained);
+		if (ret)
+			return -RPROC_EINVAL;
+		if (!range_contained) {
 			*noffset = phdrs_offset;
 			*nlen = phdrs_size;
 			return *load_state;
@@ -517,6 +525,8 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 	if ((*load_state & ELF_STATE_WAIT_FOR_SHDRS) != 0) {
 		size_t shdrs_size;
 		size_t shdrs_offset;
+		bool range_contained;
+		int ret;
 		void **shdrs;
 		const void *img_shdrs;
 
@@ -528,9 +538,15 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 			*nlen = 0;
 			return *load_state;
 		}
-		shdrs_size = elf_shnum(*img_info) * elf_shentsize(*img_info);
-		if (offset > shdrs_offset ||
-		    offset + len < shdrs_offset + shdrs_size) {
+		ret = OPENAMP_MUL_SAFE(elf_shnum(*img_info),
+				       elf_shentsize(*img_info), &shdrs_size);
+		if (ret)
+			return -RPROC_EINVAL;
+		ret = OPENAMP_RANGE_CONTAINS(offset, len, shdrs_offset,
+					     shdrs_size, &range_contained);
+		if (ret)
+			return -RPROC_EINVAL;
+		if (!range_contained) {
 			*noffset = shdrs_offset;
 			*nlen = shdrs_size;
 			return *load_state;
--- a/lib/include/internal/utilities.h
+++ b/lib/include/internal/utilities.h
@@ -5,6 +5,8 @@
  *
  */
 
+#include <stdbool.h>
+#include <stdint.h>
 #include <string.h>
 
 /**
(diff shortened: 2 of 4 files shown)


lib/remoteproc/elf_loader.c
```diff
@@ -157,6 +157,43 @@ static int elf_shstrndx(const void *elf_info)
 	}
 }
 
+static int elf_validate_header(const void *elf_info)
+{
+	if (elf_is_64(elf_info) == 0) {
+		const Elf32_Ehdr *ehdr = elf_info;
```
edmooring: I think we can ignore the compliance check failures for the camel case ELF header names. They are defined in the ELF spec this way.

lib/remoteproc/elf_loader.c
```diff
@@ -6,11 +6,46 @@
  */
 
 #include <string.h>
+#include <stdint.h>
 #include <metal/alloc.h>
 #include <metal/log.h>
 #include <openamp/elf_loader.h>
 #include <openamp/remoteproc.h>
 
+static int elf_mul_size(size_t a, size_t b, size_t *out)
```
edmooring: This applies to all the newly defined functions. They need Doxygen comments to describe them.

lib/remoteproc/elf_loader.c
```diff
@@ -6,11 +6,46 @@
...
+static int elf_add_size(size_t a, size_t b, size_t *out)
+{
+	if (b > SIZE_MAX - a)
+		return -RPROC_EINVAL;
+
+	*out = a + b;
+	return 0;
+}
+
+static int elf_range_in_chunk(size_t chunk_offset, size_t chunk_len,
```
edmooring: This function really needs the Doxygen comment to describe its return values. I think it returns zero on success, -RPROC_EINVAL for overflows, and 1 for just wrong values. If this is the case, then why can't it just return -RPROC_EINVAL for all error cases? This would simplify the code flow for users of this function.

lib/remoteproc/elf_loader.c
```diff
@@ -157,6 +157,43 @@ static int elf_shstrndx(const void *elf_info)
 	}
 }
 
+static int elf_validate_header(const void *elf_info)
```
edmooring, line 134: Doxygen comments would be nice.

lib/remoteproc/elf_loader.c
```diff
@@ -545,9 +553,15 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
-		    offset + len < shdrs_offset + shdrs_size) {
+		if (elf_mul_size(elf_shnum(*img_info),
+				 elf_shentsize(*img_info),
+				 &shdrs_size) < 0)
+			return -RPROC_EINVAL;
+		range_in_chunk = elf_range_in_chunk(offset, len, shdrs_offset,
+						    shdrs_size);
+		if (range_in_chunk < 0)
+			return range_in_chunk;
+		if (!range_in_chunk) {
```
edmooring: What happens if elf_range_in_chunk() returns 1? I think that means that the inputs are wrong, but don't overflow.

lib/remoteproc/elf_loader.c
```diff
@@ -596,8 +597,12 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
 				  &shstrtab_size, NULL, NULL,
 				  NULL, NULL);
-		if (offset > shstrtab_offset ||
-		    offset + len < shstrtab_offset + shstrtab_size) {
+		range_in_chunk = elf_range_in_chunk(offset, len,
+						    shstrtab_offset,
+						    shstrtab_size);
+		if (range_in_chunk < 0)
+			return range_in_chunk;
+		if (!range_in_chunk) {
```
edmooring: Again, what happens if elf_range_in_chunk() returns 1? Some values don't get assigned.

edmooring (changes requested):

bentheredonethat:
Hi @edmooring  

Addressed in the updated series:

For the ELF type-name compliance findings, I left the `Elf32_Ehdr`, `Elf64_Ehdr`, and related ELF names unchanged. 

Agreed that these come from the ELF spec and should not be renamed just to satisfy the camel-case check.

For the newly added helper functions, I added Doxygen comments. In particular, `elf_range_in_chunk()` now documents its return values explicitly:
  - `1`: the requested range is contained in the current image chunk
  - `0`: the requested range is not contained in the current image chunk
  - `-RPROC_EINVAL`: overflow occurred while calculating the checked range endpoints

So `0` is not the success case for this helper. It is the normal streaming-loader path where the range is not present in
the current chunk, and the caller should set `*noffset` / `*nlen` and request that range. Returning `-RPROC_EINVAL` for that case would incorrectly turn a valid “load another chunk” condition into a hard failure.

I also renamed the local variables at the call sites from `range_in_chunk` to `range_loaded` to make the control flow clearer. If `elf_range_in_chunk()` returns `1`, the requested table/string table is already present in `img_data`, so the code intentionally falls through and proceeds to copy from the current chunk. In that path, `*noffset` and `*nlen` do  ot need to be assigned because no additional chunk is being requested.

If it returns `0`, the code assigns `*noffset` / `*nlen` and returns the current load state so the caller can fetch the missing range.

lib/remoteproc/elf_loader.c
```diff
@@ -6,11 +6,75 @@
...
+ * @return 0 on success, otherwise -RPROC_EINVAL.
+ */
+static int elf_add_size(size_t a, size_t b, size_t *out)
+{
+	if (b > SIZE_MAX - a)
+		return -RPROC_EINVAL;
+
+	*out = a + b;
+	return 0;
+}
```
arnopo: what about moving tes function in utilities.c ?
renaming them` safe/add_mult `or `mult/add_overflow ` ( or something else)

lib/remoteproc/elf_loader.c
```diff
@@ -6,11 +6,75 @@
...
+
+/**
+ * @brief Check if a range is within an image chunk.
+ *
+ * @param chunk_offset	Offset of the image chunk from the start of the image.
+ * @param chunk_len	Length of the image chunk.
+ * @param range_offset	Offset of the range from the start of the image.
+ * @param range_len	Length of the range.
```
arnopo: Not clear to me what you name range and image chunk.
Here what you want to check is that a  is in a range , right?
Could be also a more generic utilities that also clarify the usage

lib/remoteproc/elf_loader.c
```diff
@@ -6,11 +6,75 @@
...
+ *
+ * @return 1 if the range is contained within the chunk, 0 if the range is
+ *	   not contained within the chunk, otherwise -RPROC_EINVAL.
```
arnopo: I would  return 0 if Ok and an error otherwize. or return a boolean renaming the function  in cluding `is'  in the name

lib/remoteproc/elf_loader.c
```diff
@@ -186,6 +186,50 @@ static int elf_shstrndx(const void *elf_info)
...
+/**
+ * @brief Validate that an ELF header matches the supported header layout.
+ *
+ * @param elf_info	ELF information containing a copied ELF header.
+ *
+ * @return 0 on success, otherwise -RPROC_EINVAL.
+ */
+static int elf_validate_header(const void *elf_info)
+{
+	if (elf_is_64(elf_info) == 0) {
```
arnopo: i would inverse the logic for readability

```
if (elf_is_64(elf_info) ) {
		const Elf64_Ehdr *ehdr = elf_info;
[...]
} else {
		const Elf32_Ehdr *ehdr = elf_info;
}
```

lib/remoteproc/elf_loader.c
```diff
@@ -186,6 +186,50 @@ static int elf_shstrndx(const void *elf_info)
...
+			return -RPROC_EINVAL;
+		if (ehdr->e_ehsize != sizeof(Elf32_Ehdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_phnum != 0 &&
+		    ehdr->e_phentsize != sizeof(Elf32_Phdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_shnum != 0 &&
+		    ehdr->e_shentsize != sizeof(Elf32_Shdr))
+			return -RPROC_EINVAL;
+		if (ehdr->e_shnum != 0 && ehdr->e_shstrndx >= ehdr->e_shnum)
```
arnopo: this one seems common to both format

lib/remoteproc/elf_loader.c
```diff
@@ -484,6 +528,8 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 				memset(*img_info, 0, infosize);
 			}
 			memcpy(*img_info, img_data, tmpsize);
+			if (elf_validate_header(*img_info) < 0)
+				return -RPROC_EINVAL;
```
arnopo: directly return the `elf_validate_header` result

lib/remoteproc/elf_loader.c
```diff
@@ -543,14 +543,21 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
 
 		metal_log(METAL_LOG_DEBUG, "Loading ELF program header.\r\n");
 		phdrs_offset = elf_phoff(*img_info);
-		phdrs_size = elf_phnum(*img_info) * elf_phentsize(*img_info);
-		if (offset > phdrs_offset ||
-		    offset + len < phdrs_offset + phdrs_size) {
+		if (elf_mul_size(elf_phnum(*img_info),
+				 elf_phentsize(*img_info),
+				 &phdrs_size) < 0)
+			return -RPROC_EINVAL;
```
arnopo, line 558: same here return `elf_mul_size` result

lib/remoteproc/elf_loader.c
```diff
@@ -581,9 +589,15 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
...
-		    offset + len < shdrs_offset + shdrs_size) {
+		if (elf_mul_size(elf_shnum(*img_info),
+				 elf_shentsize(*img_info),
+				 &shdrs_size) < 0)
+			return -RPROC_EINVAL;
+		range_loaded = elf_range_in_chunk(offset, len, shdrs_offset,
+						  shdrs_size);
+		if (range_loaded < 0)
+			return range_loaded;
+		if (!range_loaded) {
```
arnopo: my coment on `elf_range_in_chunk` return should simplify this code

lib/remoteproc/elf_loader.c
```diff
@@ -297,6 +297,35 @@ static int *elf_load_state(void *elf_info)
...
+	if (sh_name >= name_table_size)
+		return 0;
+
+	remaining = name_table_size - sh_name;
+	name_len = strlen(name);
+	if (name_len >= remaining)
+		return 0;
+
+	candidate = name_table + sh_name;
+	return memcmp(name, candidate, name_len + 1) == 0;
```
arnopo, line 264: You return a boolean here and a int otherwise . 
 I suggest to return a boolean for this function

arnopo (changes requested):
minor comments .
@bentheredonethat: Did you  try to verify with an IA if some other overflows are detected ?

bentheredonethat:
> minor comments . @bentheredonethat: Did you try to verify with an IA if some other overflows are detected ?

@arnopo  can you please clarify what is meant by this ? i did some impact analysys by passing in overflowing values

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix - and make more generic utilities helper

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

lib/remoteproc/elf_loader.c
bentheredonethat: ok will fix

arnopo:
> can you please clarify what is meant by this ? i did some impact analysys by passing in overflowing values

My point behind my question is the validation part.  Could you share test you performed.
I wonder how we can ensure that we address all the vulnerability reported in https://www.sentinelone.com/vulnerability-database/cve-2026-37540/

nathalie-ckc:
7/15/2026 OpenAMP System Reference call:
Since some fixes for ELF loader have been submitted, someone is potentially using it. So, we should fix this (but the CVE severity is believed to be over-stated), as opposed to deprecating ELF loader entirely.
Ben will be able to come back to this in a few weeks. 
Team doesn't believe it's urgent - by 2026.10 release would be good.

bentheredonethat:
> > can you please clarify what is meant by this ? i did some impact analysys by passing in overflowing values
> 
> My point behind my question is the validation part. Could you share test you performed. I wonder how we can ensure that we address all the vulnerability reported in https://www.sentinelone.com/vulnerability-database/cve-2026-37540/

I initially performed a functional regression test using a valid ELF image loaded by the R5 onto the A53. The image was parsed, loaded, and executed successfully. In addition, I performed negative tests with deliberately malformed ELF images covering the validation paths introduced by this series:

@arnopo  below are additional tests performed:
  - invalid e_ehsize
  - invalid e_phentsize
  - invalid e_shentsize
  - out-of-range e_shstrndx
  - out-of-range program-header table offset/range
  - out-of-range section-header table offset/range
  - invalid section string-table offset/size
  - sh_name beyond the string-table boundary
  - section-name entry without an in-bounds NUL terminator

with latest tree. and AFAIK the earlier review comments have been addressed

lib/remoteproc/elf_loader.c
```diff
@@ -453,6 +498,9 @@ int elf_load_header(const void *img_data, size_t offset, size_t len,
 				memset(*img_info, 0, infosize);
 			}
 			memcpy(*img_info, img_data, tmpsize);
+			ret = elf_validate_header(*img_info);
+			if (ret < 0)
+				return ret;
```
arnopo, line 529: `img_info` is never free, if allocated.  But it seems that is it also the case for some other allocations.  I'm ok that we address error management  later, in such case please create an issue to log the issue.

lib/remoteproc/elf_loader.c
```diff
@@ -5,12 +5,45 @@
...
+ *
+ * @param chunk_offset	Offset of the image chunk from the start of the image.
+ * @param chunk_len	Length of the image chunk.
+ * @param data_offset	Offset of the ELF data range from the start of the image.
+ * @param data_len	Length of the ELF data range.
+ * @param loaded	Pointer to store whether the ELF data range is loaded.
+ *
+ * @return 0 on success, otherwise -RPROC_EINVAL.
+ */
+static int elf_range_is_loaded(size_t chunk_offset, size_t chunk_len,
```
arnopo: The term loaded is quite confusing as it does check if a range of memory has been loaded in destination memory. 

More than that, seems to me that this function could be generic and move to utilities.h  ( see below my proposal for other functions)
Here is a proposal assisted by AI.

```c
/**
 * @brief Check whether a range is fully contained within a base range.
 *
 * Returns true when [@p offset, @p offset + @p len] is fully contained
 * in [@p base, @p base + @p base_len]. Overflow maps to false.
 *
 * @param base      Start of the containing range.
 * @param base_len  Length of the containing range.
 * @param offset    Start of the range to check.
 * @param len       Length of the range to check.
 *
 * @return true if the range is contained, false otherwise (including overflow).
 */
#define is_in_range(base, base_len, offset, len) \
	(__extension__({ \
		typeof(base) __iir_base = (base); \
		typeof(base_len) __iir_base_len = (base_len); \
		typeof(offset) __iir_offset = (offset); \
		typeof(len) __iir_len = (len); \
		size_t __iir_base_end; \
		size_t __iir_end; \
		!OPENAMP_ADD_SAFE(__iir_base, __iir_base_len, &__iir_base_end) && \
		!OPENAMP_ADD_SAFE(__iir_offset, __iir_len, &__iir_end) && \
		__iir_base <= __iir_offset && __iir_base_end >= __iir_end; \
	}))
```

lib/utils/utilities.c
```diff
@@ -33,3 +35,21 @@ size_t safe_strcpy(char *dst, size_t d_size, const char *src, size_t s_size)
 
 	return size - nleft;
 }
+
+int safe_add_size(size_t a, size_t b, size_t *out)
```
arnopo: After few investigations it seems that some libraries already support this .
i would keep same template. Here is a proposal assisted by AI, that seems more generic  ( to put in utilities.h)

```c
/**
 * @internal
 *
 * @brief Safely add two unsigned values with overflow detection.
 *
 * Evaluates @p a + @p b and stores the result in the object pointed to by
 * @p out only if the operation does not overflow the type of @p *out.
 * Negative inputs are rejected and treated as overflow.
 * Uses @c __builtin_add_overflow when available; otherwise falls back to a
 * portable unsigned 64-bit range check.
 *
 * @param a   First operand (unsigned, up to 64-bit).
 * @param b   Second operand (unsigned, up to 64-bit).
 * @param out Non-null pointer to store the result.
 *
 * @return 0 on success, non-zero if overflow or a negative input is detected.
 */
#if defined(__HAVE_BUILTIN_OVERFLOW) && __HAVE_BUILTIN_OVERFLOW
#define OPENAMP_ADD_SAFE(a, b, out) (__extension__({ \
	typeof(out) __openamp_add_out = (out); \
	(!__openamp_add_out) || \
		__builtin_add_overflow((a), (b), __openamp_add_out); \
}))
#else
#define OPENAMP_ADD_SAFE(a, b, out) (__extension__({ \
	typeof(out) __openamp_add_out = (out); \
	typeof(a) __openamp_add_a_in = (a); \
	typeof(b) __openamp_add_b_in = (b); \
	int __openamp_add_of = 1; \
	if (__openamp_add_out && \
	    !(((typeof(__openamp_add_a_in))-1) < 0 && __openamp_add_a_in < 0) && \
	    !(((typeof(__openamp_add_b_in))-1) < 0 && __openamp_add_b_in < 0)) { \
		uint64_t __openamp_add_a = (uint64_t)__openamp_add_a_in; \
		uint64_t __openamp_add_b = (uint64_t)__openamp_add_b_in; \
		unsigned int __openamp_add_bits = sizeof(*__openamp_add_out) * 8U; \
		uint64_t __openamp_add_umax = __openamp_add_bits == 64U ? UINT64_MAX \
			: ((UINT64_C(1) << __openamp_add_bits) - UINT64_C(1)); \
		if (__openamp_add_a <= __openamp_add_umax && \
		    __openamp_add_b <= __openamp_add_umax - __openamp_add_a) { \
			*__openamp_add_out = \
				(typeof(*__openamp_add_out))(__openamp_add_a + __openamp_add_b); \
			__openamp_add_of = 0; \
		} \
	} \
	__openamp_add_of; \
}))
#endif

/**
 * @internal
 *
 * @brief Safely subtract two unsigned values with underflow detection.
 *
 * Evaluates @p a - @p b and stores the result in the object pointed to by
 * @p out only if @p a >= @p b and neither operand overflows the type of
 * @p *out. Negative inputs are rejected.
 * Uses @c __builtin_sub_overflow when available; otherwise falls back to a
 * portable unsigned 64-bit range check.
 *
 * @param a   Minuend (unsigned, up to 64-bit).
 * @param b   Subtrahend (unsigned, up to 64-bit).
 * @param out Non-null pointer to store the result.
 *
 * @return 0 on success, non-zero if underflow or a negative input is detected.
 */
#if defined(__HAVE_BUILTIN_OVERFLOW) && __HAVE_BUILTIN_OVERFLOW
#define OPENAMP_SUB_SAFE(a, b, out) (__extension__({ \
	typeof(out) __openamp_sub_out = (out); \
	(!__openamp_sub_out) || \
		__builtin_sub_overflow((a), (b), __openamp_sub_out); \
}))
#else
#define OPENAMP_SUB_SAFE(a, b, out) (__extension__({ \
	typeof(out) __openamp_sub_out = (out); \
	typeof(a) __openamp_sub_a_in = (a); \
	typeof(b) __openamp_sub_b_in = (b); \
	int __openamp_sub_of = 1; \
	if (__openamp_sub_out && \
	    !(((typeof(__openamp_sub_a_in))-1) < 0 && __openamp_sub_a_in < 0) && \
	    !(((typeof(__openamp_sub_b_in))-1) < 0 && __openamp_sub_b_in < 0)) { \
		uint64_t __openamp_sub_a = (uint64_t)__openamp_sub_a_in; \
		uint64_t __openamp_sub_b = (uint64_t)__openamp_sub_b_in; \
		unsigned int __openamp_sub_bits = sizeof(*__openamp_sub_out) * 8U; \
		uint64_t __openamp_sub_umax = __openamp_sub_bits == 64U ? UINT64_MAX \
			: ((UINT64_C(1) << __openamp_sub_bits) - UINT64_C(1)); \
		if (__openamp_sub_a <= __openamp_sub_umax && \
		    __openamp_sub_b <= __openamp_sub_umax && \
		    __openamp_sub_a >= __openamp_sub_b) { \
			*__openamp_sub_out = \
				(typeof(*__openamp_sub_out))(__openamp_sub_a - __openamp_sub_b); \
			__openamp_sub_of = 0; \
		} \
	} \
	__openamp_sub_of; \
}))
#endif
```
to use it 

```
size_t total;
if (OPENAMP_SUB_SAFE(len1, len2, &total))
    return -EINVAL;
```
I did not test it

arnopo (commented):
few proposal, on concern that can be addressed later,  else  close to be good to me

bentheredonethat:
@arnopo  Thanks for the review and the proposed helpers. I updated the series accordingly.

The changes are split across their original commit boundaries:
- Added generic checked addition, subtraction, and multiplication helpers.
- Added a generic range-containment helper and replaced the misleading loaded terminology with contained.
- Preserved the distinction between arithmetic overflow and an otherwise valid range that is not contained in the current image chunk. Overflow returns -RPROC_EINVAL; non-containment requests
  another chunk.

- Documented and implemented the ranges as half-open ranges.
- Made the builtin and fallback arithmetic paths consistently reject negative inputs.
- Added checked multiplication because the ELF program and section table sizes require it.
- Updated the program-header, section-header, and section string-table checks to use the new helpers.

I also addressed the img_info cleanup concern directly. elf_load_header() now tracks whether the current invocation allocated img_info. If header validation fails, it frees and nulls only that
newly allocated object. A caller-provided object remains caller-owned, avoiding an unconditional free or double-free through the outer cleanup path.

I tested the updated series with:
- A focused arithmetic and range test application, including overflow, underflow, negative operands, narrow output types, zero-length ranges, containment, and wrapping ranges.
- A native OpenAMP static-library build using the staged libmetal headers.
- An AArch64 cross-build using AMD GCC 15.3 and the supplied libmetal/sysroot.
- checkpatch.pl across all seven commits.

All tests and builds passed, and checkpatch reports no errors or warnings.

arnopo:
@bentheredonethat : need to address  CI build issue

nathalie-ckc:
8/12/26 System reference call: Ben to rebase to resolve the last remaining CI failure

.checkpatch.conf
```diff
@@ -19,4 +19,5 @@
 --ignore BIT_MACRO
 --ignore PREFER_KERNEL_TYPES
 --ignore NEW_TYPEDEFS
---ignore ARRAY_SIZE
\ No newline at end of file
+--ignore ARRAY_SIZE
+--ignore CAMELCASE
```
arnopo: I'm not in favor of adding this that will mask  cameCase issue for the rest of the code.
Remove this and kust ignore the CI check patch issue for this PR. It is Ok for me to merge with the CI camel case issue reported.

arnopo (changes requested):
LGTM after removing the checkpatch.conf update

.checkpatch.conf
bentheredonethat: @arnopo  ok pushed with that as only change

arnopo (approved):

arnopo:
Checkpatch complains about camelcase. Ok to ignore for this PR

tnmysh (approved):

PR #689, opened 2026-06-27
virtio: fix NULL pointer dereference in virtqueue_notification

masc2008:
Add a NULL check for the vq pointer before dereferencing it in virtqueue_notification(). Without this check, a NULL virtqueue pointer passed to the function would cause a crash when accessing vq->callback.

### DIFF
--- a/lib/virtio/virtqueue.c
+++ b/lib/virtio/virtqueue.c
@@ -620,7 +620,7 @@ static int vq_ring_enable_interrupt(struct virtqueue *vq, uint16_t ndesc)
 void virtqueue_notification(struct virtqueue *vq)
 {
 	atomic_thread_fence(memory_order_seq_cst);
-	if (vq->callback)
+	if (vq && vq->callback)
 		vq->callback(vq);
 }
 


arnopo (approved):

lib/virtio/virtqueue.c
```diff
@@ -620,7 +620,7 @@ static int vq_ring_enable_interrupt(struct virtqueue *vq, uint16_t ndesc)
 void virtqueue_notification(struct virtqueue *vq)
 {
 	atomic_thread_fence(memory_order_seq_cst);
-	if (vq->callback)
+	if (vq && vq->callback)
```
tnmysh, line 623: To me this check seems overkill. Why virtqueue_notification should be called at all if the appropriate virtqueue was not retrieved in the first place? The caller of this function must ensure that vq is retrieved properly before calling this function.

tnmysh: Also, when vq is NULL, please add error level log that vq is not expected to be NULL.

edmooring (approved):
This looks okay to me. 

I can see a case where it might be possible to call virtqueue_notification() on a virtueue that has a vring that has not yet been created.

tnmysh:
Okay, since @arnopo and @edmooring have acked, I am okay with this patch.

arnopo:
> Okay, since @arnopo and @edmooring have acked, I am okay with this patch.

@tnmysh : Because `virtqueue_notification` is an API, it is safer to add it, even if it seems excessive.

tnmysh (approved):

PR #695, opened 2026-08-03
docs: clarify release_cb initialization requirement

### LINKED ISSUE #659: rpmsg_ept_decref
When I was running `rpmsg_unregister_endpoint`, I found that `rpmsg_ept_decref` reported an error, and the `ept->release_cb pointer` was an abnormal value.
If` release_cb `is not initialized, it is not 0 in some cases, resulting in an error in calling the callback function pointer. So add `ept->release_cb = NULL` in `rpmsg_register_endpoint` function; Would it be better?
```c
void rpmsg_register_endpoint(struct rpmsg_device *rdev,
			     struct rpmsg_endpoint *ept,
			     const char *name,
			     uint32_t src, uint32_t dest,
			     rpmsg_ept_cb cb,
			     rpmsg_ns_unbind_cb ns_unbind_cb, void *priv)
{
	if (name)
		(void)safe_strcpy(ept->name, sizeof(ept->name), name, RPMSG_NAME_SIZE);
	else
		ept->name[0] = 0;

	ept->refcnt = 1;
	ept->addr = src;
	ept->dest_addr = dest;
	ept->cb = cb;
	ept->ns_unbind_cb = ns_unbind_cb;
	ept->priv = priv;
	ept->rdev = rdev;
	ept->release_cb = NULL;
	metal_list_add_tail(&rdev->endpoints, &ept->node);
}
```

atulakella:
It adds documentation to make it explicit that callers are responsible for zero-initializing the endpoint structure before calling rpmsg_create_ept, including the release_cb field.

Fixes #659


edmooring (approved):
Looks good to go.

lib/rpmsg/rpmsg.c
```diff
@@ -321,6 +321,29 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
...
+ * @param cb        Callback function for incoming messages
+ * @param unbind_cb Callback function for unbind events
+ *
+ * @return          Pointer to endpoint on success, NULL on error
+ *
+ * @note The caller is responsible for initializing the entire endpoint
+ *       structure to zero before calling this function. Failure to do so
+ *       may result in undefined behavior, particularly with the release_cb
+ *       callback pointer.
+ */
```
arnopo: The function is already documented here: https://github.com/OpenAMP/open-amp/blob/main/lib/include/openamp/rpmsg.h#L549 
 please update the existing documentation instead of duplicating it

lib/rpmsg/rpmsg.c
```diff
@@ -321,6 +321,29 @@ void rpmsg_register_endpoint(struct rpmsg_device *rdev,
...
 
+/**
+ * @brief Create and register an endpoint on a remote device.
+ *
+ * Creates a new endpoint with the specified parameters and registers it
+ * on the remote device. The endpoint structure must be zero-initialized
+ * by the caller before use, especially the release_cb field must be set
+ * to NULL if not used.
+ *
+ * @param ept       Pointer to endpoint structure (caller must zero-initialize)
```
arnopo: Not “must”: only release_cb should be set to zero if it is not used, but it must be set if it is used. The rest of the fields are properly initialized.

arnopo (changes requested):

lib/rpmsg/rpmsg.c
atulakella: Removed the duplicate doc block from rpmsg.c and moved the clarification into the existing rpmsg_create_ept doc in rpmsg.h instead.

lib/rpmsg/rpmsg.c
atulakella: Fixed that as well, updated to say only release_cb needs explicit handling by the caller, set to a valid callback if used or NULL if not. Other fields are already handled internally, removed the "must zero the whole struct" framing.

lib/include/openamp/rpmsg.h
```diff
@@ -562,7 +562,11 @@ static inline int rpmsg_send_nocopy(struct rpmsg_endpoint *ept,
  * As an option Some rpmsg clients can specify an endpoint with a specific
  * source address.
  *
- * @param ept		Pointer to rpmsg endpoint
+ * @param ept       Pointer to rpmsg endpoint. The release_cb field of the
+ *          endpoint must be explicitly set by the caller: assign it
+ *          to a valid callback if endpoint destruction notification
+ *          is needed, or set it to NULL otherwise. Other fields do
+ *          not require caller pre-initialization.
```
arnopo: you should use tabs instead of space to align , you can also have up to 100 chars per line

```
 * @param ept		Pointer to rpmsg endpoint. The release_cb field of the 	endpoint must be
 *			explicitly set by the caller: assign it to a valid callback if endpoint
 *			destruction notification is needed, or set it to NULL otherwise. Other
 *			fields do not require caller pre-initialization.

```

arnopo (approved):
ok to me, let wait a second review on final patch before merging it ,

arnopo:
@tnmysh @edmooring:  could you review this PR please

tnmysh (approved):
LGTM.

PR #696, opened 2026-08-26
Harden untrusted input handling

bentheredonethat:
This series hardens OpenAMP paths that consume firmware-controlled                 
data, messages from a remote processor, and caller-provided lengths.               
These inputs cross trust boundaries in AMP systems. They must be                   
validated before memory access, allocation, or callback dispatch.                  
                                                                                   
The remoteproc changes validate ELF section-name offsets and header                
table entry sizes. They bound each resource-table entry before its                 
handler runs and reject unsafe vring alignments before size or pointer             
calculations. The checks cover fixed resources and variable-length                 
VDEV vring and configuration data.                                                 
                                                                                   
The RPMsg RPC and proxy changes reject malformed message lengths.                  
They initialize omitted request bytes, report callback payload lengths             
without protocol headers, and bound retarget read and write copies to              
both source and destination capacities. Name-service strings are                   
terminated before being passed to application callbacks.                           
                                                                                   
The remaining changes correct the safe_strcpy() zero-fill off-by-one.              
They also make virtio MMIO queue setup honor its allocated capacity                
instead of appending beyond it.                                                    
                                                                                   
Validation performed:                                                              
                                                                                   
- git diff --check on every patch                                                  
- CMake Debug build against local Libmetal with -Wall -Wextra                      
- static and shared OpenAMP library targets built successfully                     
- CTest invoked; this configuration provides no tests

### DIFF
--- a/lib/include/openamp/remoteproc_virtio.h
+++ b/lib/include/openamp/remoteproc_virtio.h
@@ -24,6 +24,18 @@ extern "C" {
 /* maximum number of vring descriptors for a vdev limited by 16-bit data type */
 #define	RPROC_MAX_VRING_DESC	USHRT_MAX
 
+/**
+ * @brief Validate a remoteproc vring alignment.
+ *
+ * @param align	Vring alignment supplied by the resource table.
+ *
+ * @return true if the alignment is supported, otherwise false.
+ */
+static inline bool rproc_virtio_vring_align_valid(unsigned int align)
+{
+	return align != 0 && (align & (align - 1)) == 0;
+}
+
 /* cache invalidation helpers for resource table */
 #if defined(VIRTIO_USE_DCACHE)
 #define RSC_TABLE_FLUSH(x, s)		metal_cache_flush(x, s)
@@ -94,7 +106,7 @@ void rproc_virtio_remove_vdev(struct virtio_device *vdev);
  * @param va		vring virtual address
  * @param io		Pointer to vring I/O region
  * @param num_descs	Number of descriptors
- * @param align		vring alignment
+ * @param align		Nonzero power-of-two vring alignment
  *
  * @return 0 for success, negative value for failure.
  */
--- a/lib/include/openamp/virtqueue.h
+++ b/lib/include/openamp/virtqueue.h
@@ -141,7 +141,10 @@ struct vring_alloc_info {
 	/** Vring address. */
 	void *vaddr;
 
-	/** Vring alignment. */
+	/**
+	 * Vring alignment; must be a nonzero power of two.
+	 * Zero makes vring_init() calculate a NULL used-ring pointer.
+	 */
 	uint32_t align;
 
 	/** Number of descriptors in the vring. */
--- a/lib/proxy/rpmsg_retarget.c
+++ b/lib/proxy/rpmsg_retarget.c
@@ -271,16 +271,20 @@ int _write(int fd, const char *ptr, int len)
 	int ret;
 	struct rpmsg_rpc_syscall *syscall;
 	struct rpmsg_rpc_syscall resp;
-	int payload_size = sizeof(*syscall) + len;
+	size_t payload_size;
 	struct rpmsg_rpc_data *rpc = rpmsg_default_rpc;
 	unsigned char tmpbuf[MAX_BUF_LEN];
 	unsigned char *tmpptr;
 	int null_term = 0;
 
-	if (!rpc)
+	if (!rpc || !ptr || len < 0)
 		return -EINVAL;
 	if (fd == 1)
 		null_term = 1;
+	/* Reserve space for the payload and stdout's trailing NUL byte. */
+	if ((size_t)len > MAX_BUF_LEN - sizeof(*syscall) - null_term)
+		return -EINVAL;
+	payload_size = sizeof(*syscall) + (size_t)len + null_term;
 
 	syscall = (void *)tmpbuf;
 	syscall->id = WRITE_SYSCALL_ID;
@@ -289,10 +293,8 @@ int _write(int fd, const char *ptr, int len)
 	syscall->args.data_len = len + null_term;
 	tmpptr = tmpbuf + sizeof(*syscall);
 	memcpy(tmpptr, ptr, len);
-	if (null_term == 1) {
-		*(char *)(tmpptr + len + null_term) = 0;
-		payload_size += 1;
-	}
+	if (null_term == 1)
+		tmpptr[len] = '\0';
 	resp.id = 0;
 	ret = rpmsg_rpc_send(rpc, tmpbuf, payload_size,
 			     (void *)&resp, sizeof(resp));
--- a/lib/proxy/rpmsg_retarget.c
+++ b/lib/proxy/rpmsg_retarget.c
@@ -220,7 +220,7 @@ int _read(int fd, char *buffer, int buflen)
 	unsigned char tmpbuf[MAX_BUF_LEN];
 	int ret;
 
-	if (!rpc || !buffer || buflen == 0)
+	if (!rpc || !buffer || buflen <= 0)
 		return -EINVAL;
 
 	/* Construct rpc payload */
@@ -238,12 +238,18 @@ int _read(int fd, char *buffer, int buflen)
 	if (ret >= 0) {
 		if (resp->id == READ_SYSCALL_ID) {
 			if (resp->args.int_field1 > 0) {
-				int tmplen = resp->args.data_len;
+				size_t tmplen = resp->args.data_len;
+				size_t payload_capacity;
 				unsigned char *tmpptr = tmpbuf;
 
 				tmpptr += sizeof(*resp);
-				if (tmplen > buflen)
-					tmplen = buflen;
+				payload_capacity = MAX_BUF_LEN - sizeof(*resp);
+				/* Clamp to the source capacity first. */
+				if (tmplen > payload_capacity)
+					tmplen = payload_capacity;
+				/* Only reduce the clamped length for the destination. */
+				if (tmplen > (size_t)buflen)
+					tmplen = (size_t)buflen;
 				memcpy(buffer, tmpptr, tmplen);
 			}
 			ret = resp->args.int_field1;
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -132,22 +132,18 @@ static int elf_shstrndx(const void *elf_info)
 static int elf_validate_table_entry_sizes(const void *elf_info)
 {
 	if (elf_is_64(elf_info) == 0) {
-		const Elf32_Ehdr *ehdr = elf_info;
+		const struct elf32_info *einfo = elf_info;
 
-		if (ehdr->e_phnum != 0 &&
-		    ehdr->e_phentsize != sizeof(Elf32_Phdr))
+		if (einfo->ehdr.e_phnum != 0 && einfo->ehdr.e_phentsize != sizeof(*einfo->phdrs))
 			return -RPROC_EINVAL;
-		if (ehdr->e_shnum != 0 &&
-		    ehdr->e_shentsize != sizeof(Elf32_Shdr))
+		if (einfo->ehdr.e_shnum != 0 && einfo->ehdr.e_shentsize != sizeof(*einfo->shdrs))
 			return -RPROC_EINVAL;
 	} else {
-		const Elf64_Ehdr *ehdr = elf_info;
+		const struct elf64_info *einfo = elf_info;
 
-		if (ehdr->e_phnum != 0 &&
-		    ehdr->e_phentsize != sizeof(Elf64_Phdr))
+		if (einfo->ehdr.e_phnum != 0 && einfo->ehdr.e_phentsize != sizeof(*einfo->phdrs))
 			return -RPROC_EINVAL;
-		if (ehdr->e_shnum != 0 &&
-		    ehdr->e_shentsize != sizeof(Elf64_Shdr))
+		if (einfo->ehdr.e_shnum != 0 && einfo->ehdr.e_shentsize != sizeof(*einfo->shdrs))
 			return -RPROC_EINVAL;
 	}
 
--- a/lib/remoteproc/elf_loader.c
+++ b/lib/remoteproc/elf_loader.c
@@ -161,6 +161,25 @@ static void **elf_shstrtab_ptr(void *elf_info)
 	}
 }
 
+/**
+ * @brief Store the loaded ELF section string table size.
+ *
+ * @param elf_info	ELF information.
+ * @param size		Loaded section string table size.
+ */
+static void elf_set_shstrtab_size(void *elf_info, size_t size)
+{
+	if (elf_is_64(elf_info) == 0) {
+		struct elf32_info *einfo = elf_info;
+
+		einfo->shstrtab_size = size;
+	} else {
+		struct elf64_info *einfo = elf_info;
+
+		einfo->shstrtab_size = size;
+	}
+}
+
 static int *elf_load_state(void *elf_info)
 {
 	if (elf_is_64(elf_info) == 0) {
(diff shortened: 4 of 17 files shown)


arnopo:
@bentheredonethat 
Is these updates result from a report from a tools?

bentheredonethat:
@arnopo  correct static analysis yielded a collection of vulnerabilities.

arnopo:
> @arnopo correct static analysis yielded a collection of vulnerabilities.

In such case could you used the assisted by new tag described here: https://github.com/OpenAMP/open-amp/blob/main/README.md#assisted-by

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -655,7 +655,9 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 		return RPMSG_SUCCESS;
 	metal_io_block_read(io,
 			    metal_io_virt_to_offset(io, ns_msg->name),
-			    &name, sizeof(name));
+			    &name, sizeof(ns_msg->name));
+	/* Ensure the remote-supplied service name is NUL-terminated. */
+	name[RPMSG_NAME_SIZE] = '\0';
```
arnopo, line 673: No sure that this prevent application issue. 
the RPMSG_NAME_SIZE  define the max size. 
If application define a char name[RPMSG_NAME_SIZE]  table this not prevent an overflow in application

from my POV, the fix should be in application not here

lib/utils/utilities.c
```diff
@@ -26,10 +26,12 @@ size_t safe_strcpy(char *dst, size_t d_size, const char *src, size_t s_size)
 	}
 
 	/* Fill last characters with '\0' */
-	if (size < d_size)
-		memset(dst, '\0',  d_size - size + nleft);
-	else
+	if (size < d_size) {
+		/* Clear only bytes remaining after the copied data. */
+		memset(dst, '\0', d_size - (size_t)(dst - d));
+	} else {
```
arnopo: nitpicking: remove unecessary braces to better see you real update

lib/proxy/rpmsg_retarget.c
```diff
@@ -238,12 +238,17 @@ int _read(int fd, char *buffer, int buflen)
...
 
 				tmpptr += sizeof(*resp);
-				if (tmplen > buflen)
-					tmplen = buflen;
+				payload_capacity = MAX_BUF_LEN - sizeof(*resp);
+				/* Bound the peer length to source and destination. */
+				if (tmplen > payload_capacity)
+					tmplen = payload_capacity;
+				if (tmplen > (size_t)buflen)
+					tmplen = (size_t)buflen;
```
arnopo, line 251: what's happen if payload_capacity < buflen < tmplen ?
In such cas we should have tmplen = payload_capacity, right?

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -215,13 +215,19 @@ int virtio_mmio_device_init(struct virtio_mmio_device *vmdev, uintptr_t virt_mem
 /* Register preallocated virtqueues */
 void virtio_mmio_register_device(struct virtio_device *vdev, int vq_num, struct virtqueue **vqs)
 {
+	struct virtio_vring_info *vrings_info;
 	int i;
 
-	vdev->vrings_info = metal_allocate_memory(sizeof(struct virtio_vring_info) * vq_num);
-	/* TODO: handle error case */
+	if (!vdev || !vqs || vq_num <= 0)
+		return;
```
arnopo: I wonder if here we should return an error. This would modify the function prototype  but  it should remains compatible with legacy

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -300,8 +311,8 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 			    ((uintptr_t)metal_io_virt_to_phys(&vmdev->shm_io,
 			    (char *)vq->vq_ring.desc)) / 4096);
 
-	vdev->vrings_info[vdev->vrings_num].vq = vq;
-	vdev->vrings_num++;
+	/* Store the configured queue in its registered vring slot. */
+	vdev->vrings_info[idx].vq = vq;
```
arnopo, line 317: why `vdev->vrings_num++;` line has been removed?

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -341,7 +352,8 @@ static int virtio_mmio_create_virtqueues(struct virtio_device *vdev, unsigned in
 
 	(void)flags;
 
-	if (!vdev || !names || !vdev->vrings_info)
+	if (!vdev || !names || !vdev->vrings_info ||
+	    nvqs > vdev->vrings_num)
```
arnopo: squash on one line

lib/include/openamp/remoteproc_virtio.h
```diff
@@ -24,6 +24,22 @@ extern "C" {
 /* maximum number of vring descriptors for a vdev limited by 16-bit data type */
 #define	RPROC_MAX_VRING_DESC	USHRT_MAX
 
+/* Maximum supported alignment for a remoteproc vring. */
+#define RPROC_MAX_VRING_ALIGN	4096U
```
arnopo: What is the rational behin this limitation? 
if not mandatory I prefer to avoid it.

lib/include/openamp/virtqueue.h
```diff
@@ -141,7 +141,7 @@ struct vring_alloc_info {
 	/** Vring address. */
 	void *vaddr;
 
-	/** Vring alignment. */
+	/** Vring alignment; must be a nonzero power of two. */
```
arnopo, line 144: are you sure that it can be null ?  It should work...

lib/virtio/virtqueue.c
```diff
@@ -50,7 +50,12 @@ int virtqueue_create(struct virtio_device *virt_dev, unsigned short id,
 {
 	int status = VQUEUE_SUCCESS;
 
-	VQ_PARAM_CHK(ring == NULL, status, ERROR_VQUEUE_INVLD_PARAM);
+	if (!ring)
+		return ERROR_VQUEUE_INVLD_PARAM;
```
arnopo, line 55: why not keeping the initial `VQ_PARAM_CHK(ring == NULL, status, ERROR_VQUEUE_INVLD_PARAM);` ?

bentheredonethat: will fix

lib/include/openamp/virtqueue.h
bentheredonethat: will fix

lib/include/openamp/remoteproc_virtio.h
bentheredonethat: will fix

lib/virtio_mmio/virtio_mmio_drv.c
bentheredonethat: will fix

lib/virtio_mmio/virtio_mmio_drv.c
bentheredonethat: will fix

lib/virtio_mmio/virtio_mmio_drv.c
bentheredonethat: will fix

lib/proxy/rpmsg_retarget.c
bentheredonethat: will fix

lib/utils/utilities.c
bentheredonethat: will fix

lib/rpmsg/rpmsg_virtio.c
bentheredonethat: This change is intended to protect the internal rpmsg_virtio callback path from treating an unterminated remote-supplied field as a C string. The wire-format field can contain exactly RPMSG_NAME_SIZE non-NUL bytes, so the local buffer needs one additional byte for a terminator before it is passed to endpoint lookup or application callbacks.

I agree that this does not by itself prevent an application from copying the name into char name[RPMSG_NAME_SIZE] and then using it as a NUL-terminated string. Applications must still allocate RPMSG_NAME_SIZE + 1 bytes or use bounded string operations. This change addresses the separate library-side out-of-bounds-read risk and ensures the name delivered by this callback is terminated.

lib/remoteproc/elf_loader.c
```diff
@@ -122,6 +122,38 @@ static int elf_shstrndx(const void *elf_info)
...
+ *
+ * @return 0 on success, otherwise -RPROC_EINVAL.
+ */
+static int elf_validate_table_entry_sizes(const void *elf_info)
+{
+	if (elf_is_64(elf_info) == 0) {
+		const Elf32_Ehdr *ehdr = elf_info;
+
+		if (ehdr->e_phnum != 0 &&
+		    ehdr->e_phentsize != sizeof(Elf32_Phdr))
```
arnopo: squash on one line ( same for the condition below)

lib/rpmsg/rpmsg_virtio.c
arnopo: regarding Linux implementation https://elixir.bootlin.com/linux/v7.0.14/source/drivers/rpmsg/rpmsg_ns.c#L50

Ff we want to be aligned  we could have 

```
	/* don't trust the remote processor for null terminating the name */
	name[RPMSG_NAME_SIZE - 1] = '\0';
```

lib/proxy/rpmsg_retarget.c
arnopo: still not fixed in your last update, or i missed something?

lib/virtio_mmio/virtio_mmio_drv.c
arnopo: still not fixed

lib/include/openamp/virtqueue.h
arnopo: sorry the comment should be "are you sure that it can not be null ? It should work..."

lib/proxy/rpmsg_retarget.c
bentheredonethat: will add comment sorry

lib/virtio_mmio/virtio_mmio_drv.c
bentheredonethat: will add comment sorry

lib/include/openamp/virtqueue.h
bentheredonethat: will add comment

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -653,7 +653,12 @@ static int rpmsg_virtio_ns_callback(struct rpmsg_endpoint *ept, void *data,
 		return RPMSG_SUCCESS;
 	metal_io_block_read(io,
 			    metal_io_virt_to_offset(io, ns_msg->name),
-			    &name, sizeof(name));
+			    &name, sizeof(ns_msg->name));
```
arnopo, line 656: It seems to me that `&name` should be `name`, and perhaps this is the issue reported by your tool.
I cannot figure out why updating the sizeof fixes something here.

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -236,8 +244,13 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 	struct virtio_vring_info _vring_info = {0};
 	struct virtio_vring_info *vring_info = &_vring_info;
 	struct vring_alloc_info *vring_alloc_info;
-	struct virtio_mmio_device *vmdev = metal_container_of(vdev,
-							      struct virtio_mmio_device, vdev);
+	struct virtio_mmio_device *vmdev;
+
+	if (!vdev || !vdev->vrings_info || idx >= vdev->vrings_num) {
+		metal_log(METAL_LOG_ERROR, "Invalid virtqueue index\n");
```
arnopo, line 250: not related to virtqueue index
```suggestion
		metal_log(METAL_LOG_ERROR, "Invalid virtqueue setup\n");
```

lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -300,8 +313,11 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 			    ((uintptr_t)metal_io_virt_to_phys(&vmdev->shm_io,
 			    (char *)vq->vq_ring.desc)) / 4096);
 
-	vdev->vrings_info[vdev->vrings_num].vq = vq;
-	vdev->vrings_num++;
+	/*
+	 * Registration already set vrings_num to the allocated slot count.
+	 * Configuring a slot does not increase that count.
+	 */
```
arnopo, line 319: please address this commit in a separate PR because it seems thatther is a confusion between the Index of the virtqueue and the place in the vrings_info

lib/proxy/rpmsg_retarget.c
arnopo: my bad, I rechecked, the algo is valid

lib/include/openamp/virtqueue.h
```diff
@@ -141,7 +141,10 @@ struct vring_alloc_info {
 	/** Vring address. */
 	void *vaddr;
 
-	/** Vring alignment. */
+	/**
+	 * Vring alignment; must be a nonzero power of two.
```
arnopo, line 145: It is still not confirmed that an alignment of 1 is invalid. I tried to find the limitation you added, but without success. Please point it out to me; otherwise, remove the constraint.

lib/virtio/virtqueue.c
```diff
@@ -51,6 +51,12 @@ int virtqueue_create(struct virtio_device *virt_dev, unsigned short id,
 	int status = VQUEUE_SUCCESS;
 
 	VQ_PARAM_CHK(ring == NULL, status, ERROR_VQUEUE_INVLD_PARAM);
+	if (!ring)
```
arnopo, line 54: This commit should be split into 2  one for remoteproc one for virtio

lib/remoteproc/rsc_table_parser.c
```diff
@@ -155,45 +155,124 @@ static const rsc_handler rsc_handler_table[] = {
...
+	unsigned int idx;
+	size_t offset;
+	size_t rsc_offset;
 	int status = 0;
 
 	/* Validate rsc table header fields */
 
 	/* Minimum rsc table size */
-	if (sizeof(struct resource_table) > size) {
+	if (!rsc_table || sizeof(struct resource_table) > size) {
```
arnopo, line 237:
```
if(!rsc_table)
    return -EINVAL;
```

lib/rpmsg/rpmsg_virtio.c
```diff
@@ -577,6 +579,17 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
...
+		    rp_hdr->len > len - sizeof(*rp_hdr)) {
+			len = virtqueue_get_buffer_length(rvdev->rvq, idx);
+			rpmsg_virtio_return_buffer(rvdev, rp_hdr, len, idx);
+			if (VIRTIO_ENABLED(VQ_RX_EMPTY_NOTIFY))
+				release = true;
+			else
+				virtqueue_kick(rvdev->rvq);
+			metal_mutex_release(&rdev->lock);
+			continue;
+		}
```
arnopo, line 592: what about the inverse logic?
```
if (len >= sizeof(*rp_hdr) &&    rp_hdr->len <=  len - sizeof(*rp_hdr)) {
		rp_hdr->reserved = idx;

		/* Get the channel node from the remote device channels list. */
		ept = rpmsg_get_ept_from_addr(rdev, rp_hdr->dst);
		rpmsg_ept_incref(ept);
		RPMSG_BUF_HELD_INC(rp_hdr);

		if (ept) {
			if (ept->dest_addr == RPMSG_ADDR_ANY) {
				/*
				 * First message received from the remote side,
				 * update channel destination address
				 */
				ept->dest_addr = rp_hdr->src;
			}

			RPMSG_ASSERT(status >= 0,
				     "unexpected callback status\r\n");
		}
}
```

arnopo (changes requested):

bentheredonethat:
@arnopo  per offline discussion - closing this PR - will open 4 new PRs for the MMIO, remoteproc, rpmsg-virtio and rpc/proxy commits

arnopo:
> @arnopo per offline discussion - closing this PR - will open 4 new PRs for the MMIO, remoteproc, rpmsg-virtio and rpc/proxy commits

Thanks!
for tracking new PR are 

- #701 
- #702  
- #703
- #700

PR #699, opened 2026-09-08
virtqueue: add writable buffer helper

WallaceIT:
An application, specially at device side, might need to know if a buffer has been marked as writable by the communication partner. An example is the virtio SPI protocol, in which the direction of an half-duplex communication can be determined only by checking if the buffer containing the data is writable or not.

Add a simple helper to test if a buffer is writable, that is, if the VRING_DESC_F_WRITE is present on it.


lib/include/openamp/virtqueue.h
```diff
@@ -401,6 +401,18 @@ uint32_t virtqueue_get_desc_size(struct virtqueue *vq);
 uint32_t virtqueue_get_buffer_length(struct virtqueue *vq, uint16_t idx);
 void *virtqueue_get_buffer_addr(struct virtqueue *vq, uint16_t idx);
 
+/**
+ * @internal
```
arnopo: Remove the internal tag as it is an API that could be used by application

lib/include/openamp/virtqueue.h
```diff
@@ -401,6 +401,18 @@ uint32_t virtqueue_get_desc_size(struct virtqueue *vq);
...
+ *
+ * @brief Test if buffer is marked as writable
+ *
+ * @param vq	Pointer to VirtIO queue control block
+ * @param idx	Index used in vring desc table
+ *
+ * @return true if the buffer is marked as writable, false otherwise
+ */
+bool virtqueue_is_buffer_writable(struct virtqueue *vq, uint16_t idx);
```
tnmysh: This virtqueue layer is not expected to be used by application/firmware directly, instead via rpmsg_virtio layer. I am not sure if we ever need this API for rpmsg_virito.

lib/include/openamp/virtqueue.h
```diff
@@ -401,6 +401,18 @@ uint32_t virtqueue_get_desc_size(struct virtqueue *vq);
...
 
+/**
+ * @internal
+ *
+ * @brief Test if buffer is marked as writable
+ *
+ * @param vq	Pointer to VirtIO queue control block
+ * @param idx	Index used in vring desc table
+ *
+ * @return true if the buffer is marked as writable, false otherwise
```
arnopo: The  `VRING_DESC_F_WRITE` flag means that the buffer is writable by the device. I would add the notion of device.
 ```suggestion
  * @return true if the buffer is marked as device-writable, false otherwise
```

File: lib/include/openamp/virtqueue.h (hunk 2 above)
Comment (arnopo) [inline lib/include/openamp/virtqueue.h]:
```suggestion
bool virtqueue_is_buffer_device_writable(struct virtqueue *vq, uint16_t idx);
```

WallaceIT: I am using open-amp for generic virtio over remoteproc experimentation in Zephyr, and accessing internal flags details is inappropriate. I proposed this new API because I saw that (at least) another has been integrated with the same usecase in mind (9c4ad74ab4e31f4b74be78a69900d683cbd43680).

lib/include/openamp/virtqueue.h
WallaceIT: Done!

lib/include/openamp/virtqueue.h
WallaceIT: Done!

arnopo (approved):
LGTM

lib/include/openamp/virtqueue.h
arnopo: Seems reasonable to me if needed for another virtio device  than rpmsg

edmooring (approved):
Looks reasonable to me.

PR #700, opened 2026-09-14
virtio: mmio: Bound registered virtqueue setup

bentheredonethat:
MMIO registration allocates vrings_num entries and records the value as the device's vring count. Queue setup then uses the count as an append index, so its first store is immediately beyond the allocation.

Treat vrings_num as the registered capacity, validate setup and bulk creation against it, and store each configured queue in its indexed slot. Also validate registration inputs and allocation success.

Assisted-by: Codex:GPT-5


lib/virtio_mmio/virtio_mmio_drv.c
```diff
@@ -300,8 +313,11 @@ struct virtqueue *virtio_mmio_setup_virtqueue(struct virtio_device *vdev,
 			    ((uintptr_t)metal_io_virt_to_phys(&vmdev->shm_io,
 			    (char *)vq->vq_ring.desc)) / 4096);
 
-	vdev->vrings_info[vdev->vrings_num].vq = vq;
-	vdev->vrings_num++;
+	/*
+	 * Registration already set vrings_num to the allocated slot count.
+	 * Configuring a slot does not increase that count.
+	 */
+	vdev->vrings_info[idx].vq = vq;
```
arnopo: The behavior does not seem to be the same. In the original code, `vdev->vrings_num `is incremented, but that is no longer the case.
Regarding the` virtio_mmio_create_virtqueues `function, something seems strange to me. In `virtio_mmio_create_virtqueues`, we set the `vq` pointer from `vdev->vrings_info`, and here we do the reverse.
This code probably needs a deeper fix.
@danmilea, could you take a look? I’m not able to test this piece of code on my side.

PR #702, opened 2026-09-14
Harden untrusted input rpmsg virtio

bentheredonethat:


lib/rpmsg/rpmsg_virtio.c
```diff
@@ -579,10 +581,18 @@ static void rpmsg_virtio_rx_callback(struct virtqueue *vq)
 		}
 
 		rp_hdr->reserved = idx;
+		ept = NULL;
```
arnopo, line 584: You can directly initialize it when you declare the local variale

`	struct rpmsg_endpoint *ept = NULL;`

arnopo (commented):
one extra comment else LGTM

lib/rpmsg/rpmsg_virtio.c
bentheredonethat: @arnopo 

Good catch, fixed — but not quite as written, and I want to flag why.

ept was declared at function scope, outside the while (1) loop, and nothing resets it at the bottom of the loop body. Adding the initializer to that declaration would only run once, on entry.
On any later iteration where the new length check rejects the buffer, the lookup is skipped and ept would still hold the endpoint from the previous iteration — so we'd dispatch the callback to
the wrong endpoint and call rpmsg_ept_decref() on it a second time.

So I moved the declaration into the loop body instead, which gives you the direct initialization you asked for and keeps the per-iteration reset:

```
while (1) {
        /*
         * Scoped to the loop body so that every iteration starts
         * without an endpoint. The lookup below is skipped for
         * malformed buffers, so a function-scope variable would keep
         * the endpoint found by an earlier iteration and that stale
         * pointer would be dispatched to and dereferenced again.
         */
        struct rpmsg_endpoint *ept = NULL;
```

I've added that comment in the code and a paragraph to the commit message covering the scope change, since it's not obvious from the diff alone. Folded into "rpmsg: virtio: Validate received
message lengths" for v3.

arnopo (approved):
It might be too much AI verbosity in inline comment, but that’s just my feeling. If other reviewers don’t report it, we can keep it as is.

Issue #343, opened 2022-02-04
remoteproc_virtio_notify return status not used


tammyleino:
The remoteproc_virtio_notify function is used in functions defined in remoteproc_virtio.c as rpvdev->notify.  The implementation is ultimately defined by the BSP. This function returns the status of the operation but the status is not checked by any of the calling routines.  Should the return status of remoteproc_virtio_notify be changed to void?

edmooring:
I hate losing potential error information, but I also can't see what the caller could sanely do in the face of an error. I am not opposed to changing the return status. However, this is an API change that will break existing code, so we have to be careful. 

@arnopo Do you think we could do this with a CMake setting?

arnopo:
I can see 2 issues: 

1.  The ` rpvdev->notify`  in  remoteproc_virtio.c can be null, leading to crash

=>   `rpvdev->notify ==  0` is valid if user doesn't use a mailbox to notify the remote processor. The call should be protected by `if (rpvdev->notify)`

2.  The status is not checked.  

=> what to do in case of error? This depends on the virtio backend implementation. The remoteproc_virtio is only one.
So change the API seems not a solution.
I would be in favor of  testing the status in remoteproc_virtio.c and printing a warning.

tammyleino:
@arnopo Yes, the function pointer should definitely be checked before using it.  I will open a PR eventually for this and possibly other remoteproc op function pointers.  I think logging the error will be fine for OpenAMP in general, but I think in my particular case, I need to not allow an error to be returned from the driver's notify routine.  My driver is just triggering an inter-core interrupt via a register, which won't fail.  The potential fail case is if the incoming remoteproc pointer is null.  This should have been caught at the API level before getting all the way to the driver.

atulakella:
Should I send a PR for this?
Plan is to null-check `rpvdev->notify` before calling it (a null notify is valid, just means no mailbox is configured, so return early), and log a warning via `metal_log` on nonzero return instead of changing the exported return type, per the discussion above. Will add a host-side unit test covering both the null and nonzero-return notify paths.

arnopo:
> Should I send a PR for this? Plan is to null-check `rpvdev->notify` before calling it (a null notify is valid, just means no mailbox is configured, so return early), and log a warning via `metal_log` on nonzero return instead of changing the exported return type, per the discussion above. Will add a host-side unit test covering both the null and nonzero-return notify paths.

yes feel free, thanks!

PR #705, opened 2026-09-18
remoteproc_virtio: guard null notify and check return status

atulakella:
`rpvdev->notify` is called from four places in `remoteproc_virtio.c` (`rproc_virtio_virtqueue_notify`, `rproc_virtio_set_status`, `rproc_virtio_set_features`, `rproc_virtio_write_config`). None guarded against `notify` being NULL, valid when no mailbox/transport is configured, and none checked the returned status, so a failed notify was silently dropped.

Refs #343

Added a shared `rpvdev_notify()` helper used at all four call sites: skip when `notify` is NULL, log a warning via `metal_log` on nonzero return.

The original issue referenced a single call site, but after checking the current codebase I found the same gap in four places, so the fix is consolidated into one helper rather than four inline patches.


lib/remoteproc/remoteproc_virtio.c
```diff
@@ -32,6 +32,20 @@ static void rproc_virtio_delete_virtqueues(struct virtio_device *vdev)
 	}
 }
 
+static void rpvdev_notify(struct remoteproc_virtio *rpvdev, uint32_t id)
+{
+	int ret;
+
+	if (!rpvdev->notify)
```
tnmysh, line 39: I'd prefer to put this check way early where rpvdev is created. If the notify() callback is not provided, then probably fail there. This way, we force platforms to provide atleast stub function when notify() callback is not needed. This way, we don't have to make this check everytime.

As far as the error check goes, we should leave it to the platforms if they want to print error message or not on the failure. Here it's just printing a warning message and not taking any other action.

atulakella: agreed, the platform's own `notify()` implementation is better positioned to decide whether/how to report a failure than a generic warning at this layer. I'll drop the `metal_log` call.

On moving the check to vdev-creation time, want to flag that this reverses something explicit in the original discussion, #343. arnopo's comment states `rpvdev->notify == 0` is valid when a platform doesn't use a mailbox to notify the remote processor. If we move the check to `rproc_virtio_create_vdev()` and fail there, any platform currently passing `NULL` for that reason would need to start supplying an explicit stub instead. That's a real behavior change for existing callers, not just a relocation of the same check.

I'm fine making that change if it's the direction you'd rather go, happy to enforce a non-NULL `notify` at creation time (in `rproc_virtio_create_vdev()`) and drop the per-call guard entirely. Just want to confirm that's the intended tradeoff before I do it.

arnopo: Since we cannot guarantee that polling mode is not used, we have to keep it. Otherwise, we should apply the deprecation process.

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -32,6 +32,20 @@ static void rproc_virtio_delete_virtqueues(struct virtio_device *vdev)
...
+		return;
+
+	ret = rpvdev->notify(rpvdev->priv, id);
+	if (ret)
+		metal_log(METAL_LOG_WARNING,
+			  "rpvdev notify failed: id=%lu, ret=%d\r\n",
```
arnopo: nitpicking: remove rpvdev to save few bytes

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -32,6 +32,20 @@ static void rproc_virtio_delete_virtqueues(struct virtio_device *vdev)
...
+static void rpvdev_notify(struct remoteproc_virtio *rpvdev, uint32_t id)
+{
+	int ret;
+
+	if (!rpvdev->notify)
+		return;
+
+	ret = rpvdev->notify(rpvdev->priv, id);
+	if (ret)
+		metal_log(METAL_LOG_WARNING,
```
arnopo: This is an error not a warning

lib/remoteproc/remoteproc_virtio.c
atulakella: I'll change that, and amend.

lib/remoteproc/remoteproc_virtio.c
atulakella: Got it!

atulakella:
Pushed an update: kept the NULL guard for polling mode, switched to METAL_LOG_ERROR, and dropped the rpvdev prefix from the message. Also rebased onto current main.

PS: would the maintainers be open to an OpenAMP Discord server for broader, less formal conversations? Review threads are great for code, but design questions like the polling-mode/deprecation discussion here might fit a lower-friction channel.

tnmysh (approved):

arnopo:
> PS: would the maintainers be open to an OpenAMP Discord server for broader, less formal conversations? Review threads are great for code, but design questions like the polling-mode/deprecation discussion here might fit a lower-friction channel.

The channel already exists https://discord.com/channels/881957442341208095/1030188457739427911

lib/remoteproc/remoteproc_virtio.c
```diff
@@ -32,6 +32,20 @@ static void rproc_virtio_delete_virtqueues(struct virtio_device *vdev)
...
 
+static void rpvdev_notify(struct remoteproc_virtio *rpvdev, uint32_t id)
+{
+	int ret;
+
+	if (!rpvdev->notify)
+		return;
+
+	ret = rpvdev->notify(rpvdev->priv, id);
+	if (ret)
```
arnopo: Missing it during my previous review.
`ret` is not a boolean better to have 

```suggestion
	if (ret !=  0)
```

